PR #2, opened 2015-02-15
Add CIFAR10 dataset

vdumoulin:
Based off `MNIST`, uses data that can currently be obtained through Pylearn2's `download_cifar10.sh` script.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ sudo: false
 cache:
   directories:
     - $TRAVIS_BUILD_DIR/mnist
+    - $TRAVIS_BUILD_DIR/cifar10
 branches:
   only:
     - master
@@ -26,6 +27,13 @@ before_install:
              -O http://yann.lecun.com/exdb/mnist/t10k-images-idx3-ubyte.gz \
              -O http://yann.lecun.com/exdb/mnist/t10k-labels-idx1-ubyte.gz
         gunzip *-ubyte.gz
+        cd -
+      fi
+      # Download CIFAR10 for tests if not loaded from cache
+      if find cifar10 -empty | read; then
+        mkdir cifar10
+        cd cifar10
+        curl http://www.cs.utoronto.ca/~kriz/cifar-10-python.tar.gz | tar xzf -
         cd -
       fi
       export FUEL_DATA_PATH=$PWD
--- a/fuel/datasets/__init__.py
+++ b/fuel/datasets/__init__.py
@@ -1,6 +1,7 @@
 from fuel.datasets.base import Dataset, InMemoryDataset
 
 from fuel.datasets.container import ContainerDataset
+from fuel.datasets.cifar10 import CIFAR10
 from fuel.datasets.mnist import MNIST
 from fuel.datasets.text import TextFile
 from fuel.datasets.billion import OneBillionWord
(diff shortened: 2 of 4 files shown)


bartvm:
Tests are just failing right now because the data is missing on Travis. I think we can add that as with MNIST (we can store it in the cache in Travis's new container environment so that it doesn't need to be redownloaded each time).

I'll make a ticket though, because I feel that there is a lot of code duplication between this and MNIST, and there's a variety of things that could be factored out for use by future datasets as well.

vdumoulin:
@bartvm Even though I think I've properly added the CIFAR10 files to cache, the tests fail. I'm not very Travis-savvy, is there something I left out?

bartvm:
On my phone, but the log says:

tar: Old option `f' requires an argument.
Try `tar --help' or `tar --usage' for more information.

Maybe there's something wrong with the Tar command so that it doesn't actually unpack?

bartvm:
On my phone, but the log says:

tar: Old option `f' requires an argument.
Try `tar --help' or `tar --usage' for more information.

Maybe there's something wrong with the Tar command so that it doesn't actually unpack?

vdumoulin:
I think I found what the issue was, but I can't fix it since I don't think I can clear the cache by myself.

bartvm:
I think the bash command still doesn't fully work. The problem is that `curl -O` doesn't write to `stdout`, but to a file instead. Also, `tar f` requires you to give a filename to read from, but in this case you want to read from `stdin` for which you need to pass `-`. I think the final command should look like this:
```bash
curl http://www.cs.utoronto.ca/~kriz/cifar-10-python.tar.gz | tar xzf -
```

vdumoulin:
Yes, I think you're right. I gleaned the command from the Pylearn2 script, but I missed the dash at the end of the `tar` command. There's also apparently a way to use the same trick with `curl -O` which is present in the original command:

```
curl -O http://www.cs.utoronto.ca/~kriz/cifar-10-python.tar.gz - | tar xvzf -
```

I'll try to push a commit to force resetting the cache properly, and if it works I'll squash all commits related to fixing my mistake.

vdumoulin:
@bartvm I wasn't able to repair my cache mistake. I put the right `curl` command and cleaned up the commits, but I think removing `cifar10` from the cache will require manual intervention.

Sorry about that!

bartvm:
No problem! I cleared the cache manually and restarted the build

bartvm:
Python 2 passes now, but you should use `from six.moves import cPickle` for Python 3 compatibility.

vdumoulin:
@bartvm Yes! For `xrange` as well. I'm on it.

bartvm:
Try opening it with mode `'rb'`, for Python 3 refuses to cast automatically between string- and byte-encoded data.

bartvm:
Sorry, my bad, I only just realised you were trying to unpickle the file that you downloaded, which I guess was pickled with Python 2... They use different encoding schemes, but this could work:
```python
cPickle.load(f, encoding='latin1')
```

vdumoulin:
That will fail under Python 2, since `cPickle` doesn't accept the `encoding` argument.

vdumoulin:
One option would be this:

```python
try:
    data = cPickle.load(f, encoding='latin1')
except TypeError:
    data = cPickle.load(f)
```

but it isn't very readable in my opinion.

bartvm:
I agree, but you can use `if six.PY3:` which is okay I think.

vdumoulin:
@bartvm Thanks for the tip. The tests now pass.

PR #8, opened 2015-02-19
add sources

memimo:
Sources are in the parameters docstring but missing in actual func define.

### DIFF
--- a/fuel/streams/__init__.py
+++ b/fuel/streams/__init__.py
@@ -37,8 +37,9 @@ class AbstractDataStream(object):
         given by the dataset.
 
     """
-    def __init__(self, iteration_scheme=None):
+    def __init__(self, iteration_scheme=None, sources=None):
         self.iteration_scheme = iteration_scheme
+        self._sources = sources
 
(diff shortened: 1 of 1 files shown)


bartvm:
They're in the docstring as an attribute though, and both concrete classes do implement them as properties. I don't think it makes much sense to pass them as a keyword argument, because the `sources` attribute represents the sources that are *actually* returned by the data stream. If you set them like this, your data stream could have `sources = ('features',)` but still be returning both features and targets, which would mess things up.

A particular implementation of a data stream could edit the sources of course, and allow those to be given as an argument. For example:

```python
class CopyStream(DataStreamWrapper):
    """Adds a new copy of the first data source.

    The copy gets the given source name.

    """
    def __init__(self, source, **kwargs):
        super(CopyStream, self).__init__(**kwargs)
        self.sources = self.sources + (source,)

        data += (data[0],)
        return data
```

memimo:
Sorry for late reply. I see, I guess also what I wanted can be handled by passing sources to dataset, which I missed.

PR #15, opened 2015-02-23
Add BinarizedMNIST dataset

vdumoulin:
Thin wrapper around Hugo Larochelle's publicly available binarized MNIST dataset.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,19 +34,9 @@ before_install:
       if find binarized_mnist -empty | read; then
         mkdir binarized_mnist
         cd binarized_mnist
-        python -c "import numpy; numpy.save('binarized_mnist_train.npy',\
-                   numpy.genfromtxt(\
-                   'http://www.cs.toronto.edu/~larocheh/public/datasets/binarized_mnist/binarized_mnist_train.amat',\
-                   dtype='uint8'))"
-        python -c "import numpy; numpy.save('binarized_mnist_valid.npy',\
-                   numpy.genfromtxt(\
-                   'http://www.cs.toronto.edu/~larocheh/public/datasets/binarized_mnist/binarized_mnist_valid.amat',\
-                   dtype='uint8'))"
-        python -c "import numpy; numpy.save('binarized_mnist_test.npy',\
-                   numpy.genfromtxt(\
-                   'http://www.cs.toronto.edu/~larocheh/public/datasets/binarized_mnist/binarized_mnist_test.amat',\
-                   dtype='uint8'))"
-        rm -r www.cs.toronto.edu
+        curl -O http://www.cs.toronto.edu/~larocheh/public/datasets/binarized_mnist/binarized_mnist_train.amat\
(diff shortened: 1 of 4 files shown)


vdumoulin:
Right, I forgot about the unit tests. I'll add them.

bartvm:
Cool! I'd be okay with merging this so that it doesn't end up getting stalled endlessly, but we should move away from hardcoding the file name in `load`, so that the user can override it if necessary. See https://github.com/bartvm/fuel/pull/9#discussion_r25128068.

bartvm:
You'll need to add the data to the cache as well. We should probably copy the `skip_if_not_available` function from Blocks' `tests/__init__.py` as well so that these tests get automatically skipped if the data is not available though.

vdumoulin:
Right, I also forgot about the cache.

I factored out the hardcoded path from the `load` method, by the way.

vdumoulin:
@bartvm I think I messed up in adapting `.travis.yml`, could you have a look and tell me if you see something obvious missing?

vdumoulin:
I think I found it... I'm using numpy before it gets installed.

bartvm:
How long does the conversion take? If it's not too long I'd argue that the dataset itself should just perform the conversion when loading the data. I like the idea of datasets reading the original source data instead of data that was pre-processed in some way. That way the user doesn't have to perform the conversion manually.

vdumoulin:
Actually the dataset could just load the original file and never bother about the conversion: the `.amat` file is simply an ASCII file with 784 columns and either 10,000 or 50,000 rows, filled with zero or one 
characters.

The downside is that it takes longer to load (above 100 times slower) than if the `.amat` file was converted in a `.npy` file, and also uses more space on disk (15M for the `.amat` file vs. 7.5M for the converted `.npy` file.

Another option would be that the dataset only requires the `.amat` file to be present and does the conversion once, when the `.npy` file can't be found.

Not too long, it's just a matter of chan

bartvm:
In the age of terabyte datasets, I don't think 7.5M is barely worthy of our consideration :p

The loading takes 5s on my computer, which I would argue is actually not too bad. It's only done once or twice anyway, and it keeps the code nice and simple if we just read the original files. But having it as a fallback works for me too.

vdumoulin:
I'm not sure what you meant. Do you mean that you would prefer that we take the 5s hit and keep with the original `.amat` file to favor code simplicity?

bartvm:
Yes, but I might very well be unreasonable puritan there, so I'm fine with having `.amat` as a fallback as you suggested as well.

vdumoulin:
I pushed the changes, although manually clearing the cache in `binarized_mnist` may be required for the tests to pass.

vdumoulin:
No cache clearing is necessary after all. The tests pass!

bartvm:
Last comment, and then I'll merge, promise!

It's feasible that a user has read rights, but no access rights. Instead of saving the file, I would suggest raising a warning in the log saying that the fallback has happened.

bartvm:
*write rights

vdumoulin:
@bartvm Done!

Issue #14, opened 2015-02-23
Refactor lazy properties


bartvm:
Right now lazy properties are part of the `InMemoryDataset`, however, their principle is useful in other cases as well, see e.g. https://github.com/bartvm/fuel/pull/9#discussion_r25125787.

We need to rename lazy properties to something more descriptive, move it out of `InMemoryDataset`, and redo the documentation to highlight its more general applicability.

bartvm:
Fixed in #16

PR #16, opened 2015-02-23
Refactored lazy_properties

dmitriy-serdyuk:
Work is still in progress. Can you recommend a new name for `lazy_property_factory`? Or is it better to save this terminology?

### DIFF
--- a/fuel/datasets/base.py
+++ b/fuel/datasets/base.py
@@ -1,6 +1,4 @@
-import collections
 from abc import ABCMeta, abstractmethod
-
 from six import add_metaclass
 
 from fuel.streams import DataStream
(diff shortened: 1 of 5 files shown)


dmitriy-serdyuk:
This pull request conserns issue #14 .

fuel/datasets/base.py
```diff
@@ -263,73 +262,76 @@ def load(self):
...
         pass
 
-    @staticmethod
-    def lazy_properties(*lazy_properties):
-        r"""Decorator to assign lazy properties.
 
-        Used to assign "lazy properties" on :class:`InMemoryDataset`
-        classes.  Please see the documentation there for a discussion on
-        what lazy properties are and why they are needed.
+def do_not_pickle_properties(*lazy_properties):
```
bartvm: I would say `do_not_pickle_attributes` because for the user they are attributes, it's only through the internals of this decorator that they are made into properties.

bartvm:
Great! Looks good to me, but this has become a very general utility, so maybe move it to something like `fuel/utils.py` instead.

dmitriy-serdyuk:
@bartvm , I deleted check for `isinstance(attr, collections.Iterator)` because I'd like to serialize file handlers.

bartvm:
Instead, use `_iter` from the `picklable_itertools` package, this will make file handles picklable and actually save the state.

Issue #19, opened 2015-02-25
A Scheme to Iterate over Single Examples


rizar:
It is currently impossible to iterate over MNIST examples, not batches of examples.

Even `SequentialScheme(num_examples, 1)` does not fill this gap, because the requests it provides are singleton lists.

I would solve that by adding a new scheme, let's say `ExamplesScheme`, whose requests would be integers, not lists.

dwf:
Why would you want singletons rather than batch size 1?

rizar:
It find it much more convenient to write preprocessing for this format. Otherwise every preprocessing wrapper that I have must first unpack the batch and in the end pack the batch back.

bartvm:
Fixed in #20.

PR #20, opened 2015-02-26
WIP: Refactor

bartvm:
Bunch of things. Mostly:

- [x] Renamed `ContainerDataset` to the more correct `IterableDataset`
- [x] Introduced an `IndexableDataset` that can be used for e.g. NumPy arrays (such as MNIST, CIFAR-10) etc. this also factors out the `start` and `stop` arguments.
- [x] Make filenames and folders attributes, so that users can override them after unpickling in case files are in weird locations.
- [x] Introduce the convention that `num_examples` can be a number, `NaN` if we don't know, or `Inf` when the dataset is infinite
- [x] Introduced `SequentialExampleScheme` and `ShuffledExampleScheme`
- [x] Renamed `get_default_stream` to `get_example_stream`, which now per definition only returns single examples (which means it uses `iteration_scheme` equal to `None` for `IterableDataset` and `SequentialExampleScheme` for `IndexableDataset`)
- [x] Simplify `NGramStream` by working on examples instead of batches (current version was broken)
- [x] Add a bunch of tests
- [x] Remove `InMemoryDataset`, which was defunct after #16.
- [x] Enable Scrutinizer


fuel/datasets/base.py
```diff
@@ -28,10 +29,6 @@ class Dataset(object):
         'targets')`` for MNIST (regardless of which data the data stream
         actually requests). Any implementation of a dataset should set this
         attribute on the class (or at least before calling ``super``).
-    default_iteration_scheme : :class:`.IterationScheme`, optional
```
rizar: May I ask why do we get rid of this?

bartvm: I wanted to propose it for two reasons really: The main thing I'm trying to do here is to simplify the code. After just 3 datasets (MNIST, CIFAR-10 and binarized MNIST) it turned out they share way too much code, so I factored a bunch of things out. In the process, I realised that `default_iteration_scheme` is barely ever used, and writing `DataStream(dataset)` is actually shorter than `dataset.get_default_stream()` for the cases where there is no iteration scheme.

For the cases where there is an actual default iteration scheme, it has started to bother me that the choice is very arbitrary. Especially considering your recent ticket #19. Should these datasets return single examples by default? Batches of size 1? A minibatch of reasonable size, considering that is what we train on in most simple settings? The choice is so arbitrary I'd rather have it made explicit to the user. Having a default only makes sense if it works in 90% of the cases, but here it really doesn't.

rizar: Hold on: `DataStream(dataset)` works only for the `IterableDataset` currently, right? This is exactly the problem that I encountered: how can I simply iterate through the examples of a `IndexableDataset`?

I would argue that there is no such thing as the reasonable size, and `DataStream(dataset)` should always be the stream of examples. For this reason I would keep the default scheme feature but use the `ExampleScheme` as the default for the `IndexableDataset`.

bartvm: Yes, `DataStream(dataset)` only works on datasets which don't need an iteration scheme, that's the idea. Because I'm not convinced `ExampleScheme` is a reasonable default. I almost never use that, which is kind of my point: The sensible default completely depends on the person. If you are using lengthy pipelines that perform pre-processing on an example-level, `ExampleScheme` is a reasonable default. However, that default would crash our most basic tutorial simply because most people use minibatches and expect 2D matrices, which would make the current `SequentialScheme` a better default. Doesn't the fact that you are suggesting switching defaults already show that the choice is ambiguous?

Also, under the banner of "explicit is better than implicit" at least `DataStream(dataset, iteration_scheme=SequentialExamples(len(dataset))` makes clear what to expect while `dataset.get_default_stream()` leaves everyone guessing; will we get minibatches of size 1, individual examples, or something else? Changing the default later on will break things needlessly.

Then there's the last, but slightly OCD argument: The default scheme makes all the dataset files depend on the schemes. I like the idea of there not being any dependency.

rizar: > However, that default would crash our most basic tutorial simply because most people use minibatches and expect 2D matrices, which would make the current `SequentialScheme` a better default. 

Do our tutorial use default scheme? At least http://blocks.readthedocs.org/en/latest/tutorial.html does not. And there is no sense to have a tutorial without showing how to specify the batch size.

Counter-proposal: instead of `get_default_scheme` we provide `get_example_stream` which provides stream of examples. This is clearer than default, and this is the only reasonable default, sorry for this poor word-play. I just do not like that currently I can not just iterate over a given dataset, I have to check its type and create a data stream with a custom scheme.

rizar:
I am curious how we will ensure that this or another comparably large PR does not break Blocks.

fuel/datasets/base.py
bartvm: No, I meant that if you *would* use the default stream, something as simple as we present in the tutorial (training a simple classifier) doesn't work, showing that it's not a sensible default of an iteration scheme.

I see your point of not wanting to know what kind of dataset you're dealing with if you're just going to use something like `BatchedStream` afterwards. While it sounds like a technicality, I like `get_example_stream` better than `get_default_stream` because it clearly shows what to expect in return: A stream of individual examples. It doesn't pretend to be a "default", just a particular way of iterating.

rizar: Okay, I am 100% happy to have `get_example_stream` instead. Would you prefer to make it a part of this PR or of a later? 

```python
class ExampleScheme(BatchScheme):

    def __init__(self, num_examples):
        super(ExampleScheme, self).__init__(num_examples, 1)

    def get_request_iterator(self):
        return _iter(range(self.num_examples))
```

bartvm:
My idea was to make a PR in Blocks with the requirements.txt file pointing
to this branch in Fuel. That way, it will run all the tests with this
refactored version. Once everything passes, you remove the branch from
requirements.txt and merge both.

fuel/datasets/base.py
bartvm: I'll make it part of this one, uprooting everything anyway :/ hopefully for
the better.

rizar:
Right, but that we will need to do for every big PR...

What about running Blocks tests when testing Fuel? Yes they take 10 minutes, but it is not that much.

fuel/datasets/base.py
```diff
@@ -177,88 +170,124 @@ def filter_sources(self, data):
...
+                raise ValueError("sources have different lengths")
+        else:
+            self.indexables = [indexables]
+
+        def property_factory(source):
+            def property_(self):
+                return self.indexables[self.sources.index(source)]
+            return property_
+        for source in self.sources:
+            setattr(self.__class__, source, property(property_factory(source)))
```
rizar: I am still trying to understand what happens here, but it looks like a bit like a suicide: modifying the class every time an object is created??

bartvm:
Well, you have the same problem though. This version will only work with
corresponding changes in Blocks, so you would need to run the tests from
the relevant branch of Blocks.

fuel/datasets/base.py
bartvm: Right, that's WIP. I actually want to add these properties to the child class (e.g. MNIST), but haven't found an easy way of doing that yet.

rizar:
I do not understand what you mean. If we download Blocks when testing a Fuel branch and run their tests with this branch, we can it least be sure the proposed changes in Fuel are compatible with Blocks master. Just like currently we check that the proposed changes for Blocks for compatible with Fuel master.

bartvm:
No, because in this case I renamed e.g. ContainerDataset, so the master
branch of Blocks will crash against this PR. You need two parallel PRs (and
perform tests with these two) whenever you make these kind of changes.

bartvm:
Just to be clear: Running Blocks tests is still a good idea in cases where you *think* you don't need to make changes to Blocks (so there is no corresponding PR). I guess in that case it is easier to change the Blocks branch in Fuel and change that back before merging both simultaneously.

rizar:
Okay. The price for having these two repos separately is quite high, I should say.

bartvm:
I think it's fine, you just develop both simultaneously in two branches. It's just two `push` commands instead of one, and you need to make sure that Travis knows which blocks tests to pull.

fuel/datasets/base.py
bartvm: Thought about this, and it's actually not much of a problem. The worst thing that can happen is that it defines a property, and that some of the instances don't have a corresponding source. Currently that results in an `IndexError`, while it should really given an `AttributeError`, so I changed that.

bartvm:
@rizar @vdumoulin I think this is ready to merge, please let me know if you guys have any comments. I know it's a bit of a monster PR, but a lot of changes were related, so separate PRs would have been an endless rebasing nightmare... See opening post for a list of changes made.

vdumoulin:
I'll review this morning.

fuel/datasets/base.py
```diff
@@ -177,88 +190,136 @@ def filter_sources(self, data):
...
 
-        Any interaction with the outside world e.g. the file system,
-        database connections, servers, etc. should be done in this method.
-        This allows datasets to be pickled and unpickled, even in
-        environments where the original data is unavailable or has changed
-        position.
+    @property
+    def num_examples(self):
+        try:
+            return min(len(iterable) for iterable in self.iterables)
```
vdumoulin: Isn't there a problem if all iterables don't have the same length? They're supposed to represent different data sources for the same set of examples, no?

fuel/datasets/base.py
```diff
@@ -177,88 +190,136 @@ def filter_sources(self, data):
...
+        return izip(*iterators)
 
+    def get_data(self, state=None, request=None):
+        if state is None or request is not None:
+            raise ValueError
+        return next(state)
 
 
+class IndexableDataset(Dataset):
+    """Creates a dataset from a set of indexible containers.
```
vdumoulin: `indexible -> indexable`?

fuel/datasets/base.py
bartvm: Mmm, well, the thing is, you don't necessarily know the length of the iterables, because iterables aren't required to implement `__len__`, so there is no way of checking whether they are all the same length (they probably should be though!).

Given that, the iteration happens with `izip` in line 251, which means that *if* the user (wrongly) gives iterators of different lengths, the `izip` iterator would stop at the first one, hence the `min`.

I guess an alternative would be to use `izip_longest` which some sort of singleton padding value, and raise an error if that padding value actually ever appears.

fuel/datasets/base.py
bartvm: Fixed.

fuel/datasets/base.py
vdumoulin: I'm being picky, but _if_ all iterables implement `__len__`, shouldn't we be able to catch that here and raise an error if they don't have the same length?

bartvm: Yes, that particular case could be intercepted before iteration even begins. My suggestion would only be able to raise an error at the end of the first epoch, so maybe it's better to just silently ignore that and only check at the beginning.

vdumoulin: Sounds good.

vdumoulin:
The PR looks good to me.

fuel/datasets/base.py
bartvm: Okay, implemented! Had to fix a few tests which were actually passing variable length iterables, so it was a good call to check.

fuel/datasets/base.py
```diff
@@ -177,88 +190,142 @@ def filter_sources(self, data):
...
+            self.num_examples)
+
+        def property_factory(source):
+            def property_(self):
+                if source not in self.sources:
+                    raise AttributeError
+                return self.indexables[self.sources.index(source)]
+            return property_
+        for source in self.sources:
+            setattr(self.__class__, source, property(property_factory(source)))
```
rizar: I do not think this is beautiful. Instead you could declare a metaclass and create properties for `provided_sources` just like you create `Application`'s for a brick.

bartvm: The `provides_sources` isn't defined until after the call to `__init__` (see lines 296, 298) so a metaclass would have to attach the properties to the class at exactly the same point as I do now (i.e. after calling `__init__` from the metaclass's `__call__` method). So in effect you would do *exactly* the same thing, the only difference being that the code is hidden in a metaclass instead of being accessible to users in the code.

bartvm:
@rizar Can I go ahead and merge this soon? Because I'd like to get started on some other things, but this PR is holding things back a bit. Overall I think it's an improvement (better test coverage, example schemes, way less duplication of code, etc.) Let me know if you're not satisfied with the properties and I can look at it again.

fuel/datasets/base.py
rizar: Okay, I see. 

But I have another idea: we can handle these requests in `__getattr__`. I think it is more natural than changing the class in the constructor.

bartvm: That works too. I'll make the change.

rizar:
Go ahead and merge, because it will take me some time to properly review this code. That's a lot of refactoring.

fuel/datasets/mnist.py
```diff
@@ -45,50 +39,47 @@ class MNIST(InMemoryDataset):
...
-                                             numpy.newaxis]
-        self.features = x
-        self.targets = y
-
-    def get_data(self, state=None, request=None):
-        if state is not None:
-            raise ValueError("MNIST does not have a state")
-        return self.filter_sources((self.features[request],
-                                    self.targets[request]))
+        self.indexables = [data[self.start:self.stop] for source, data
```
rizar: The content of `load` is effectively boilerplate code with the only varying part being the name of the loading method, e.g. `_load_mnist` or `_load_cifar`. I think there should be a default implementation for `load` in the `IndexableDataset`.

bartvm: It shouldn't go in `IndexableDataset`; that class is too general IMHO. `IndexableDataset` can also just be used with things like `IndexableDataset(range(100))`, in which case a default implementation of load doesn't make sense. I'm also afraid that a user might subclass it, just add a lazy property (which doesn't necessarily have to be `indexables`) and suddenly gets errors because it's trying to call a method that doesn't exist.

I do agree that this could probably be filtered out, but there'd need to be another base class, `IndexableFromDisk` or something like that maybe?

rizar: I like the idea of adding another class in the hierarchy for indexable datasets fitting into memory. `IndexableFromDisk` might be confused with some weird index-based fetching from outer memory though. 
Maybe `IndexableLoadableDataset`? The name can be arbitrary long as this is not front-end. 

Also I would rename the current `load` name into let's say `reload`, and use `load` as a part of the `IndexableLoadable` interface.

PR #26, opened 2015-02-27
WIP: Allow custom index range for IterationSchemes

arasmus:
A feature I need. If the iteration scheme is using only a subset of the indices from the datastream, I'd like to define the subset and not always assume I want to use the first ```num_examples```.

This requirement comes from the problem that if I artificially shrink my dataset with ```num_examples```, I might get a unbalanced number of different classes. Imagine someone giving ```num_examples=10```. It is unlikely, that all MNIST classes are represented there.

Another way to implement this would be to have a DataStreamWrapper that reorders indices or DataSet that reorders them, but this approach results in the smallest diff, I think.

Alternative implementation: ```num_examples``` parameter could also made to accept either an integer or an iterable in which case it would be renamed to, say, ```examples``` and code would like:

```self.indices = xrange(examples) if type(examples) is int else examples```

### DIFF
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -1,4 +1,5 @@
 from abc import ABCMeta, abstractmethod
+from collections import Iterable
 
 import numpy
 from picklable_itertools import chain, repeat, imap, iter_
@@ -58,9 +59,28 @@ class BatchScheme(IterationScheme):
     to be kept in memory, like MNIST) we can provide slices or lists of
     labels to the dataset.
 
+    Parameters
+    ----------
+    examples : int or list
+        Defines which examples from the dataset are iterated.
+        If list, its items are the indices of examples.
+        If an integer, it will use that many examples from the beginning
+        of the dataset, i.e. it is interpreted as range(examples)
+    batch_size : int
+        The request iterator will return slices or list of indices in
+        batches of size `batch_size` until the end of `examples` is
+        reached.
+        Note that this means that the last batch size returned could be
+        smaller than `batch_size`. If you want to ensure all batches are
+        of equal size, then ensure len(`examples`) or `examples` is a
+        multiple of `batch_size`.
+
     """
-    def __init__(self, num_examples, batch_size):
-        self.num_examples = num_examples
+    def __init__(self, examples, batch_size):
+        if isinstance(examples, Iterable):
+            self.indices = examples
+        else:
+            self.indices = xrange(examples)
         self.batch_size = batch_size
 
 
@@ -72,8 +92,11 @@ class IndexScheme(IterationScheme):
     but where we want to return single examples instead of batches.
 
     """
-    def __init__(self, num_examples):
-        self.num_examples = num_examples
+    def __init__(self, examples):
+        if isinstance(examples, Iterable):
+            self.indices = examples
+        else:
+            self.indices = xrange(examples)
 
 
 class ConstantScheme(BatchSizeScheme):
@@ -124,8 +147,7 @@ class SequentialScheme(BatchScheme):
 
     """
     def get_request_iterator(self):
-        return imap(list, partition_all(self.batch_size,
-                                        xrange(self.num_examples)))
+        return imap(list, partition_all(self.batch_size, self.indices))
 
 
 class ShuffledScheme(BatchScheme):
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by rizar** [comment]
Overall it looks good, but this is very much core code and it would be nice to cover it with tests.

**Comment by bartvm** [comment]
LGTM to, but agreed on the tests. Also, docstring! :)

I prefer this over `num_examples` being allowed to be an iterable, the semantics of that are a bit more obscure.

**Comment by bartvm** [comment]
We should give the new `ExampleScheme`s the same features btw, but we can do that in another PR.

**Comment by arasmus** [comment]
What bothers me is that if one gives ```indices```, still ```num_examples``` has to be given and they can be different lengths. Some options to deal with this:
1. Now I have an assertion ensuring they match, which is not perfect.
2. Ignore ```num_examples``` if ```indices``` is given. This can be confusing.
3. Allow ```num_examples``` to be an iterable.
4. Only take as many indices as indicated by ```num_examples```. E.g. ```self.indices=indices[:num_examples]```

What do you think?

I'll add tests and docstrings.

**Comment by bartvm** [comment]
I see your point. I agree that ignoring one argument when another is given isn't very standard. However, having two optional arguments of which only can be passed is something fairly common. In that case, it would become something like
```python
def __init__(batch_size, num_examples=None, indices=None)
    if (num_examples and indices) or (not num_examples and not indices):
        raise ValueError("need to provide exactly one of `num_examples` and `indices`")
    ...
```
Maybe that's an option? Otherwise maybe renaming `num_examples` to `examples` and letting it be an iterable isn't such a bad idea. NumPy does that in many cases, allowing you to pass either an integer or a vector, in which case the integer is interpreted as `np.arange(val)`, which is kind of the same here.

arasmus:
I decided to stick with this to minimize interface changes. No example code need to be modified. Changing the order of ```num_examples``` and ```batch_size``` like you suggested would've required a change to those files.

But now I noticed that tutorial.rst is actually using the old argument name, so this wasn't that bullet proof after all? So I quickly made this to fix tutorials https://github.com/bartvm/blocks/pull/390

fuel/schemes.py
```diff
@@ -58,9 +58,25 @@ class BatchScheme(IterationScheme):
...
+        Note that this means that the last batch size returned could be
+        smaller than `batch_size`. If you want to ensure all batches are
+        of equal size, then ensure len(`examples`) or `examples` is a
+        multiple of `batch_size`.
+
     """
-    def __init__(self, num_examples, batch_size):
-        self.num_examples = num_examples
+    def __init__(self, examples, batch_size):
+        self.indices = xrange(examples) if type(examples) is int else examples
```
bartvm: We should probably allow for duck typing instead of checking explicitly if the type is `int`. There's two options I think: Either use
```python
from numbers import Integral
isinstance(examples, Integral)
```
or
```
from collections import Iterable
not isinstance(x, Iterable)
```

bartvm:
It's okay to break the interface :) While working towards 0.1 you can break at will, so that we have a tried and tested interface by then. Otherwise we end up with legacy code before we even get there.

arasmus:
Now the type changing is more robust allowing duck typing.

bartvm:
Great, seems like there's a lingering `num_examples` in `test_serialization.py`, but otherwise it LGTM.

arasmus:
Now everything should be alright!

bartvm:
Great, thanks!

dwf:
Does supporting this kind of thing instead of `start`/`stop` in the dataset classes themselves make sense?

bartvm:
It would make a difference for memory use though. I can imagine datasets using the `start` and `stop` arguments to load only a subset into memory, whereas having them in the request iterator only would load everything into memory but just never request some of them.

arasmus:
If you look at my comment that started the thread, the main reason behind such thing was ordering of indices or samples. Perhaps Dataset could be made to support user-requested ordering instead?

dwf:
@bartvm I agree that it would probably make a difference. However, being able to pass an `xrange`, or even a list of indices, to the dataset seems like a more flexible generic interface, even if not every dataset chooses to support every sort of selector that you can dream up.

dwf:
s/probably/definitely/

bartvm:
Sorry, I misread your suggestion! I thought you wanted to move the `start`/`stop` behaviour from the dataset to the request iterator.

The general idea sounds reasonable, but probably should be carefully implemented. For in-memory indexable datasets it might worth performing the shuffling in memory once, while for e.g. HDF5 you will probably want to translate each `get_data` request to the new shuffled indices e.g. `self.data[self.indices[request]]`.

PR #28, opened 2015-02-27
The big rename (Transformers). Fixes #23.

### LINKED ISSUE #23: Rename `DataStreamWrapper` to `Transformer`
`DataStreamWrapper` made sense when we were designing this, because the point was that data streams and data streams wrappers share the same interface. (Initially we actually tried making them one and the same class, but that ended up not happening, and now they just share a base class.)

The problem with `DataStreamWrapper` is that it's a bit obscure of a name, and I think it actually confuses people sometimes, because they're not sure what the difference between `DataStream` and `DataStreamWrapper` is. I actually like Pylearn2's `Transformer` better. Another alternative is `Preprocessor`, but that sounds too much like just doing whitening, etc. and not routine transformations like sorting, merging, etc.

So the terminology then becomes: Instantiate a **dataset**, and create a **data stream** that reads from this dataset (potentially using a particular iteration scheme). Afterwards you can apply a series of **transformations** on the data stream (some transformers can support the `iteration_scheme` argument, some don't, but the `get_epoch_iterator` interface is the same).

Thoughts? Here I go randomly tagging people for opinions again: @rizar @lamblin @vdumoulin @laurent-dinh @dwf @pbrakel @jbornschein

bartvm:
This renames `DataStreamWrapper`s to `Transformer` as discussed in #23.

This also means that a lot of the names we had (`MappingStream`, `PaddingStream`, etc.) seemed wrong now. After some pondering, I figured that maybe it was just best to remove all those redundant suffixes. From the module they are in, it's clear that they are wrappers/streams/transformers, so just `Mapping`, `Padding`, etc. should do just fine IMHO.

### DIFF
--- a/fuel/datasets/base.py
+++ b/fuel/datasets/base.py
@@ -218,7 +218,7 @@ class IterableDataset(Dataset):
     decorator.
 
     To iterate over a container in batches, combine this dataset with the
-    :class:`BatchDataStream` data stream.
+    :class:`Batch` data stream.
 
     """
     example_iteration_scheme = None
--- a/fuel/iterator.py
+++ b/fuel/iterator.py
@@ -6,7 +6,7 @@ class DataIterator(six.Iterator):
 
     Parameters
     ----------
-    data_stream : :class:`DataStream` or :class:`DataStreamWrapper`
+    data_stream : :class:`DataStream` or :class:`Transformer`
         The data stream over which to iterate.
     request_iterator : iterator
         An iterator which returns the request to pass to the data stream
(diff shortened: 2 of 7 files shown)


vdumoulin:
I've reviewed your PR and confirm it's ready to merge once the tests pass (my approval probably doesn't add much to the reviewing process since it's mostly renaming and refactoring, but hey, why not play by the rules?)

bartvm:
I guess that renames break everyone's code, so I guess it's good to try and make sure that they don't break more than that :p 

@rizar Are you okay with dropping all the `DataStream` prefixes and suffixes as well?

rizar:
I hated these prefixes and suffixes from the beginning, so I am happy not to have them.

PR #30, opened 2015-02-27
_iter -> iter_, use picklable_itertools.xrange

dwf:


bartvm:
Sorry, merge conflicts because I just merged #29

dwf:
Woops. Merge conflicts. Hold on.

fuel/schemes.py
```diff
@@ -1,10 +1,9 @@
 from abc import ABCMeta, abstractmethod
 
 import numpy
-from picklable_itertools import chain, repeat, imap, _iter
+from picklable_itertools import chain, repeat, imap, xrange, iter_
```
bartvm: What does `xrange` from `picklable_itertools` offer over the built-in? Built-in `xrange` is picklable, right?

dwf: Two things: 

* If you iterate on it, you needn't call `iter_()`; you'll get a `range_iterator` automatically.
* Some Python 3 features (`.count()`, `.index()`) are available in this implementation that aren't generally.

#29 makes adding it here irrelevant, since `partition_all`, etc. call `iter_` internally, so I can remove it if you like. It does lead to terser code in places where you'd need an additional `iter_` call though.

bartvm: Mm... I'm undecided. The thing is, the main selling point seems to be the ability to use the `for ... in xrange(...)` construct. The thing is, you would never pickle in the middle of a `for`-loop, since the iterator is implicit in that case it won't be serialized, and you would lose your position within the loop. Also, if someone ends up using this for a very large number of examples for very short iterations, the additional overhead might actually become noticeable (or maybe not?). On the other hand, maybe it's better to stay on the safe side and use it? I really don't know, so fine either way.

dwf: Yes, now that I look at it, the `_iter` call in the code that #29 replaced was in fact entirely unnecessary. I'll ditch that one for now, but it's there if we need it (e.g. for the methods).

dwf:
Travis failure due to transformers rename. Force-pushed a rebase.

bartvm:
Haha, sorry, this is turning into a really complicated 5-line PR :p

dwf:
Looks like it passed.

PR #32, opened 2015-03-01
Speed up TextFile dataset and Merge transformer

bartvm:
Oops, the `Merge` transformer was converting the streams to lists in memory, making it unworkable for large streams. This should fix that, and also speeds up the `TextFile` dataset a little bit by extending lists instead of creating new, concatenated ones.


bartvm:
Depends on this bug fix being merged into picklable itertools: https://github.com/dwf/picklable_itertools/pull/38

Issue #7, opened 2015-02-19
Wishlist: "server" process to do preprocessing in a separate thread


vdumoulin:
A bit like Bokeh, it would be great to have the option of launching a Fuel server which can do preprocessing in a separate thread.

I imagine a scenario like this: you'd ask the Fuel server to maintain a queue of 10 preprocessed batches of examples from some dataset (_e.g._ ZCA-whitened CIFAR10 images) according to some iteration scheme, and your client application (_e.g._ Blocks, Pylearn2) could "consume" this data while Fuel replenishes the queue in parallel.

bartvm:
I think this is a great idea, and something I've been thinking about as well. The main thing I ran into: What kind of interprocess communication to use that is fast enough not to turn into a bottleneck as well as user-friendly. I haven't worked with a lot of them, but I guess that options are:

* Sockets
* (Named) pipes
* Shared memory

Then there's a loong list of libraries that we could use to manage any of these. It seems silly to program all of this from scratch; it must be a problem that has been solved a thousand times.

So my fist suggestion that I think is worth looking into is limiting the server to transferring NumPy arrays, and using ZeroMQ (PyZMQ) for transferring the data. A few reasons:

* Although sockets are theoretically not as fast as shared memory, they're a lot more user-friendly
* Using sockets could potentially allow (eventually) for multiple clients, eventually allowing us to e.g. pass data to multiple clients on a cluster using
* ZeroMQ is an excellent library
* It seems to be a popular solution (e.g. [`python-matlab-bridge`](https://github.com/arokem/python-matlab-bridge) and [`zmqnumpy`](https://pypi.python.org/pypi/zmqnumpy/0.1)) and it [seems straightforward](https://zeromq.github.io/pyzmq/serialization.html) to get started with.

I imagine then that we would have a server module which basically takes as configuration an entirely configured datastream. It reads batches from this data stream into a pool as you said. Each query made to the sever results in either (a) a batch of data being returned or (b) a `StopIteration` message (signaling the end of an epoch). On the receiving side, the model simply takes as argument a `ServerDataStream` which is in charge of replicating the behaviour of an iterator based on the data it receives from the server. Data could optionally be compressed with e.g. [`bloscpack`](https://github.com/esc/bloscpack#numpy) before transferring.

[...]

PR #33, opened 2015-03-01
Add multiprocessing stream from @lukemetz

bartvm:
Haven't benchmarked it, but it seems to work and @lukemetz has reported speedups, so we might as well put it in here to address #7 for the cases where we don't need a super-optimized solution.

I took @lukemetz's original code and added some documentation. Instead of relying on `__del__` to exit the process I turned it into a daemon, which means it will just get killed as soon as the main process stops. I'm not 100% sure if this could cause problems in case file handles are left open or something, but I think it's safer than relying on `__del__`.

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -1,4 +1,5 @@
 from abc import ABCMeta
+from multiprocessing import Process, Queue
 
 import numpy
 from picklable_itertools import chain, ifilter, izip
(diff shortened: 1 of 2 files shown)


fuel/transformers/__init__.py
```diff
@@ -370,3 +371,61 @@ def get_epoch_iterator(self, **kwargs):
         batches = chain(*izip(*[data_stream.get_epoch_iterator()
                                 for data_stream in self.data_streams]))
         return partition(len(self.sources), chain(*batches))
+
+
+class BackgroudProcess(object):
```
dwf: Typo.

fuel/transformers/__init__.py
```diff
@@ -370,3 +371,61 @@ def get_epoch_iterator(self, **kwargs):
...
+    """A background process that reads batches and stores them in a queue."""
```
dwf: Docstring needs work.

bartvm:
Just putting this here for future reference: http://eli.thegreenplace.net/2009/06/12/safely-using-destructors-in-python

Switching to `__del__` might be a good idea, but need to check to make sure all the stuff about circular references and such holds. Also, don't want to get the child-process stuck because it's stuck at a blocking call to `queue.put`.

PR #35, opened 2015-03-03
SVHN format 2

laurent-dinh:


vdumoulin:
@laurent-dinh  Things have stabilized a little concerning external datasets. If you're still interested in adding this dataset to Fuel, have a look at PR #75, which details how to proceed.

vdumoulin:
Since SVHN is now part of the built-in datasets, I'll close this PR.

PR #37, opened 2015-03-04
Added Unpack Transformer

Thrandis:
Hey Bart,
Here is my CCW for this week! Dima had already implemented this class, so I reused his code. It is not the solution you proposed, so I hope it will be ok!
Let me know if I need to change something.
César

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -258,6 +258,33 @@ class Batch(Transformer):
         return tuple(numpy.asarray(source_data) for source_data in data)
 
 
+class Unpack(Transformer):
+    """Unpacks batches to compose a stream of examples.
+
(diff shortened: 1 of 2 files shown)


bartvm:
Did you try my suggested method? The reason I prefer that one is because it results in (a) less code and (b) less function calls. It should just be

```python
class Unpack(Transformer):
    def get_epoch_iterator(self):
        return chain.from_iterable(self.data_stream.get_epoch_iterator())
```

Thrandis:
Yes, but for some reason it returned only one example form each batch!

bartvm:
Ah, that makes sense actually. You would end up doing something more complicate I guess, like `chain.from_iterable(imap(chain.from_iterable, imap(izip, self.data_stream.get_epoch_iterator())))`. No idea whether that actually works, but it gets so illegible we might as well stick with this approach :)

PR #40, opened 2015-03-06
change DataIterator to work on get_example and get_batch and add methods...

mducoffe:
... and batch attributes to Transformer object

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -72,13 +72,11 @@ class Transformer(AbstractDataStream):
 
     def get_data_from_example(self, request=None):
         raise NotImplementedError(str(type(self)) + 
-        "does not have an example method")
+            "does not have an example method")
 
-    def get_data_from_batch(self,request=None):
-        raise NotImplementedError(str(type(self)) + 
-        "does not have a batch input method")
-
-
+    def get_data_from_batch(self, request=None):
+        raise NotImplementedError(str(type(self)) +
+            "does not have a batch input method")
 
 
 class Mapping(Transformer):
@@ -495,4 +493,3 @@ class MultiProcessing(Transformer):
         if data == StopIteration:
             raise StopIteration
         return data
-
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -22,10 +22,14 @@ class Transformer(AbstractDataStream):
         this attribute. Use it to access data from the wrapped data stream
         by calling ``next(self.child_epoch_iterator)``.
 
+    batch : boolean
+        Determine wheter the model is working on examples or on batches
+
     """
-    def __init__(self, data_stream, **kwargs):
+    def __init__(self, data_stream, batch_input=False,  **kwargs):
         super(Transformer, self).__init__(**kwargs)
         self.data_stream = data_stream
+        self.batch_input = batch_input
 
     @property
     def sources(self):
@@ -60,6 +64,19 @@ class Transformer(AbstractDataStream):
         self.child_epoch_iterator = self.data_stream.get_epoch_iterator()
         return super(Transformer, self).get_epoch_iterator(**kwargs)
 
+    def get_data(self, request=None):
+        if self.batch_input:
+            return self.get_data_from_example(request)
+        else:
+            return self.get_data_from_batch(request)
+
+    def get_data_from_example(self, request=None):
+        raise NotImplementedError(str(type(self))
+                                    + " does not implement get_data_from_example")
+
+    def get_data_from_batch(self, request=None):
+        raise NotImplementedError(str(type(self))
+                                    + " does not implement get_data_from_batch")
 
 class Mapping(Transformer):
     """Applies a mapping to the data of the wrapped data stream.
@@ -164,7 +181,7 @@ class Cache(Transformer):
             data_stream, iteration_scheme=iteration_scheme)
         self.cache = [[] for _ in self.sources]
 
-    def get_data(self, request=None):
+    def get_data_from_example(self, request=None):
         if request > len(self.cache[0]):
             self._cache()
         data = []
@@ -237,7 +254,7 @@ class Batch(Transformer):
             data_stream, iteration_scheme=iteration_scheme)
         self.strictness = strictness
 
-    def get_data(self, request=None):
+    def get_data_from_example(self, request=None):
         """Get data from the dataset."""
         if request is None:
             raise ValueError
@@ -271,10 +288,10 @@ class Unpack(Transformer):
 
     """
     def __init__(self, data_stream):
-        super(Unpack, self).__init__(data_stream)
+        super(Unpack, self).__init__(data_stream, batch_input=True)
         self.data = None
 
-    def get_data(self, request=None):
+    def get_data_from_batch(self, request=None):
         if not self.data:
             data = next(self.child_epoch_iterator)
             self.data = izip(*data)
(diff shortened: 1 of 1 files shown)


mducoffe:
Just to know if the code was okay. Also do I have to call get_epoch_iterator on every transformer class with example or batch stream for tests ? (Have also to correct indentation error)

fuel/iterator.py
```diff
@@ -23,9 +23,15 @@ def __iter__(self):
 
     def __next__(self):
         if self.request_iterator is not None:
-            data = self.data_stream.get_data(next(self.request_iterator))
+            if self.data_stream.batch :
```
bartvm: Mmmm, maybe it's better to move this logic to the `Transformer` base class. That way child classes can override it if necessary and don't need to duplicate methods if the implementation for batch/example is the same.

fuel/transformers/__init__.py
```diff
@@ -75,8 +77,8 @@ class Mapping(Transformer):
         data under source names `add_sources`.
 
     """
-    def __init__(self, data_stream, mapping, add_sources=None):
-        super(Mapping, self).__init__(data_stream)
+    def __init__(self, data_stream, mapping, batch=False, add_sources=None):
```
bartvm: It's probably better to use `...add_sources=None, **kwargs)` and then do `super(Mapping, self).__init__(data_stream, **kwargs)`.

fuel/transformers/__init__.py
```diff
@@ -257,6 +276,44 @@ def get_data(self, request=None):
...
+        try:
+            return next(self.data)
+        except StopIteration:
+            self.data = None
+            return self.get_data()
+
+    def get_example(self, request=None):
+        return self.get_data(request)
+
+    def get_batch(self,request=None):
```
bartvm: Instead of doing this for each transformer, implement a `get_batch` and `get_example` method that raises a `NotImplementedError` in the `Transformer` base class.

bartvm:
There are a few transformers which can share the implementation of their `get_example` and `get_batch` method, so here's how I would propose doing it:
* Implement a `get_data` method on the `Transformer` base class that calls `get_example` and `get_batch` (like the logic you implemented in the `DataIterator` now). Create two `get_example` and `get_batch` methods which raise a `NotImplementedError` when called
* Transformers that have the same logic for batch and example just override the `get_data` method
* Transformers for which the methods are different, or which only implement one, override `get_example`/`get_batch` instead.

Hope that was clear

fuel/transformers/__init__.py
bartvm: No need to add the keyword argument here. It has a default after all.

fuel/transformers/__init__.py
```diff
@@ -75,8 +91,8 @@ class Mapping(Transformer):
         data under source names `add_sources`.
 
     """
-    def __init__(self, data_stream, mapping, add_sources=None):
-        super(Mapping, self).__init__(data_stream)
+    def __init__(self, data_stream, mapping, batch=False, add_sources=None):
+        super(Mapping, self).__init__(data_stream, batch)
```
bartvm: So you can remove it here too.

fuel/transformers/__init__.py
```diff
@@ -97,8 +113,8 @@ def get_data(self, request=None):
 
 class ForceFloatX(Transformer):
     """Force all floating point numpy arrays to be floatX."""
-    def __init__(self, data_stream):
-        super(ForceFloatX, self).__init__(data_stream)
+    def __init__(self, data_stream, batch=False):
```
bartvm: Same as above. It's probably better if it crashes when you try to pass `batch` anyway, because it won't have any effect.

fuel/transformers/__init__.py
```diff
@@ -258,6 +274,30 @@ def get_data(self, request=None):
...
+class Unpack(Transformer):
+    """Unpacks batches to compose a stream of examples.
+    This class is the inverse of the Batch class: it turns a minibatch into
+    a stream of examples.
+    Parameters
+    ----------
+    data_stream : :class:`AbstractDataStream` instance
+        The data stream to unpack
+    """
+    def __init__(self, data_stream, batch=False):
```
bartvm: No need to add it to the signature here either, because it won't make sense for the user to pass `batch=True`

bartvm:
Seems like there's a merge conflict, so you'll need to rebase. Let me know if you need some help doing that!

mducoffe:
@bartvm it is eventually done !

mducoffe:
@bartvm do u know what is coverage test in the github tests ?

bartvm:
Coverage is the number of lines of code that are run by unit tests divided by the total number of lines of code. If it decreases it means that this percentage dropped. In this case it's because you added 9 new lines of code to `transformers/__init__.py`, but only 7 of those are tested, so the coverage decreases slightly. There's no need to worry about this too much, it's mostly to remind people that their code needs to be tested, and if the drop is very large then it is a sign that tests need to be written. You can actually click on the details and see exactly which line is currently untested.

mducoffe:
@bartvm is there any way to see which lines are not covered by the tests ? (sources are not available)

bartvm:
Not sure, never seen that before! It might be because of the merge conflicts earlier. You might want to try rebasing?

fuel/transformers/__init__.py
```diff
@@ -22,10 +22,14 @@ class Transformer(AbstractDataStream):
         this attribute. Use it to access data from the wrapped data stream
         by calling ``next(self.child_epoch_iterator)``.
 
+    batch : boolean
```
bartvm: `batch_input` instead of `batch`, and there should be no new line between parameter descriptions.

fuel/transformers/__init__.py
```diff
@@ -60,6 +64,24 @@ def get_epoch_iterator(self, **kwargs):
...
 
+    def get_data(self, request=None):
+        if self.batch_input:
+            return self.get_data_from_batch(request)
+        else:
+            return self.get_data_from_example(request)
+
+    def get_data_from_example(self, request=None):
+        raise NotImplementedError(
+            str(type(self)) +
```
bartvm: We generally prefer to use formatting instead of string concatenation, it looks a bit cleaner IMHO. Something slightly more informative would be good too e.g. ```"`{}` does not support examples as inputs, but `batch_input` was set to `False`".format(type(self))```

fuel/transformers/__init__.py
```diff
@@ -164,7 +186,7 @@ def __init__(self, data_stream, iteration_scheme):
             data_stream, iteration_scheme=iteration_scheme)
         self.cache = [[] for _ in self.sources]
 
-    def get_data(self, request=None):
+    def get_data_from_example(self, request=None):
```
bartvm: This should be `get_data_from_batch` (don't forget to pass `batch_input=True` to `super`).

fuel/transformers/__init__.py
```diff
@@ -261,20 +283,23 @@ def get_data(self, request=None):
 class Unpack(Transformer):
     """Unpacks batches to compose a stream of examples.
 
+
```
bartvm: Not sure why new lines were added here, but there should be only one.

fuel/transformers/__init__.py
```diff
@@ -459,13 +484,13 @@ class MultiProcessing(Transformer):
...
     def __init__(self, data_stream, max_store=100):
-        super(MultiProcessing, self).__init__(data_stream)
+        super(MultiProcessing, self).__init__(data_stream, batch_input=True)
         self.background = BackgroundProcess(data_stream, max_store)
         self.proc = Process(target=self.background.main)
         self.proc.daemon = True
         self.proc.start()
 
-    def get_data(self, request=None):
+    def get_data_from_batch(self, request=None):
```
bartvm: Since this one just passes on data, I think its agnostic to whether the input is a batch or single example. You can just leave the `get_data` method here.

bartvm:
Wait, I just realised you're working in #45 now, so I'll close this one then?

mducoffe:
yes please. And by the way thanks for the dropout on blocks, it is really
convenient to use with the computation graph

2015-03-20 13:48 GMT-04:00 Bart van Merriënboer <notifications@github.com>:

PR #43, opened 2015-03-09
[WIP] h5py dataset

vdumoulin:


fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,127 @@ def get_data(self, state=None, request=None):
...
+
+@do_not_pickle_attributes('data_sources')
+class H5PYDataset(Dataset):
+    """An h5py-fueled HDF5 dataset.
+
+    This dataset class assumes a particular file layout:
+
+    * Data sources reside in the root group, and their names define the
+      source names.
+    * The dataset is not explicitly split. Instead, splits are defined as
```
dwf: Maybe I'm missing something but why is this desirable?

bartvm: https://github.com/bartvm/fuel/issues/41#issuecomment-77774481

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,127 @@ def get_data(self, state=None, request=None):
...
+      shape (2,), with the first element being the starting point
+      (inclusive) of the split and the last element being the stopping
+      point (exclusive) of the split.
+
+    The `which_set`, `start` and `stop` parameters work together in the
+    following way:
+
+    * `which_set` is resolved first. If it is `None`, the whole dataset is
+      used.
+    * `start` and `stop` define a slice *within the context of*
```
dwf: I've been thinking of proposing that we start using range objects instead of 'start' and 'stop', since a) one less function parameter and b) allows for the possibility of taking step > 1. (Even further, we could one day accept either a range object or a list of indices, much as if you were indexing a NumPy array.) @bartvm @vdumoulin ?

fuel/datasets/hdf5.py
dwf: Alright, well, I don't totally understand the corpora argument and I can't think of when you'd ever want to do this for a dataset like MNIST. If we go this direction, we should test this _very_ thoroughly. Nothing more frustrating then realizing you've accidentally trained on the test data and your results are bullshit. :)

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,127 @@ def get_data(self, state=None, request=None):
...
+    path : str
+        Path to the HDF5 file.
+    which_set : str, optional
+        Name of the root group attribute containing the split information.
+        Defaults to `None`, in which case the whole dataset is used.
+    start : int
+        Start index *within the split*.
+    stop : int
+        Stop index *within the split*.
+    driver : str, optional
```
dwf: Missing load_in_memory

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,127 @@ def get_data(self, state=None, request=None):
...
+        self._start = start if self.start is None else start + self.start
+        self._stop = stop if self.stop is None else start + self.stop
+        self.num_examples = self._stop - self._start
+        self.data_sources = ([data_source[self._start: self._stop] for
+                              data_source in handle.values()]
+                             if self.load_in_memory else None)
+        self._out_of_memory_close(handle)
+
+    def open(self):
+        return (self._in_memory_open() if self.load_in_memory
```
dwf: Is there a reason for not just returning None here? WHy have in_memory_open and in_memory_close at all?

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,127 @@ def get_data(self, state=None, request=None):
...
+                filter(lambda x: x.name == self.path, self.ref_counts.keys()))
+        except StopIteration:
+            return self.path
+
+    def load(self):
+        handle = self._out_of_memory_open()
+        self.provides_sources = list(handle.keys())
+        shapes = [data_source.shape for data_source in handle.values()]
+        if any(s[0] != shapes[0][0] for s in shapes):
+            raise ValueError("sources have different lenghts")
```
dwf: typo

fuel/datasets/hdf5.py
bartvm: The corpora argument basically just comes down to this: WMT15 data consists of several corpora, and even the test set consists of multiple different corpora. It would be nice if you can do `WMT15('europarl')`, `WMT15('newstest1213')`, `WMT15('test')`, etc. Here `WMT15('test')` is a combination of `WMT15('newstest1213')` as well as `WMT15('newstest14')`, but depending on whether you're combining with an SMT system and need a separate set to tune, you might want to use either `test` or `newstest1213` as your test set.

fuel/datasets/hdf5.py
vdumoulin: Implementation-wise this may make things more complicated when reading from disk, but I'm not against it.

fuel/datasets/hdf5.py
dwf: Got it, although this will only work here if all of the atomic sets can be
arranged such that any concatenated set you might like will be contiguous,
right?

bartvm: For now, but to quote myself:

> If we ever run into a case where we want splits that are non-contiguous, we could use something like h5file.attrs['alternative_train'] = '/split/alternative_train' and store lists of indices in /split/alternative_train, /split/alternative_test, etc. The logic is then:

> * MNIST() loads the HDF5 file in its entirety
* MNIST('train') checks the split = h5file.attrs['train'] attribute and loads
    * if it finds an array it slices h5file[source][split[0]:split[1]]
    * if it finds a string, it loads h5file[source][h5file[split]]

fuel/datasets/hdf5.py
bartvm: Mm, maybe we should accept `slice` objects instead? That's kind of what they're made for anyway. The main problem I can think of with `range` is that in Python 2 it creates a list in memory. You could use `six.moves.xrange` instead, but that's not as user-friendly I guess.

dwf: slice objects actually sound good to me, yes.

dwf: If we don't want to support non-unit steps yet we can just assert that `slice.step` is None or 1 as a stop-gap.

vdumoulin: Sounds good, will implement.

fuel/datasets/hdf5.py
vdumoulin: Symmetry? More seriously, I don't have any solid reason, I'll refactor that.

vdumoulin:
@dwf @bartvm I addressed all comments.

Regarding the split concatenation, I think it's pretty straightforward to implement unit tests that verify that the splits you get out of the HDF5 version of the dataset and the splits you get when you download the data are the same.

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,124 @@ def get_data(self, state=None, request=None):
...
+        self.subset = subset if subset else slice(None, None, None)
+        self.load_in_memory = load_in_memory
+        self.driver = driver
+        self.load()
+        super(H5PYDataset, self).__init__(**kwargs)
+
+    def _get_file_id(self):
+        try:
+            return next(
+                filter(lambda x: x.name == self.path, self.ref_counts.keys()))
```
bartvm: Surprised this pickles actually, but nonetheless, I'd prefer something like `file_id, = [f.name for f in self.ref_counts.keys()]`

bartvm: And instead of `try` and `except` maybe
```python
file_id = [f.name for f in self.ref_counts.keys()]
if not file_id:
    return self.path
file_id, = file_id  # Raises an error if there are multiple matches, which we probably want?
return file_id
```

bartvm: *`[f for f in self.ref_counts.keys() if f.name == self.path]`

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,124 @@ def get_data(self, state=None, request=None):
...
+        self.data_sources = ([data_source[self.subset] for
+                              data_source in handle.values()]
+                             if self.load_in_memory else None)
+        self._out_of_memory_close(handle)
+
+    def open(self):
+        return None if self.load_in_memory else self._out_of_memory_open()
+
+    def _out_of_memory_open(self):
+        file_id = self._get_file_id()
```
bartvm: I'm a bit confused here, but maybe that's just because I don't know `h5py`'s interface very well. `file_id` can be either a path (string) or a file handle here, right? But then you pass the `file_id` to `h5py.File`, does that make sense?

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,124 @@ def get_data(self, state=None, request=None):
...
+        state = h5py.File(name=file_id, mode="r", driver=self.driver)
+        self.ref_counts[state.id] = self.ref_counts.get(state.id, 0) + 1
```
bartvm: You can make `ref_counts` a `defaultdict(int)` I guess, that way you can just do `self.ref_counts[state.id] += 1`.

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,124 @@ def get_data(self, state=None, request=None):
...
+            self._out_of_memory_close(state)
+
+    def _out_of_memory_close(self, state):
+        self.ref_counts[state.id] -= 1
+        if not self.ref_counts[state.id]:
+            del self.ref_counts[state.id]
+            state.close()
+
+    def get_data(self, state=None, request=None):
+        get_data_method = (self._in_memory_get_data if self.load_in_memory
```
bartvm: This is just a matter of preference I guess, but some people would argue the ternary operator isn't very pythonic. Sometimes I think it's useful, but here I would go for the less dense
```python
if self.load_in_memory:
    return self._in_memory_get_data(*args, **kwargs)
else:
    return self._out_of_memory_get_data(*args, **kwargs)
```

fuel/datasets/hdf5.py
vdumoulin: From the [h5py documentation](http://docs.h5py.org/en/latest/high/file.html#File):

    Name of file (str or unicode), or an instance of h5f.FileID to bind
    to an existing file identifier.

bartvm: RTFM I guess, thanks!

vdumoulin:
@bartvm Your comments have been addressed.

fuel/datasets/hdf5.py
```diff
@@ -64,3 +69,125 @@ def get_data(self, state=None, request=None):
...
+    def get_data(self, state=None, request=None):
+        if self.load_in_memory:
+            return self._in_memory_get_data(state=state, request=request)
+        else:
+            return self._out_of_memory_get_data(state=state, request=request)
+
+    def _in_memory_get_data(self, state=None, request=None):
+        if state is not None or request is None:
+            raise ValueError
+        return self.filter_sources([data_source[request] for data_source
```
bartvm: We can potentially leave this for a later PR, but we should get rid of `filter_sources` at one point. Instead, the `__init__` method should read the `sources` keyword argument and use it to determine which datasets to load from the HDF5 file. `filter_sources` is really a stopgap solution because it assumes all the data/sources are loaded, and then discards the unwanted sources just before returning. This can be very wasteful for large datasets.

vdumoulin: I'll try to hack at it. If it takes me less than 10 minutes it goes into this PR and otherwise I'll make another PR out of it.

vdumoulin: Got it to work.

vdumoulin:
I just rebased on top of master.

bartvm:
Okay, let's merge and take it from there with `fuel-convert` and converting the existing databases. We probably need to write a kind of "HDF5 dataset specification" that shows how to create a dataset that can be interpreted by this class.

vdumoulin:
I can write the documentation. I'll make a PR when it's ready.

fuel/datasets/hdf5.py
```diff
@@ -64,3 +67,137 @@ def get_data(self, state=None, request=None):
...
+        if not file_id:
+            return self.path
+        file_id, = file_id
+        return file_id
+
+    @property
+    def provides_sources(self):
+        if not hasattr(self, '_provides_sources'):
+            handle = self._out_of_memory_open()
+            self._provides_sources = list(handle.keys())
```
bartvm: @vdumoulin Oops, missed this. This should be a `tuple`, not a `list`.

PR #52, opened 2015-03-24
Binarized Sketch

udibr:
I added a data set which is unlabeled binarized images of sketches made by Humans. 
Original data is from http://cybertron.cg.tu-berlin.de/eitz/projects/classifysketch/ released under CC.

This is a replication of binarized_mnist it would make more sense to build a super class for both

### DIFF
--- a/fuel/datasets/Sketch.ipynb
+++ b/fuel/datasets/Sketch.ipynb
@@ -1,0 +1,843 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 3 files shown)


bartvm:
Thanks! I'll have to give the same answer as in https://github.com/bartvm/fuel/pull/48#issuecomment-85021532 though, and wait a bit before merging considering we just merged https://github.com/bartvm/fuel/pull/43, which changes the way we deal with external datasets.

vdumoulin:
@udibr Things have stabilized a little concerning external datasets. If you're still interested in adding this dataset to Fuel, have a look at PR #75, which details how to proceed.

vdumoulin:
Since this PR hasn't seen any activity in nearly three months, I'll close it. Feel free to re-open it if you'd like us to take another look.

PR #54, opened 2015-03-26
HDF5 dataset refactor

vdumoulin:
This PR aims at

* introducing a dataset conversion framework, and
* refactoring one built-in dataset to serve as reference for future PRs.


vdumoulin:
Lots of things to do before it's ready, I'll ping whoever wants to review this when I'm ready.

vdumoulin:
@bartvm I'd like to have your opinion on built-in dataset testing.

The move to HDF5 adds the additional burden of converting the downloaded data in order to run the tests. I see two ways of doing this.

If we convert once and for all, the tests will run faster but the amount of cache space we need will double. We'll also have to go through the awkward process of cache clearing / re-downloading / re-converting every time we change something in the conversion scripts.

If we do the conversion every time the tests are run, we'll be minimizing the cache space needed and the process of changing and testing the conversion scripts will be more straightforward, but the tests will run slower.

What option would you prefer? Do you see a third option I haven't taken into account?

bartvm:
I don't see a third option, but I would go with caching everything. There's no limit I'm aware of, clearing the cache is relatively straightforward (3-4 terminal commands), and it's nice to keep Travis running time to a minimum.

vdumoulin:
Good, I'll implement that.

bin/fuel-convert
```diff
@@ -0,0 +1,24 @@
...
+from fuel import converters
+
+
+if __name__ == "__main__":
+    dataset_options = [
+        m[1] for m in pkgutil.iter_modules(converters.__path__) if not m[2]]
+    parser = argparse.ArgumentParser()
+    parser.add_argument("dataset", help="what dataset to convert",
+                        choices=dataset_options)
+    parser.add_argument("-d", "--directory", help="directory in which input " +
```
bartvm: I think it'd be nice if the directory defaulted to `os.getcwd()`, and the output file to `os.path.join(os.getcwd(), '{}.hdf'.format(dataset))`. So if I downloaded MNIST I can just run `fuel-convert mnist` in the directory I downloaded it to to get `mnist.hdf5`.

bartvm: Ah, just noticed you set the default in `binarized_mnist`, but I would set it here actually.

bin/fuel-convert
```diff
@@ -0,0 +1,24 @@
...
+                        "files reside", type=str, default=None)
+    parser.add_argument("-o", "--output-file", help="where to save the " +
+                        "dataset", type=str, default=None)
+    args = parser.parse_args()
+
+    converter_module = importlib.import_module(
+        'fuel.converters.{}'.format(args.dataset))
```
bartvm: Does this force each converter to be implemented as an independent module? Maybe it would be nicer to assume it is an importable function instead, so that we don't have to put every 4-line convert function in a separate file later on?

fuel/datasets/binarized_mnist.py
```diff
@@ -47,40 +42,27 @@ class BinarizedMNIST(IndexableDataset):
...
-            images = numpy.loadtxt(self.data_path[:-3] + 'amat',
-                                   dtype=config.floatX)
-            if not self.flatten:
-                images = images.reshape((len(images), 28, 28))
-        return [images]
+        return os.path.join(config.data_path, self.folder, self.filename)
+
+    def get_data(self, state=None, request=None):
+        data = super(BinarizedMNIST, self).get_data(state, request)
+        if 'features' in self.sources and self.flatten:
```
bartvm: Maybe not necessary for this PR, but I think we should move the `flatten` argument to the `H5PYDataset` class eventually, because it's something that will come back very often, and should be easy to implement in general (I guess the most general way to implement it is by taking an argument like `flatten_sources=('features',)`.

fuel/datasets/binarized_mnist.py
```diff
@@ -1,19 +1,14 @@
...
-from fuel.datasets import IndexableDataset
-from fuel.utils import do_not_pickle_attributes
+from fuel.datasets import H5PYDataset
 
 logger = logging.getLogger(__name__)
 
 
-@do_not_pickle_attributes('indexables')
-class BinarizedMNIST(IndexableDataset):
+class BinarizedMNIST(H5PYDataset):
```
bartvm: I know that @dwf was quite keen on delegating to instead of inheriting from `H5PYDataset`. It's true that delegating might be beneficial if we later on want to give people the option to switch between `h5py` and PyTables, but I'm not sure how he envisioned that in detail though.

bin/fuel-convert
```diff
@@ -0,0 +1,24 @@
...
+import argparse
+import importlib
+import pkgutil
+
+from fuel import converters
+
+
+if __name__ == "__main__":
+    dataset_options = [
+        m[1] for m in pkgutil.iter_modules(converters.__path__) if not m[2]]
```
bartvm: Another way of doing this would be to set `__all__` in `fuel.converters`, and then read the options from there directly, `dataset_options = converters.__all__`.

bin/fuel-convert
vdumoulin: It does, and I agree with you it's awkward. How about we scrape the `fuel.converters` sub-package for importable functions whose name starts with `convert_`?

fuel/datasets/binarized_mnist.py
vdumoulin: I'm not sure how that would be implemented either, I'm curious to see his opinion on that.

bin/fuel-convert
vdumoulin: I disagree on that; I think putting the HDF5 file where the dataset class expects it is a good default.

vdumoulin:
No problem, I can make the change.

bin/fuel-convert
bartvm: Argh. This DDoS attack on GitHub is driving me crazy.

Anyway, how about we just name the functions after the dataset they convert, e.g. `mnist(directory, save_path)`, and then put those functions in an `__all__` list. That way there's no scraping whatsoever required.

bin/fuel-convert
bartvm: Right, but the only reason we put files in separate directories right now is because datasets could consist of a multitude of files. If we store everything in a single HDF5 file, I think the dataset should expect to read `$FUEL_DATA_PATH/mnist.hdf5`, not `$FUEL_DATA_PATH/mnist/mnist.hdf5`, which is a bit redundant.

The workflow is then simply:
```bash
cd $FUEL_DATA_PATH
curl -O [mnist_files]
fuel-convert mnist
```

vdumoulin:
Would something like

```
  - |
      if [ ! -f binarized_mnist/binarized_mnist.hdf5 ]; then
        ./bin/fuel-convert binarized_mnist \
                           -d binarized_mnist \
                           -o binarized_mnist/binarized_mnist.hdf5
      fi
```

in the `script` section of `.travis.yml` work?

I see you're using a pipe for multi-line commands; what's the reason for that?

bartvm:
The pipe is just YAML notation for a multi-line block of text. If you would split it into multiple lines, Travis would complain saying that `if [ ! -f binarized_mnist/binarized_mnist.hdf5 ]; then` is not a valid command.

We should run `pip install -e . -r requirements.txt` of something in the sorts in `.travis.yml`, no? In that case `fuel-convert` should be in the path I guess, so just `fuel-convert` should do.

bin/fuel-convert
vdumoulin: Agreed. The amount of maintaining is pretty minimal, and it saves me the trouble of implementing the scraping logic.

bin/fuel-convert
vdumoulin: That's a good point. How about we finish this PR and I address this in a PR where I implement `fuel-download`?

fuel/datasets/binarized_mnist.py
vdumoulin: I'll look into that.

bin/fuel-convert
bartvm: Sounds good to me.

fuel/datasets/binarized_mnist.py
vdumoulin: I ended up implementing it as part of this pull request.

vdumoulin:
Tests pass. Here's a summary of what hasn't been addressed yet:

* I'd like to have @dwf 's opinion on the fact that `BinarizedMNIST` directly subclasses `H5PYDataset`.
* Simplifying the defaults of `fuel-convert` and the structure of `$FUEL_DATA_PATH` (will be fixed in a subsequent PR).

fuel/converters/__init__.py
```diff
@@ -0,0 +1,20 @@
...
+                dataset expects to find the data.
+* `save_path` : Where to save the converted data. If not specified,
+                `convert` should provide a sensible default name and
+                put the file in the directory in which the corresponding
+                built-in dataset expects to find the data.
+
+"""
+from fuel.converters.binarized_mnist import binarized_mnist
+
+__all__ = (binarized_mnist,)
```
bartvm: This should be a list of strings i.e. `__all__ = ['binarized_mnist']`.

vdumoulin: Already addressed, Scrutinizer was bugging me about it ;).

fuel/datasets/hdf5.py
```diff
@@ -199,5 +213,10 @@ def _out_of_memory_get_data(self, state=None, request=None):
...
             raise ValueError
-        return self.filter_sources([data_source[request] for data_source in
-                                    state.values()])
+        rval = []
+        for source_name, data_source in state.items():
+            data = data_source[request]
+            if source_name in self.flatten:
+                data = data.reshape((data.shape[0], -1))
+            rval.append(data)
+        return self.filter_sources(rval)
```
bartvm: A `if source_name not in self.sources: continue` should do the trick as well I guess, and means we don't end up loading and reshaping data unnecessarily, right?

vdumoulin: Right. I'll change that.

bin/fuel-convert
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
+import argparse
+import importlib
+import pkgutil
+
+from fuel import converters
+
+
+if __name__ == "__main__":
```
dwf: I think what we will want to do in the long term is add [sub-commands](https://docs.python.org/3/library/argparse.html#sub-commands) for each dataset that lets you control options related to that dataset specifically.

fuel/converters/binarized_mnist.py
```diff
@@ -0,0 +1,64 @@
...
+default_directory = os.path.join(fuel.config.data_path, 'binarized_mnist')
+default_save_path = os.path.join(default_directory, 'binarized_mnist.hdf5')
+
+
+def binarized_mnist(directory=None, save_path=None):
+    """Converts the binarized MNIST dataset to HDF5.
+
+    Converts the binarized MNIST dataset used in R. Salakhutdinov's DBN
+    paper [DBN] to an HDF5 dataset compatible with
+    :class:`fuel.datasets.binarized_mnist.BinarizedMNIST`.
```
dwf: I'd be in favour of using `__init__.py` imports in `datasets/` to flatten this namespace. What do you think @bartvm?

bin/fuel-convert
vdumoulin: Interesting. Would you like this done as part of this PR, or should we open an issue for it?

fuel/datasets/binarized_mnist.py
dwf: I was mostly thinking in terms of handling the multiple file opening conundrum, but if we've got that. What I'm mostly against is implementation details leaking into.

@bartvm is right right in that delegation leaves the door open to multiple backends but I suppose that Python's a dynamic language and we can change the inheritance tree later without many adverse effects, so the play in that possible future is basically to change datasets to inherit from something other than H5PYDataset and then make _that_ class do the delegation.

fuel/converters/binarized_mnist.py
vdumoulin: It [already is](https://github.com/bartvm/fuel/blob/master/fuel/datasets/__init__.py#L5). I guess I should change the reference here, though.

That reminds me of a question I had: Scrutinizer complains about circular imports between `fuel.datasets` and its sub-modules, should we do something about it?

bin/fuel-convert
dwf: Let's leave it for now.

fuel/datasets/binarized_mnist.py
dwf: Woops. My first paragraph went unfinished. What I'm mainly against is h5py-specific implementation details leaking into dataset classes. That isn't happening here, so no big deal as far as I'm concerned.

dwf:
This looks good to me.

PR #55, opened 2015-03-26
Fixing concatenation of list to tuple bug

jfsantos:
self.data_stream.sources is a list, but it was being concatenated to a tuple. I converted the list comprehension to a list so it will not fail.

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -104,8 +104,8 @@ class Mapping(Transformer):
 
     @property
     def sources(self):
-        return self.data_stream.sources + (self.add_sources
-                                           if self.add_sources else ())
+        return self.data_stream.sources + [self.add_sources
+                                           if self.add_sources else ()]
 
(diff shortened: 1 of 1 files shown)


bartvm:
What you changed wasn't a comprehension, it was just a ternary statement (`if-else` shorthand) in parentheses. Right now it could return a `list` with an empty `tuple` inside of it, which isn't right.

If `self.data_stream.sources` was a list, we need to figure out why it is, because it shouldn't be, it should be a `tuple`.

jfsantos:
Oops, sorry about that. The fix may be wrong, but `self.data_stream.sources` was indeed a list in my code. I am using an `H5PYDataset`, and the following as a stream:

```
Mapping(DataStream(timit_test,
                                   iteration_scheme=SequentialScheme(
                                       50000, 500)), remove_inf)
```
(where `remove_inf` is just a function to remove `inf` values from the data).

bartvm:
#57

Issue #47, opened 2015-03-18
MultiProcessing doesn't work with HDF5 files


dmitriy-serdyuk:
Seems, that there is an issue with `MultiProcessing` transformer when HDF5 file is used.

Here is a stack trace:
```
HDF5ExtError: HDF5 error back trace

  File "H5Dio.c", line 182, in H5Dread
    can't read data
  File "H5Dio.c", line 550, in H5D__read
    can't read data
  File "H5Dchunk.c", line 1866, in H5D__chunk_read
    chunked read failed
  File "H5Dscatgath.c", line 542, in H5D__scatgath_read
    datatype conversion failed
  File "H5T.c", line 4809, in H5T_convert
    data type conversion failed
  File "H5Tconv.c", line 3216, in H5T__conv_vlen
    can't read VL data
  File "H5Tvlen.c", line 891, in H5T_vlen_disk_read
    Unable to read VL information
  File "H5HG.c", line 622, in H5HG_read
    unable to protect global heap
  File "H5HG.c", line 262, in H5HG_protect
    unable to protect global heap
  File "H5AC.c", line 1329, in H5AC_protect
    H5C_protect() failed.
  File "H5C.c", line 3570, in H5C_protect
    can't load entry
  File "H5C.c", line 7950, in H5C_load_entry
    unable to load entry
  File "H5HGcache.c", line 141, in H5HG_load
    bad global heap collection signature

End of HDF5 error back trace
```

dwf:
And shit like this is why processes are not a substitute for threads, folks.

bartvm:
http://docs.h5py.org/en/latest/mpi.html Might be worth a read

dwf:
Seems kinda limited/awkward...

bartvm:
@dmitriy-serdyuk Did you try changing `multiprocessing` to `threading`? They should be almost drop-in replacements, and I'm just wondering whether that solves the issue. If so, we can let the user choose whether to use multiprocessing or threading?

dmitriy-serdyuk:
@bartvm , no, seems, it doesn't work.

bartvm:
For reference: https://github.com/h5py/h5py/blob/master/examples/multiprocessing_example.py

bartvm:
[...]

PR #56, opened 2015-03-27
WIP: Data processing server

bartvm:
As requested in #7. We already have some basic multi-processing, but we've already run into issues with HDF5 files (#47) so I figured it was worth revisiting the idea of having a separate server that pushes data around through sockets (optimized for NumPy data).

As I'm kind of figuring out ZMQ and messaging frameworks as I go along, this is a very rough draft, and I might be doing it completely wrong. I'm basically abusing the extended [request-and-reply pattern](http://zguide.zeromq.org/page:all#Shared-Queue-DEALER-and-ROUTER-sockets), making the broker act as a buffer. The idea is that the user creates his data stream in one file and calls `start_server(data_stream)` at that end end. In the script with the main loop the data stream is then given as `ServerDataStream()`.

Long to-do list:

- [ ] If the client dies, it can't reconnect to the server. If we want to allow this, we need something like the [lazy pirate pattern](http://zguide.zeromq.org/page:all#Client-Side-Reliability-Lazy-Pirate-Pattern) I think.
- [x] Support for multiple sources.
- [x] Server needs to send stop iteration message to client
- [x] Documentation
- [x] Client interface (`ServerDataStream`)
- [x] Benchmarking
- [x] Python 3 compatibility (byte strings, `buffer` interface)
- [ ] Checkpointing support
- [ ] Iterator protocol support (e.g. go to next iteration)
- [ ] Automate sources discovery

Any feedback welcome. Like I said, making this up as I go along.


bartvm:
Crude way of sending `StopIteration` messages now, and also wrote a `ServerDataStream` class (which largely ignores the entire data stream protocol). But this should work now:

```bash
python fuel/server.py
```
```python
form fuel.streams import ServerDataStream
data_stream = ServerDataStream(('features',))
it = data_stream.get_epoch_iterator()
next(it)
```

bartvm:
Woohoo! For my dogs vs. cats model, testing this on a single run, the time to read data fell from 76.84 seconds per epoch to 5.61 seconds. That means it went from being 38% of the training time to 4% of the training time :)

bartvm:
Whoever wants to have a look, this should be good for initial review (@dwf, @vdumoulin, @rizar)

vdumoulin:
I'm by no means acquainted with networking, but if you provide a minimal starting example I can "kick the tires" and tell you what I think.

bartvm:
```python
# server.py
from fuel.datasets import MNIST
mnist = MNIST('train')
data_stream = DataStream(
    mnist, iteration_scheme=SequentialScheme(1500, 500)
)
start_server(data_stream)
```
```python
# train.py
from fuel.streams import ServerDataStream
data_stream = ServerDataStream(('features', 'targets'))
epoch = data_stream.get_epoch_iterator()
batch = next(epoch)
print(batch)
```
Running `python server.py` followed by `python train.py` should work now.

vdumoulin:
Thanks, I'll have a look at it.

vdumoulin:
I still haven't read your code into detail, but if I call `python train.py` twice after calling `python server.py` (which needs an import for `start_server` to work), it gets stuck on the second call.

bartvm:
Oops, forgot the `start_server` import.

I'm aware of the inability for the client to reconnect, it's the "lazy pirate pattern" problem I referred to the in the description. Basically, ZMQ binds the processes together in the background somehow. If the client dies and tries to reconnect, its ID changes and it can't connect to the server anymore. The only way to fix this is by manually implementing a pattern where the client forcefully closes the socket, deregisters from the poll, and reconnects (see e.g. [this code](http://zguide.zeromq.org/py:lpclient)).

I was planning on fixing that in a later PR, if at all. It normally doesn't make sense to reconnect the client without restarting the server. The client would start receiving data from the middle of an epoch, which doesn't make much sense.

vdumoulin:
Makes sense.

fuel/server.py
```diff
@@ -0,0 +1,177 @@
...
+import zmq
+
+from fuel.utils import buffer_
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level='INFO')
+
+
+def send_arrays(socket, arrays, flags=0, copy=True, track=False, stop=False):
+    """Send a NumPy array using the buffer interface and some metadata.
```
dwf: singular, yet below you say we're sending a list of numpy arrays

fuel/server.py
```diff
@@ -0,0 +1,177 @@
...
+    arrays : list
+        A list of :class:`numpy.ndarray` to transfer.
+    stop : bool, optional
+        Instead of sending a series of NumPy arrays, send a JSON object
+        with a single `stop` key. The :func:`recv_arrays` will raise
+        ``StopIteration`` when it receives this.
+
+    See Also
+    --------
+    :meth:`zmq.Socket.send` for other arguments
```
dwf: Does it make sense to expose these arguments? Are they used at all in client code you've written thus far?

bartvm: I'm not sure :) I need to read into ZMQ a bit more to know what the effect of these flags would be, but you're probably right that I can remove them from now.

fuel/server.py
bartvm: Right, should be plural.

fuel/server.py
```diff
@@ -0,0 +1,177 @@
...
+    shape and data type for each array is transferred first. Subsequently
+    the arrays are sent as bytestreams (through NumPy's support of the
+    buffering protocol).
+
+    """
+    if stop:
+        mds = {'stop': True}
+        return socket.send_json(mds, flags)
+    else:
+        mds = [{'dtype': str(array.dtype), 'shape': array.shape}
```
dwf: I'd add two things:

* A C-ordered boolean flag, along with a check making sure this is the case that raises `NotImplementedError` for now (we don't necessarily need to support it, ever, but if someone is using some exotic library it could obviate some truly bizarre bugs). It's the one piece of metadata that the [NPY format](https://github.com/numpy/numpy/blob/master/doc/neps/npy-format.rst) stores that we're currently missing here, I think.
* Perhaps a format specifier that says that this is a NumPy array in the first place. This would future-proof it against a possible scenario where we want to send e.g. various types of `scipy.sparse` objects instead.

fuel/server.py
```diff
@@ -0,0 +1,177 @@
...
+               for array in arrays]
+        socket.send_json(mds, flags | zmq.SNDMORE)
+        for array in arrays[:-1]:
+            socket.send(array, flags | zmq.SNDMORE, copy=copy, track=track)
+        return socket.send(arrays[-1], flags, copy=copy, track=track)
+
+
+def recv_arrays(socket, flags=0, copy=True, track=False):
+    """Receive a NumPy array.
```
dwf: Same plurality issue here.

fuel/server.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def broker(server_port, client_port):
+    """The broker is run in a separate process and holds the queue."""
+    logger.debug('binding broker to sockets')
+    context = zmq.Context()
+    frontend = context.socket(zmq.ROUTER)
+    backend = context.socket(zmq.DEALER)
+    frontend.bind("tcp://*:{}".format(client_port))
+    backend.bind("tcp://*:{}".format(server_port))
```
dwf: Is tcp what we want to be using always? what about unix sockets? (I have no idea)

Also, what exactly does the * entail? (Will look into this after dinner if you haven't gotten back)

fuel/streams.py
```diff
@@ -135,3 +138,42 @@ def get_epoch_iterator(self, **kwargs):
         else:
             self._fresh_state = False
         return super(DataStream, self).get_epoch_iterator(**kwargs)
+
+
+class ServerDataStream(AbstractDataStream):
```
dwf: We should file a ticket about querying `num_examples` and `num_batches` from data streams in general, and over the network in particular...

dwf: (I think ZMQ has some sort of side-channel things that IPython uses for this kind of riff-raff metadata.)

fuel/server.py
bartvm: No idea... This code comes straight [from the manual](http://zguide.zeromq.org/py:rrbroker). I looked into it briefly, but I'm still not entirely clear as to how the socket connections work and what `*` means. I guess it means it's just listening on those ports, accepting connections from any host?

Not sure about using `udp` or `ipc` sockets either. I did find this [question on Stack Overflow](http://stackoverflow.com/questions/24027579/which-protocol-should-i-use-for-pyzmq) where one person says:

> I don't know if pyzmq has its own delivery guarantees on top of the transport protocol. If it doesn't, TCP will guarantee in-order delivery of all messages, while UDP may drop messages if the network is congested.

> If you don't know what you want, TCP is the simplest and safest choice.

A bit more googling didn't give any definitive answers about the whole `tcp` vs. `udp` vs `ipc` question.. I could benchmark them and try, but at the very least I'd say that `tcp` is the safe option (and it technically makes it possible to run the processing server on a separate computer).

fuel/streams.py
bartvm: Yeah, there's a bunch of things that strictly speaking need to be communicated: Getting a new iterator, `num_examples` and `num_batches`, resetting the state, etc.

If you can find a particularly clean way to do it, let me know, although I think it won't be too bad to do it the naive way. Right now the server responds to every message with a batch, but it could easily check the message and send different data instead.

fuel/server.py
bartvm: Thanks for the link to the NPY format. I ended up just using the `numpy.lib.format.header_data_from_array_1_0` function, which is the function the NPY format uses to extract the necessary metadata from an array. Added support for Fortran arrays in the process.

I'll leave the format specifier for a later PR if you think that's okay. I think it's clear enough from the function right now that it just supports NumPy arrays. Anything more exotic will probably just crash, like it should.

fuel/server.py
bartvm: `udp` gives an error saying it's not supported (for this kind of socket I guess), `ipc` just hangs and doesn't do anything... `tcp` for now it is I would say!

bartvm:
Nothing I can do about the drop in coverage by the way. Coverage doesn't count lines run in subprocesses. It actually has an experimental flag for this in the latest alpha releases, but I can't get that to play nice with Coveralls (whose support for Coverage 4 is alpha level, so alpha level support for an alpha level release...). Give it a month or two and I guess we should be able to enable it.

bartvm:
As I said, figuring this out as I go along: After reading the documentation a bit more I figured I could switch to a push-pull pattern and use ZeroMQ's high-water mark to limit the queue instead, which turns out to be significantly faster. Reading time for my Dogs vs. Cats model (originally 76.84 s) went from 5.61 s to now 1.9 s, 3 times faster and only 1.4% of the training time now.

PR #57, opened 2015-03-27
Sources H5PYDataset should be tuple

bartvm:


vdumoulin:
Oh, my bad! Feel free to merge when the tests pass.

PR #59, opened 2015-03-28
Add download script

vdumoulin:
This PR introduces a `fuel-download` script to download built-in datasets.

Its implementation is similar to `fuel-convert`'s implementation: the script allows to choose amongst the download functions listed in `fuel.downloaders.__all__`.

I changed the default location to the current working directory following a discussion in #54.

The workflow for downloading a built-in dataset now looks like this:

```
cd $FUEL_DATA_PATH
fuel-download mnist
fuel-convert mnist
```

### DIFF
--- a/fuel/downloaders/__init__.py
+++ b/fuel/downloaders/__init__.py
@@ -1,7 +1,9 @@
 """Download modules for built-in datasets.
 
-Download methods accept a single argument, `save_directory`, which tells
-where to save the downloaded files.
+Download methods accept two arguments:
+
+* `save_directory` : Where to save the downloaded files
+* `clear` : If `True`, clear the downloaded files. Defaults to `False`.
 
 """
 from fuel.downloaders.binarized_mnist import binarized_mnist
--- a/fuel/downloaders/__init__.py
+++ b/fuel/downloaders/__init__.py
@@ -1,0 +1,9 @@
+"""Download modules for built-in datasets.
+
+Download methods accept a single argument, `save_path`, which tells
+where to save the downloaded files.
+
+"""
+from fuel.downloaders.binarized_mnist import binarized_mnist
+
+__all__ = ('binarized_mnist',)
--- a/fuel/downloaders/__init__.py
+++ b/fuel/downloaders/__init__.py
@@ -1,6 +1,6 @@
 """Download modules for built-in datasets.
 
-Download methods accept a single argument, `save_path`, which tells
+Download methods accept a single argument, `save_directory`, which tells
 where to save the downloaded files.
 
 """
--- a/fuel/downloaders/base.py
+++ b/fuel/downloaders/base.py
@@ -1,0 +1,35 @@
+import os
+import shutil
+
+import urllib3
+from urllib3.util.url import parse_url
+
+
+def download(url, path=None):
+    """Downloads a given URL to a specific directory or file.
+
+    Parameters
+    ----------
+    url : str
+        URL to download.
+    path : str, optional
+        Where to save the downloaded URL. Defaults to `None`, in which
+        case the current working directory is used.
+
+    """
+    http = urllib3.PoolManager()
+    with http.request('GET', url, preload_content=False) as response:
+        if not path:
+            path = os.getcwd()
+        if os.path.isdir(path):
+            headers = response.getheaders()
+            if 'Content-Disposition' in headers:
+                filename = headers[
+                    'Content-Disposition'].split('filename=')[1].trim('"')
+            else:
+                filename = os.path.basename(parse_url(url).path)
+            if not filename:
+                raise ValueError("no filename given")
+            path = os.path.join(path, filename)
+        with open(path, 'wb') as out_file:
+            shutil.copyfileobj(response, out_file)
--- a/fuel/downloaders/base.py
+++ b/fuel/downloaders/base.py
@@ -33,3 +33,27 @@ def download(url, path=None):
             path = os.path.join(path, filename)
         with open(path, 'wb') as out_file:
             shutil.copyfileobj(response, out_file)
+
+
+def default_manager(urls, filenames):
+    """Returns a function to download or clear files from a list of URLs.
+
+    Parameters
+    ----------
+    urls : list of str
+        URLs to download from.
+    filenames : list of str
+        Files to save to.
+
+    """
+    def manager_fn(args):
+        save_directory = args.directory
+        files = [os.path.join(save_directory, f) for f in filenames]
+        if args.clear:
+            for f in files:
+                if os.path.isfile(f):
+                    os.remove(f)
+        else:
+            for url, f in zip(urls, files):
+                download(url, f)
+    return manager_fn
--- a/fuel/downloaders/base.py
+++ b/fuel/downloaders/base.py
@@ -1,6 +1,7 @@
 import os
 import shutil
 
+import certifi
 import urllib3
 from urllib3.util.url import parse_url
 
@@ -17,7 +18,8 @@ def download(url, path=None):
         case the current working directory is used.
 
     """
-    http = urllib3.PoolManager()
+    http = urllib3.PoolManager(
+        cert_reqs='CERT_REQUIRED', ca_certs=certifi.where())
     with http.request('GET', url, preload_content=False) as response:
         if not path:
             path = os.getcwd()
--- a/fuel/downloaders/base.py
+++ b/fuel/downloaders/base.py
@@ -6,35 +6,42 @@ import urllib3
 from urllib3.util.url import parse_url
 
 
-def download(url, path=None):
-    """Downloads a given URL to a specific directory or file.
+def filename_from_url(url, path=None):
+    """Parses a URL to determine a file name.
 
     Parameters
     ----------
     url : str
-        URL to download.
-    path : str, optional
-        Where to save the downloaded URL. Defaults to `None`, in which
-        case the current working directory is used.
+        URL to parse.
 
     """
     http = urllib3.PoolManager(
         cert_reqs='CERT_REQUIRED', ca_certs=certifi.where())
     with http.request('GET', url, preload_content=False) as response:
-        if not path:
-            path = os.getcwd()
-        if os.path.isdir(path):
-            headers = response.getheaders()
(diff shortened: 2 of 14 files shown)


vdumoulin:
@bartvm The last thing needed for this PR to be ready for review is to change where `BinarizedMNIST` expects the HDF5 file from `$FUEL_DATA_PATH/binarized_mnist/` to `$FUEL_DATA_PATH`, following our discussion in #54.

In order to do that, the Travis cache needs to be cleared.

bartvm:
Sure, do you want me to clear it now? Or did you not make the changes yet?

fuel/downloaders/binarized_mnist.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Parameters
+    ----------
+    save_directory : str
+        Where to save the downloaded files.
+
+    """
+    base_url = ('http://www.cs.toronto.edu/~larocheh/public/datasets/' +
+                'binarized_mnist/binarized_mnist_')
+    call(['curl', base_url + 'train.amat', '-o',
```
bartvm: I'd prefer using `requests` for this since `curl` might not be installed on some systems (they might have `wget`, or something different entirely), plus it's just nicer if we don't have to use bash commands.

fuel/downloaders/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""Download modules for built-in datasets.
+
+Download methods accept a single argument, `save_directory`, which tells
+where to save the downloaded files.
+
+"""
+from fuel.downloaders.binarized_mnist import binarized_mnist
+
+__all__ = ('binarized_mnist',)
```
bartvm: Could you turn this into a list? Same for the `converters/__init__.py` actually. The Python specification says it should be a list, and I've never seen a tuple used actually, so it might lead to some strange behavior down the road (maybe?)

vdumoulin: Scrutinizer complained when I was using a list (it said the list was mutable so it couldn't guarantee that everything was tested properly).

vdumoulin:
I just pushed a commit that makes the changes, could you validate that the changes to `.travis.yml` are correct?

Now would be a good time to clear the cache, in order to remove the unneeded `$TRAVIS_BUILD_DIR/binarized_mnist`.

fuel/downloaders/__init__.py
bartvm: Strange, I traced it back to https://github.com/GreenSteam/pep257/issues/62. Sounds a bit paranoid to me, but sure, let's keep the tuples then.

fuel/downloaders/binarized_mnist.py
bartvm: We'll need something like this btw: http://stackoverflow.com/a/16696317

bartvm:
Cache cleared!

vdumoulin:
I just rebased and force-pushed.

bartvm:
Btw, not sure, but I think you should have rights to clear the cache as well. If you've got the Travis CLI installed it should be a matter of `travis login` followed by `travis cache --repo bartvm/fuel --delete`.

vdumoulin:
Good to know, thanks!

vdumoulin:
The only thing left to verify is whether Travis will cache the generated `binarized_mnist.hdf5`.

fuel/downloaders/binarized_mnist.py
bartvm: Or actually, maybe this is a better option, because most people already have it installed: http://stackoverflow.com/a/27389016

bartvm:
It will say at the very bottom of the Travis log which files it is caching.

fuel/downloaders/binarized_mnist.py
vdumoulin: Are you sure most people have urllib3 installed?

vdumoulin: I managed to get it working with `six.moves.urllib`, although for some reason the object returned by `urllib.request.urlopen` can't be used as a context manager, which makes my implementation a bit less clean than I would like.

dwf:
Just in reference to `urllib3`: it's a very stable package and it's pure Python, so it's basically Python packaging on Easy Mode, and I'd be fine depending on that if it helps us at all. (Odd trivia, it is written and maintained by my undergrad homeboy @shazow -- I've never worked on a package that had any reason to use it... *OH GOD WE'RE CROSSING THE STREAMS*).

dwf:
I'm going to raise something because I want to very soon add ImageNet support to this tool: some datasets that we would like built-in do not have a public download URL. In ImageNet's case, there is a download URL, but not a public one. What would be a good way to handle this? I'm thinking argparse subcommands are the most flexible option.

shazow:
If you don't care about anything, then yoloswag stdlib urllib is fine. If you care about verified HTTPS/TLS certs, then you probably want [urllib3+certifi](https://urllib3.readthedocs.org/en/latest/#usage), or if you care about re-using sockets/connection pooling, thread safety, etc. Alternatively you could use requests which bundles urllib3+certifi and a bunch of other things.

vdumoulin:
@dwf We could turn functions defined in `fuel.downloaders` into functions that accept a subparser and fill in its arguments.

vdumoulin:
I've refactored things a bit. We're now using `urllib3` (thanks @shazow!) to download, and downloading has been factored into the `fuel.downloaders.base.download` function like @bartvm suggested.

@dwf I managed to get subparsers working for fuel-download. Functions defined in `fuel.downloaders` now take an `argparse.ArgumentParser` object as input and configure it.

The logic in `fuel.downloaders.binarized_mnist` could be further extracted into a generic function when downloading the data amounts to downloading a list of files.

vdumoulin:
I'll try to increase test coverage (if relevant) to make coveralls pass.

vdumoulin:
@dwf You mentioned testing the conversion script. Do you have a suggestion on how this could be done effectively?

vdumoulin:
@dwf Ping RE.: conversion script testing. Do you have a suggestion on how this could be done effectively?

dwf:
I have been thinking about this especially in light of my in progress
script for imagenet. I think one way is to create in-memory mock files with
io.BytesIO that is much smaller than the original and verify things that
way. What do you think? (Going to have another look and try and offer
testability refactoring comments specifically)

vdumoulin:
Sounds good. One of the things that could be done is factor the logic of concatenating data sources and creating attributes into a function that operates on file handles and test that function instead.

vdumoulin:
I made a small step in the right direction by factoring out of `converters.binarized_mnist` the logic of filling the HDF5 file.

I could go one step further and make conversion methods accept either a path or a file handle for input/output files so we can give our mock files as input.

The only thing that's not clear to me yet is what we'll do about shapes: conversion scripts will likely make assumptions about the size of the data stored in the input file which will break if we make small mock files. Maybe there's a way to store sparse arrays filled with zeros that won't blow up memory usage but will retain the right shapes?

vdumoulin:
Maybe [this](http://docs.h5py.org/en/latest/high/dataset.html#filter-pipeline) would be the way to go.

vdumoulin:
@dwf @bartvm I feel like everything that needed to be addressed for this PR has been addressed; in fact, this PR even started spilling onto other issues.

I propose that we leave unit testing conversion modules for another PR and merge this one if everything's okay on your end.

That will allow me to rebase PR #64 and go forward with it.

fuel/downloaders/base.py
```diff
@@ -0,0 +1,84 @@
...
+        filenames[i] = filename
+    files = [os.path.join(save_directory, f) for f in filenames]
+
+    if args.clear:
+        for f in files:
+            if os.path.isfile(f):
+                os.remove(f)
+    else:
+        for url, f in zip(urls, files):
+            with open(f, 'w') as file_handle:
```
bartvm: Should be `wb` I think, else Python 3 complains.

tests/test_downloaders.py
```diff
@@ -0,0 +1,66 @@
...
+        for key, val in kwargs.items():
+            setattr(self, key, val)
+
+
+def test_filename_from_url():
+    assert filename_from_url(iris_url) == 'iris.data'
+
+
+def test_download():
+    f = tempfile.SpooledTemporaryFile()
```
bartvm: Does opening this with `mode='wb'` and getting ride of the `encode` part solve the Travis error?

bartvm:
I'm happy to merge this as soon as Python 3 on Travis is pacified. There are some tiny details I'd change, but I'd rather merge and beautify afterwards than stalling PRs for too long.

dwf:
Agreed, let's merge once Travis is happy.

vdumoulin:
What ended up fixing the test failures was changing the mode from `'w'` to `'wb'` in `fuel.downloaders.base.default_downloader` and removing the `encode` part in `test_download`.

I'm not 100% confident in the fix, could any of you confirm that I'm not doing something stupid that just happens to make these tests pass?

dwf:
wb is good practice anyway for cross platform compatibility buy I'll have a
look

tests/test_downloaders.py
```diff
@@ -0,0 +1,66 @@
...
+    download(iris_url, f)
+    f.seek(0)
+    assert hashlib.sha256(f.read()).hexdigest() == iris_hash
```
bartvm: Looks good to me, but maybe switch to MD5? On my computer that's almost 3 times faster for a 1GB file (1.4 s vs 3.6 s), which might pay off for very large datasets.

dwf: I'm -0.5 on MD5, given that it has devastating security problems, but I
suppose it's not a big risk since we're not using it to verify executables.

bartvm: Yeah, it's just for file download integrity, right? Spreading a virus
through fuel-download is probably not the most efficient attack vector
anyway :p

vdumoulin: Sounds good, I changed it. When we do implement checksum verification of downloaded files, I guess that's something that can be left as a user choice, provided that we compute the checksums using whatever options we want to offer.

vdumoulin:
Looks like all tests are passing (except Coveralls, which fails for a meager 0.01% decrease in test coverage).

PR #64, opened 2015-04-01
[WIP] Change H5PYDataset dataset split interface

vdumoulin:
Not all tests are guaranteed to pass at the moment.

### DIFF
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -117,22 +117,22 @@ attribute of the root group.
 >>> split_array = numpy.empty(
 ...     6,
 ...     dtype=numpy.dtype([
-...         ('split', numpy.str_, 5),
-...         ('source', numpy.str_, 15),
+...         ('split', 'a', 5),
+...         ('source', 'a', 15),
 ...         ('start', numpy.int64, 1), ('stop', numpy.int64, 1),
 ...         ('available', numpy.bool, 1),
-...         ('comment', numpy.str_, 1)]))
->>> split_array[0:3]['split'] = 'train'
->>> split_array[3:6]['split'] = 'test'
->>> split_array[0:6:3]['source'] = 'vector_features'
->>> split_array[1:6:3]['source'] = 'image_features'
->>> split_array[2:6:3]['source'] = 'targets'
+...         ('comment', 'a', 1)]))
+>>> split_array[0:3]['split'] = 'train'.encode('utf8')
+>>> split_array[3:6]['split'] = 'test'.encode('utf8')
+>>> split_array[0:6:3]['source'] = 'vector_features'.encode('utf8')
+>>> split_array[1:6:3]['source'] = 'image_features'.encode('utf8')
+>>> split_array[2:6:3]['source'] = 'targets'.encode('utf8')
 >>> split_array[0:3]['start'] = 0
 >>> split_array[0:3]['stop'] = 90
 >>> split_array[3:6]['start'] = 90
 >>> split_array[3:6]['stop'] = 100
 >>> split_array[:]['available'] = True
->>> split_array[:]['comment'] = '.'
+>>> split_array[:]['comment'] = '.'.encode('utf8')
 >>> f.attrs['split'] = split_array
 
 We created a 1D numpy array with six elements. The ``dtype`` for this array
@@ -143,6 +143,11 @@ maximum length we needed to store: that's 5 for the ``split`` element
 (``'vector_features'`` being the longest source name). We didn't include any
 comment, so the length for that element was set to 1. Due to a quirk in pickling
 empty strings, we put ``'.'`` as the comment value.
+
+.. warning::
+
+    Due to limitations in h5py, you must make sure to use bytes for ``split``,
+    ``source`` and ``comment``.
 
 :class:`~.datasets.hdf5.H5PYDataset` expects the ``split`` attribute of the
 root node to contain as many elements as the cartesian product of all sources
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -55,11 +55,17 @@ things if your data happens to meet these assumptions:
 * All data is stored into a single HDF5 file.
 * Data sources reside in the root group, and their names define the source
   names.
-* Data sources are not explicitly split. Instead, splits are defined as
-  attributes of the root group. They're expected to be numpy arrays of
-  shape ``(2,)``, with the first element being the starting point
-  (inclusive) of the split and the last element being the stopping
-  point (exclusive) of the split.
+* Data sources are not explicitly split. Instead, splits are defined in the
+  ``split`` attribute of the root group. It's expected to be a 1D numpy
+  array of compound ``dtype`` with six fields, organized as follows:
+
+  1. ``split`` : string identifier for the split name
+  2. ``source`` : string identifier for the source name
+  3. ``start`` : start index (inclusive) of the split in the source array
+  4. ``stop`` : stop index (exclusive) of the split in the source array
+  5. ``available`` : boolean, ``False`` is this split is not available for this
+     source
+  6. ``comment`` : comment string
 
 .. tip::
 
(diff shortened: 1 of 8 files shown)


fuel/datasets/hdf5.py
```diff
@@ -113,25 +115,86 @@ class H5PYDataset(Dataset):
     """
     ref_counts = defaultdict(int)
 
-    def __init__(self, path, which_set=None, subset=None, load_in_memory=False,
+    def __init__(self, path, which_set, subset=None, load_in_memory=False,
```
dwf: You should remove the "optional" tag from the `which_set` docstring.

fuel/datasets/hdf5.py
```diff
@@ -113,25 +115,86 @@ class H5PYDataset(Dataset):
...
                  flatten=None, driver=None, **kwargs):
         self.path = path
+        self.driver = driver
+        if which_set not in self.available_splits:
+            raise ValueError(
+                "'{}' split is not provided by this dataset".format(which_set))
         self.which_set = which_set
-        self.subset = subset if subset else slice(None, None, None)
+        subset = subset if subset else slice(None, None, None)
```
dwf: I think `slice(None)` works.

fuel/datasets/hdf5.py
```diff
@@ -113,25 +115,86 @@ class H5PYDataset(Dataset):
...
+        if subset.step not in (1, None):
+            raise ValueError("subset.step must be either 1 or None")
+        self.subsets = [subset for source in self.provides_sources]
```
dwf: Do we really need both `self.subset` and `self.subsets`?

fuel/datasets/hdf5.py
```diff
@@ -113,25 +115,86 @@ class H5PYDataset(Dataset):
...
+                    "not provided by the '{}' split".format(self.which_set))
+
+        super(H5PYDataset, self).__init__(**kwargs)
 
         self.load()
 
+    @staticmethod
+    def create_split_array(split_dict):
+        # Determine maximum split, source and string lengths
+        splits = tuple(split_dict.keys())
```
dwf: Seems like you could use `split_dict` everywhere you use this (since iterating on a dict iterates on keys).

fuel/datasets/hdf5.py
vdumoulin: We don't have `self.subset` at the moment.

dwf: Oops.

vdumoulin:
The tests won't pass at the moment. I'll have to clear the Travis cache in order to re-convert the binarized MNIST dataset, and then the tests should pass.

There are a couple more unit tests I could write for `fuel.converters.base.fill_hdf5_file`. I'll do that tomorrow.

vdumoulin:
@dwf @bartvm I think this PR is ready for a final review.

In order for the tests to pass, I'll have to clear the Travis cache and re-build it, which will make `BinarizedMNIST`-related tests fail in other opened PRs.

For that reason, I'll wait until we all agree that the only thing remaining for this PR to be merged is a confirmation that Travis tests pass, and then I'll proceed.

vdumoulin:
Oh, one other thing I'll need to do: change the tutorial.

vdumoulin:
I've changed the tutorial to reflect changes in the interface.

vdumoulin:
@bartvm By the way, it seems that Scrutinizer has stalled on another build, can it be restarted?

bartvm:
I cancelled the stalling build, hopefully Scrutinizer will catch up in a
bit.

docs/h5py_dataset.rst
```diff
@@ -55,11 +55,17 @@ things if your data happens to meet these assumptions:
 * All data is stored into a single HDF5 file.
 * Data sources reside in the root group, and their names define the source
   names.
-* Data sources are not explicitly split. Instead, splits are defined as
-  attributes of the root group. They're expected to be numpy arrays of
-  shape ``(2,)``, with the first element being the starting point
-  (inclusive) of the split and the last element being the stopping
-  point (exclusive) of the split.
+* Data sources are not explicitly split. Instead, splits are defined in the
```
dwf: I would say, "not explicitly split into separate HDF5 datasets or separate HDF5 files."

docs/h5py_dataset.rst
```diff
@@ -55,11 +55,17 @@ things if your data happens to meet these assumptions:
...
+  ``split`` attribute of the root group. It's expected to be a 1D numpy
+  array of compound ``dtype`` with six fields, organized as follows:
+
+  1. ``split`` : string identifier for the split name
+  2. ``source`` : string identifier for the source name
+  3. ``start`` : start index (inclusive) of the split in the source array
+  4. ``stop`` : stop index (exclusive) of the split in the source array
+  5. ``available`` : boolean, ``False`` is this split is not available for this
```
dwf: if* this split is not available

docs/h5py_dataset.rst
```diff
@@ -105,11 +111,73 @@ with training examples and the last 10 rows with test examples.
...
+>>> split_array[0:3]['start'] = 0
+>>> split_array[0:3]['stop'] = 90
+>>> split_array[3:6]['start'] = 90
+>>> split_array[3:6]['stop'] = 100
+>>> split_array[:]['available'] = True
+>>> split_array[:]['comment'] = '.'.encode('utf8')
+>>> f.attrs['split'] = split_array
+
+We created a 1D numpy array with six elements. The ``dtype`` for this array
+is a compound type: every element of the array is a tuple of ``(str, str, int,
```
dwf: Maybe mention that the strs are bytes on Python 3? Or do you handle this somewhere else?

dwf: Nevermind I see below.

docs/h5py_dataset.rst
```diff
@@ -105,11 +111,73 @@ with training examples and the last 10 rows with test examples.
...
+a dictionary mapping split names to dictionaries. Those dictionaries map source
+names to tuples of length 2 or 3: the first two elements correspond to the start
+and stop indexes and the last element is an optional comment for the
+split/source pair. The method will create the array behind the scenes,
+choose the string lengths automatically and populate it with the information
+in the split dictionary. If a particular split/source combination isn't present,
+its ``available`` attribute is set to ``False``, which allows us to specify
+only what's actually present in the HDF5 file we created.
 
 We flush, close the file and *voilà*!
```
dwf: @bartvm Is Sphinx cool with a unicode literal in source file?

fuel/datasets/hdf5.py
dwf: ^-- Reminder about this comment, doesn't seem like  it was addressed yet...

docs/h5py_dataset.rst
bartvm: I guess it should be, it's fine with unicode characters in docstrings as long as you use `u""""`.

fuel/datasets/hdf5.py
```diff
@@ -113,25 +115,105 @@ class H5PYDataset(Dataset):
...
         self.load()
 
+    @staticmethod
+    def create_split_array(split_dict):
+        """Create a valid array for the `split` attribute of the root node.
+
+        Parameters
+        ----------
+        split_dict : dict
+            Maps split names to dict. Those dict map source names to
```
dwf: It just occurred to me, did you mention anywhere that there is no assumption that splits are mutually exclusive? (There is no assumption about that, right?)

dwf:
LGTM.

Since we're increasingly defining a custom format built on top of HDF5, it occurs to me that we should start versioning the files that we write. Should you want to add a `fuel_format_version` somewhere and choose this to be format version 1.0? cc @bartvm

dwf:
I'm happy to merge after my comments are fixed.

docs/h5py_dataset.rst
vdumoulin: It builds fine locally on my computer.

fuel/datasets/hdf5.py
vdumoulin: I don't mention it anywhere, but you're right, there is no such assumption.

There are some assumptions being made in the utility functions, but that's mostly a result of me not wanting to cover every imaginable case at the same time.

Would you like me to put that somewhere?

vdumoulin:
I don't know if now is the right time for file versioning.

I think we should implement versioning in Fuel itself first, and include the Fuel version in a `fuel_version` attribute of the root node of the HDF5 file.

dwf:
Au contraire, I think versioning the file format is more important than
versioning fuel itself.

Issue #65, opened 2015-04-04
H5PYDataset does not play well with DataStream for serialization


jfsantos:
This discussion started at the [mailing list](https://groups.google.com/forum/#!topic/blocks-users/E01_gG7hPD4).

`ref_sources` is a class attribute in `H5PYDataset`, which causes it to not be serialized. However, if you use a `DataStream` in your code, a handle to the HDF5 file is saved as an instance attribute, which causes it to be serialized. However, most objects in h5py can't be serialized (more info [here](https://github.com/h5py/h5py/issues/531)), which causes loading a pickled `DataStream` to fail. I am not sure on what could be done, as saving the DataStream state is useful when the dataset is not an `H5PYDataset`. The handle to the HDF5 file does indeed hold a state while the file is loaded (which driver is being used, buffering settings, etc), but this information is not useful if reloading an HDF5 file from scratch.

dwf:
Yeah, it is a careful dance we have to do, mutating a self.classvar with
methods or += works but assigning to it makes it an instance attribute,
which is annoying. Maybe we just need to adopt a convention of always using
self.__class__ so these errors don't creep in again.

dwf:
Oh oops, I misread.

Maybe we need a proxy for the state that is not the HDF5 file handle itself.

jfsantos:
What about keeping the handle together with the counter in a tuple inside the `ref_counts` dict and using `handle.id` as the state? I think that would actually fix the class to enable sharing file handles, as the way the class is written today, if one creates two streams accessing the same dataset, they will each have a different file handle: they call `open` which calls `_out_of_memory_open`, which is currently always creating a new handle by calling `h5py.File`.

Would you mind commenting on this, @vdumoulin?

bartvm:
`h5py.File` doesn't always create a new handle. If you pass it an existing handle (as is done in the current implementation) it will simply return this, so currently classes already share a handle.

Passing around the file id as state instead of the handle itself sounds like a good solution though.

jfsantos:
Right, somehow I overlooked this detail. However, `h5py.h5f.FileID` is also not picklable. This is the reason I decided to use the file path as state instead of the handle. I changed a few things to make this work but it's kind of messy, so we could just keep everything else as it was (ref_counts storing handles as keys and counts as values) and just add a method `_get_handle` to get the right handle from the keys list based on the file path (if it exists). A problem is that one would have to look for it every time, but this is what was being done in `_get_file_id` anyway. What do you think?

[...]

PR #67, opened 2015-04-06
Changed H5PYDataset to enable sharing file handles and fix serialization

jfsantos:
This is a tentative implementation to fix the issues reported in #65. The "state" for an out-of-memory H5PYDataset is now just the HDF5 file name, and the class will share a handle to that file as long as there is someone accessing it (ref_counts for that handle >  1).

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -133,7 +133,7 @@ class H5PYDataset(Dataset):
         self.load()
 
     def _get_file_id(self):
-        file_id = [f for f in self.ref_counts.keys() if f.name == self.path]
+        file_id = [f for f in H5PYDataset.ref_counts.keys() if f == self.path]
         if not file_id:
             return self.path
         file_id, = file_id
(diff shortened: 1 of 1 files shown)


vdumoulin:
@jfsantos Would you mind rebasing your code on top of `master`? I'll have a look at it afterwards.

jfsantos:
My fix was a bit messy, so maybe it's better to think on another way of solving this. The big thing is that `DataStream` can't use anything coming from h5py (neither the handle nor the file id) as a state, since the `DataStream` instance is going to be serialized with the main loop and it keeps the state as an instance attribute. Maybe one way would be simply keeping the handle in a dictionary in H5PYDataset (as a class attribute) and pass the filepath or a hash of the handle object as a state.

dwf:
I doubt the hash would be preserved across interpreter sessions, if that is
the desired behaviour.

jfsantos:
You are right. Maybe the filename would be the safest bet?

vdumoulin:
I think you found the right solution, I'd just polish it a little more. How about this:

* The `state` passed around is a file path (using the full path for uniqueness).
* We maintain a private `_ref_counts` class attribute which is a `dict` mapping file paths to reference counts.
* We also maintain a private `_file_handles` class attribute which is a `dict` mapping file paths to HDF5 file handles.
* We implement `get_file_handle` and `release_file_handle` class methods which are responsible for maintaining the state of `_ref_counts` and `_file_handles`.

vdumoulin:
Also, if you're able to come up with a minimal bug-reproducing example we should make a unit test out of that.

jfsantos:
I'll go ahead and close this since the issue was fixed by #100

vdumoulin:
Good, thanks again for pointing out the issue!

PR #68, opened 2015-04-08
WIP: ILSVRC 2010 (a.k.a. ImageNet)

dwf:
Putting this up so that the interested can have a look. In particular looking for some feedback from @vdumoulin, as well as @bartvm for the right place we could split this up in order to speed it up with multiple processes.


vdumoulin:
Is ImageNet colloquial enough that we could use it as the name of the dataset without causing too much confusion?

dwf:
Well, as test data for later year competitions become available, we'd like
to support evaluation on those too. There's also the 22k class "ImageNet
data release" version for which there is no standard train/valid/test set,
so...

My hope is that a lot of the same code will be applicable to other
"ImageNet-family" challenges, but we should probably name the class after
the official task.

dwf:
@bartvm @vdumoulin I've got this converter into a near workable state now, and could use some eyes on it suggesting refactorings before I go and try to add unit tests.

I've adopted this kwargs-based approach to debug logging that I quite like, and think we could use a similar thing (or maybe a virtually identical thing) for passing information to a progress bar via logging calls.

vdumoulin:
You should eventually think of moving the code to `fuel.converters.ilsvrc2010`.

In order to support kwargs in the converter itself, we should eventually refactor converter functions to use subparsers like downloader functions do. I can review that if you choose to make it part of your PR.

vdumoulin:
Oh, nevermind, I just saw that this was already on your TODO list.

dwf:
Yes, though I have a slight refactor I'll propose for the dispatched function, something like

```python
from argparse import Namespace
from functools import wraps

def accept_namespace(func):
    @wraps(func)
    def wrapped(*args, **kwargs):
        if len(kwargs) == 0 and len(args) == 1 and isinstance(args[0], Namespace):
            return func(**vars(args[0]))
        else:
            return func(*args, **kwargs)
```

That way you can write functions like this 

```
@accepts_namespace
def foo(a, b, c):
    ...
```

and use it as a dispatch function for a subparser with matching argument names, while still having both a function that can be used programmatically without awkwardly instantiating a namespace and also can be documented more naturally.

vdumoulin:
I like that idea!

dwf:
@vdumoulin I've rewritten the history to put this file in fuel/converters/ (git filter-branch is neat). I also realized I hadn't included an entire file.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,732 @@
...
+    # Read what's necessary from the development kit.
+    devkit_path = os.path.join(directory, DEVKIT_ARCHIVE)
+    synsets, cost_matrix, raw_valid_groundtruth = read_devkit(devkit_path)
+
+    # Mapping to take WordNet IDs to our internal 0-999 encoding.
+    wnid_map = dict(zip((s.decode('utf8') for s in synsets['WNID']),
+                        xrange(1000)))
+
+    # Mapping to take ILSVRC2010 (integer) IDs to our internal 0-999 encoding.
+    # label_map = dict(zip(synsets['ILSVRC2010_ID'], xrange(1000)))
```
vdumoulin: Is it still useful?

dwf: Yep, I haven't pushed in a while but it is necessary for the valid and test
sets (yes, they distribute the ground truth for the valid and test sets
differently).

vdumoulin: But it looks like you're building that later on (`ilsvrc_id_to_zero_based`, lines 123-124).

dwf:
@vdumoulin A very rough guide to ZeroMQ: 
* A PUSH socket can have many PULLers, that's what's done with the ventilator/workers. Likewise, a PULL can have many PUSHers. This is what happens between workers and the sink.
* Atomicity is guaranteed, with a few caveats. That is to say, if you send a message on a zeromq socket, a client will get all of it or none of it. This also applies to sequences of messages where all-but-the-last message is sent with zmq.SNDMORE in the flags: this is used to tell zeromq that one or more further send calls are part of the same logical message, and should be received together. The only caveat I'm aware of is that if your process quits before all messages have been received, you may lose the last few messages; this is why I loop and sleep at the end of the ventilator function. (I'm not sure why this happens since zmq.LINGER should be infinite by default in my case...).
* The high water mark stuff is just to stop buffers accumulating ad infinitum.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+
+
+def train_set_ventilator(f, ventilator_port=5557, sink_port=5558,
+                         logging_port=5559, high_water_mark=10):
+    context = zmq.Context()
+    configure_zmq_process_logger(log, context, logging_port)
+    debug = make_debug_logging_function(log, 'VENTILATOR')
+    debug(status='START')
+    sender = context.socket(zmq.PUSH)
+    sender.hwm = high_water_mark
```
vdumoulin: Should we use `sender.set_hwm` setter instead?

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+#             return f(*args, **kwargs)
+#         except zmq.ZMQError as e:
+#             if e.errno == errno.EINTR:
+#                 # interrupted, try again
+#                 continue
+#             else:
+#                 # real error, raise it
+#                 raise
+
+def make_debug_logging_function(logger, process_type, **additional_kwargs):
```
vdumoulin: Could you add a docstring for this function?

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+               num_workers=6, worker_batch_size=1024):
+    """Converter for data from the ImageNet Large Scale Visual Recognition
+    Challenge (ILSVRC) 2010 competition.
+
+    Source files for this dataset can be obtained by registering at
+    [ILSVRC2010WEB].
+
+    Parameters
+    ----------
+    directory : str
```
vdumoulin: directory -> input_directory

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+
+
+def process_train_set(hdf5_file, train_archive, patch_archive,
+                      train_images_per_class, wnid_map, n_train, image_dim,
+                      num_workers, worker_batch_size):
+    """Process the ILSVRC2010 training set.
+
+    Parameters
+    ----------
+    hdf5_file : :class:`h5py.File` instance
```
vdumoulin: Could you complete this?

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+            sys.exit(1)
+
+
+def process_other_set(hdf5_file, archive, patch_archive, groundtruth,
+                      which_set, worker_batch_size, image_dim, offset):
+    """Process and convert either the validation set or the test set.
+
+    Parameters
+    ----------
+    hdf5_file : :class:`h5py.File` instance
```
vdumoulin: Idem for completion.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+            this_chunk = len(images)
+            features[start:start + this_chunk] = numpy.concatenate(images)
+            targets[start:start + this_chunk] = labels
+            filenames[start:start + this_chunk] = files
+            start += this_chunk
+            yield start
+
+
+def train_set_ventilator(f, ventilator_port=5557, sink_port=5558,
+                         logging_port=5559, high_water_mark=10):
```
vdumoulin: Could you add a docstring?

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+                debug(status='SENT_TAR', tar_filename=inner_tar.name,
+                      number=num)
+    while True:
+        time.sleep(10000)
+
+
+def train_set_worker(f, patch_images_path, wnid_map, images_per_class,
+                     image_dim=256, worker_batch_size=128,
+                     ventilator_port=5557, sink_port=5558, logging_port=5559,
+                     receiver_high_water_mark=10, sender_high_water_mark=10):
```
vdumoulin: Idem for docstring.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+                      "only found %d", os.getpid(), name.split('.')[0], label,
+                      images_per_class[label], total_images)
+        debug(status='FINISHED_TAR', tar_filename=name, number=num,
+              total=total_images, label=label)
+    debug(status='SHUTDOWN')
+
+
+def train_set_sink(hdf5_file, num_images, images_per_class,
+                   flush_frequency=256, shuffle_seed=(2015, 4, 9),
+                   sink_port=5558, logging_port=5559, high_water_mark=10):
```
vdumoulin: Idem for docstring.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+        numpy.concatenate((c,
+                           -numpy.ones(new_dtype['children'].shape[0] - len(c),
+                                       dtype=numpy.int16)))
+        for c in children
+    ]
+    new_synsets['children'] = padded_children
+    return new_synsets, cost_matrix
+
+
+def extract_patch_images(f, which_set=None):
```
vdumoulin: Why not make `which_set` an `arg`, since you explicitly require that it is either `'train'`, `'valid'` or `'test'`?

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+            filename = tokens[-1]
+            image = _imread(tar.extractfile(info_obj.name))
+            patch_images[filename] = image
+    return patch_images
+
+
+import progressbar
+
+
+class ProgressBarHandler(logging.Handler):
```
vdumoulin: Idem for docstring.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,826 @@
...
+        self.curr_val_attr = curr_val_attr
+        self.widgets = widgets
+        self.start_predicate = (start_predicate if start_predicate is not None
+                                else lambda x: hasattr(x, self.max_val_attr))
+        self.update_predicate = (update_predicate
+                                 if update_predicate is not None
+                                 else lambda x: hasattr(x, self.curr_val_attr))
+        self.level = level
+
+    def handle(self, record):
```
vdumoulin: Idem for docstring.

fuel/utils/logging.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Parameters
+    ----------
+    logger : object
+        Logger-like object with a `handle()` method similar that
+        accepts :class:`logging.LogRecord` instances.
+    processes : sequence, optional
+        Collection containing :class:`multiprocessing.Process` objects.
+        The loop will continue until none of these processes is alive.
+        If empty (default), this
```
vdumoulin: Sentence is truncated.

vdumoulin:
@dwf I've gone through a superficial review of docstrings. I'm still reading the code to give it an in-depth review.

fuel/utils/logging.py
```diff
@@ -0,0 +1,140 @@
...
+
+    Notes
+    -----
+    Mutates the logger object by removing any existing handlers,
+    setting the `propagate` attribute to `False`, and adding a
+    :class:`ZMQLoggingHandler` set up to log messages to a socket
+    connected on `logging_port`.
+
+    """
+    logger = logging.getLogger(__name__)
```
vdumoulin: Why do you need `logger` as argument when you immediately assign it here?

dwf: Good catch. 37e5f2b.

fuel/utils/logging.py
dwf: 6e4ffbe

fuel/converters/ilsvrc2010.py
dwf: Probably going to move this to some utility module.

fuel/converters/ilsvrc2010.py
dwf: a69b35b.

fuel/converters/ilsvrc2010.py
dwf: `hwm` is a property which invokes `set_hwm` anyway, so no real difference.

fuel/converters/ilsvrc2010.py
dwf: b0d6b45

fuel/converters/ilsvrc2010.py
dwf: 8a931b7

fuel/converters/ilsvrc2010.py
dwf: 65f3e52

fuel/converters/ilsvrc2010.py
dwf: 65f3e52

fuel/converters/ilsvrc2010.py
dwf: 65f3e52

fuel/converters/ilsvrc2010.py
dwf: good point.

fuel/converters/ilsvrc2010.py
vdumoulin: Good to know!

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,908 @@
...
+    """Process and convert either the validation set or the test set.
+
+    Parameters
+    ----------
+    hdf5_file : :class:`h5py.File` instance
+        HDF5 file handle to which to write. Assumes `features`, `targets`
+        and `filenames` already exist and are at least as long as
+        `offset` plus the number of files in `archive`.
+    archive : str or file-like object
+    patch_archive : str or file-like object
```
vdumoulin: I'll allow myself to be annoying and request more information for the two arguments above.

dwf: No, it's fine, I meant to actually write this but didb5.

dwf:
@vdumoulin I just pushed a pretty major update. Lots of new docstrings, and better factored.

fuel/converters/ilsvrc2010.py
dwf: 2f7a812

fuel/converters/ilsvrc2010.py
dwf: 6265a52

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,956 @@
...
+        so that a sequential read through the training set is shuffled
+        by default.
+    shuffle_seed : int or sequence, optional
+        Seed for a `numpy.random.RandomState` used to shuffle the training
+        set order.
+    num_workers : int, optional
+        The number of worker processes to deploy.
+    worker_batch_size : int, optional
+        The number of images the workers should send to the sink at a
+        time.
```
vdumoulin: Add documentation for `output_filename`.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,956 @@
...
+    Parameters
+    ----------
+    patch_images_archive : str or file-like object
+        The path or file-handle from which to read the patch images
+        archive.
+    wnid_map : dict
+        A dictionary mapping WordNet IDs to class indices.
+    images_per_class : sequence
+        A sequence containing the number of images in each class,
+        with as many elements as there are classes.
```
vdumoulin: We're missing doc for `image_dim` and `worker_batch_size`.

fuel/utils/logging.py
```diff
@@ -0,0 +1,224 @@
...
+    """A :class:`~logging.Handler` that drives a progress bar.
+
+    Parameters
+    ----------
+    max_val_attr : str
+        The attribute name used for the maximum value of the
+        progress bar, used to trigger a new progress bar being started.
+    curr_val_attr : str
+        The attribute name used for the current value of the progress
+        bar, used to trigger an update to the progress bar.
```
vdumoulin: We're missing doc for `widgets`.

fuel/utils/logging.py
```diff
@@ -0,0 +1,224 @@
...
+
+def zmq_log_and_monitor(logger, context, processes=(), logging_port=5559,
+                        failure_threshold=logging.CRITICAL):
+    """Feed `LogRecord`s received on a ZeroMQ socket to a logger.
+
+    Parameters
+    ----------
+    logger : object
+        Logger-like object with a `handle()` method similar that
+        accepts :class:`logging.LogRecord` instances.
```
vdumoulin: We're missing doc for `context`.

vdumoulin:
I think you can cross 'split metadata' off the TODO list.

fuel/utils/logging.py
dwf: bdae911

fuel/utils/logging.py
dwf: 774eb4b

dwf: 774eb4b

fuel/converters/ilsvrc2010.py
dwf: 7f07f890

tests/converters/test_convert_ilsvrc2010.py
```diff
@@ -0,0 +1,601 @@
...
+                                data=image.tobytes())
+    jpeg_data = io.BytesIO()
+    pil_image.save(jpeg_data, format='JPEG')
+    return jpeg_data.getvalue()
+
+
+def create_fake_jpeg_tar(seed, min_num_images=5, max_num_images=50,
+                         min_size=20, size_range=30, filenames=None,
+                         random=True, gzip_probability=0.2):
+    """Create a TAR file of ranodmly generated JPEG files.
```
vdumoulin: Typo: ranodmly

dwf: 5b5194

dwf:
@lamblin @dmitriy-serdyuk This is ready for a preliminary review, though it requires some more tests before merge.

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,570 @@
...
+                       output_filename='ilsvrc2010.hdf5',
+                       shuffle_seed=config.default_seed):
+    """Converter for data from the ILSVRC 2010 competition.
+
+    Source files for this dataset can be obtained by registering at
+    [ILSVRC2010WEB].
+
+    Parameters
+    ----------
+    input_directory : str
```
vdumoulin: `input_directory -> directory`

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,570 @@
...
+    return (output_path,)
+
+
+def fill_subparser(subparser):
+    """Sets up a subparser to convert the ILSVRC2010 dataset files.
+
+    Parameters
+    ----------
+    subparser : :class:`argparse.ArgumentParser`
+        Subparser handling the `mnist` command.
```
vdumoulin: `mnist -> ilsvrc2010`

fuel/converters/ilsvrc2010.py
```diff
@@ -0,0 +1,570 @@
...
+            if patched:
+                num_patched += 1
+            socket.send_pyobj((filename, class_index), zmq.SNDMORE)
+            socket.send(image_data)
+    if num_patched != len(patch_images):
+        raise Exception
+
+
+def load_from_tar_or_patch(tar, image_filename, patch_images):
+    """Do everything necessary to process a image inside a TAR.
```
vdumoulin: `a image -> an image`?

fuel/datasets/imagenet.py
```diff
@@ -0,0 +1,42 @@
...
+       Khosla, Michael Bernstein, Alexander C. Berg and Li Fei-Fei.
+       *ImageNet Large Scale Visual Recognition Challenge*. IJCV, 2015.
+
+    .. [SIGNUP] http://www.image-net.org/signup
+
+    Parameters
+    ----------
+    which_sets : tuple of str
+        Which split to load. Valid values are 'train' and 'test',
+        corresponding to the training set (50,000 examples) and the test
```
vdumoulin: I'm pretty sure there are more than 50,000 training examples ;).

fuel/transformers/defaults.py
```diff
@@ -1,8 +1,40 @@
...
+
+    Notes
+    -----
+    Used for retrieving variable-length byte data stored as, e.g. a uint8
+    ragged array.
+
+    """
+    def __init__(self, stream, **kwargs):
+        kwargs.setdefault('produces_examples', stream.produces_examples)
+        axis_labels = stream.axis_labels
```
vdumoulin: I think this may crash if `stream.axis_labels is None`.

fuel/transformers/defaults.py
```diff
@@ -1,8 +1,40 @@
...
+    def transform_source_example(self, example, _):
+        return example.tostring()
+
+    def transform_source_batch(self, batch, _):
+        return [example.tostring() for example in batch]
+
+
+def rgb_images_from_encoded_bytes(which_sources):
+    return ((ToBytes, [], {'which_sources': ('encoded_images',)}),
+            (ImagesFromBytes, [], {'which_sources': ('encoded_images',)}))
```
vdumoulin: Should we rename the `encoded_images` source to `images` as well?

fuel/utils/parallel.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def _spawn_producer(f, port, addr='tcp://127.0.0.1'):
+    """Start a process that sends results on a PUSH socket.
+
+    Parameters
+    ----------
+    f : callable
+        Callable that takes a single argument, a handle
+        for a ZeroMQ PUSH socket. Must be picklable.
```
vdumoulin: Documentation for `port` and `addr` is missing here.

fuel/utils/parallel.py
```diff
@@ -0,0 +1,110 @@
...
+    """A producer-consumer pattern.
+
+    Parameters
+    ----------
+    producer : callable
+        Callable that takes a single argument, a handle
+        for a ZeroMQ PUSH socket. Must be picklable.
+    consumer : callable
+        Callable that takes a single argument, a handle
+        for a ZeroMQ PULL socket.
```
vdumoulin: Documentation for `addr` is missing here.

vdumoulin:
@dwf I did a quick first pass. I'm not very familiar with ZMQ, so my review for this part of the PR will focus on annoying docstring remarks.

rodrigob:
this seems to be moving well. looking forward for giving it a trial.

dwf:
Unfortunately storing the data like this comes at a cost, in the sense that
you do incur the cost of the JPEG decoding. You can pipeline it away with
ServerDataStream of course, to an extent. Libjpeg-turbo could help with
this, but it's not super easy to just substitute it in to PIL. Ideally
someone would go and write a batteries included Cython wrapper around the
non-legacy API for libjpeg-turbo (much nicer/less stupid than the legacy
libjpeg API), but that should probably be a separate package.

rodrigob:
From [what I read](https://github.com/python-pillow/Pillow/issues/1037) PIL will use libjpeg-turbo if it is installed in the system. So that feature is already in there.
Otherwise there is [jpeg4py](https://pypi.python.org/pypi/jpeg4py/) which also seems to implement what you are saying.
Do you know if "PIL using jpeg-turbo" is faster/slower at decoding than jpeg4py ? 
(maybe [this would help](https://github.com/ajkxyz/jpeg4py/blob/master/tests/benchmark.py))

dwf:
@rodrigob jpeg4py is indeed exactly what I was thinking should exist. Thank you for pointing that out and thank you Samsung!

fuel/transformers/defaults.py
dwf: This actually should be using which_sources from the argument, eh?

vdumoulin: Yes

fuel/converters/ilsvrc2010.py
dwf: Done.

fuel/converters/ilsvrc2010.py
dwf: ae3ad17

fuel/datasets/imagenet.py
dwf: 114bc41

fuel/transformers/defaults.py
dwf: b823233

fuel/utils/parallel.py
dwf: Already done.

fuel/utils/parallel.py
dwf: ea0121e

vdumoulin:
@dwf and I did an offline test converting the ImageNet dataset, and it works! Coverage is down 1.2%, but considering the importance of this PR, I'm ready to merge right away.

PR #69, opened 2015-04-08
Add `Spiral` toy dataset

jbornschein:
This adds a simple, 2-dimensional "Spiral" toy dataset.

I think simple datasets like this are valuable in certain situations and we might want to include a few (e.g. swiss-roll, etc.).

In the case we want to merge this:

* should the module be called `fuel.datasets.toy`? Or rather `fuel.datasets.small`? Or something else?
* this PR activates the `matplotlib.sphinxext.plot_directive` Sphinx extension to visualize the dataset in the documentation. I believe a visualization is helpfull here. Is this something we want?

### DIFF
--- a/docs/api/index.rst
+++ b/docs/api/index.rst
@@ -1,0 +1,12 @@
+API Reference
+=============
+
+.. warning::
+
+   This API reference is currently nothing but a dump of docstrings, ordered
+   alphabetically.
+
+.. toctree::
+    :glob:
+
+    *
--- a/docs/api/toy-datasets.rst
+++ b/docs/api/toy-datasets.rst
@@ -1,0 +1,8 @@
+Toy datasets 
+============
+
+.. automodule:: fuel.datasets.toy
+    :members:
+    :undoc-members:
+    :show-inheritance:
+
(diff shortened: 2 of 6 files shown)


dwf:
Hmm. Are you planning to publish on this? Otherwise, it might be better to use existing datasets, such as [the swiss roll from the Isomap paper](http://isomap.stanford.edu/datasets.html)?

jbornschein:
Swiss roll is certainly the next one I would add. 

But I also saw a couple of publications where 2d spiral was used -- the last one being [Deep Unsupervised Learning using Nonequilibrium Thermodynamics](http://arxiv.org/abs/1503.03585).

jbornschein:
I'd consider this in a mergable state.

Any comments?


The SwissRoll dataset in this PR is normalized to abs(features) <= 1, which is different from the normalization used in the isomap paper (abs(features) <= 27) and which again seems different from normalizations used in various other implementations (e.g. scikit-learn: https://github.com/scikit-learn/scikit-learn/blob/bb39b49/sklearn/datasets/samples_generator.py#L1228)

jbornschein:
Ping?!

Are we interested in this?

bartvm:
Yeah, this looks good to me! Matplot visualizations are fine by me, but we'll have to wait and see whether ReadTheDocs will take them I guess.

PR #71, opened 2015-04-09
Merge transformer returns iterator with dictionary

harmdevries89:
Merge transformer is now able to return iterator in dictionary form. You can probably avoid making a separate class/function to create the dictionary, but didn't really know how to do it.

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -2,7 +2,7 @@ from abc import ABCMeta
 from multiprocessing import Process, Queue
 
 import numpy
-from picklable_itertools import chain, ifilter, izip
+from picklable_itertools import chain, ifilter, izip, imap
 from picklable_itertools.extras import partition
 from six import add_metaclass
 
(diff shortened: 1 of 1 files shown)


fuel/transformers/__init__.py
```diff
@@ -420,7 +420,22 @@ def get_epoch_iterator(self, **kwargs):
         batches = chain.from_iterable(
             izip(*[data_stream.get_epoch_iterator()
                    for data_stream in self.data_streams]))
-        return partition(len(self.sources), chain.from_iterable(batches))
+
+        part = partition(len(self.sources), chain.from_iterable(batches))
+        as_dict = kwargs.get('as_dict', False)
+        if as_dict:
+            return imap(make_dict(self.sources), part)
```
bartvm: Does something like `imap(dict, imap(zip, izip(part, repeat(self.sources)))` work? Is a bit more straightforward than having a separate class.

harmdevries89:
No, it doesn't. I think the ```imap(zip, ...)``` doesn't really do what ```zip(*x)``` does.

harmdevries89:
For future reference: ```starmap``` can apply a function mapping while unpacking its arguments.

PR #73, opened 2015-04-12
Fix nondeterministic behaviour in H5PYDataset

vdumoulin:
This should fix the broken build on master branch.


fuel/datasets/hdf5.py
```diff
@@ -159,10 +159,12 @@ def create_split_array(split_dict):
         """
         # Determine maximum split, source and string lengths
         split_len = max(len(split) for split in split_dict)
-        sources = set()
+        sources = []
         comment_len = 1
         for split in split_dict.values():
-            sources |= set(split.keys())
+            for source in split.keys():
```
dwf: add sorted() around split.keys()

dwf:
Merging, coverage drop notwithstanding.

PR #75, opened 2015-04-13
Tutorial for creating new built-in datasets

vdumoulin:
After this is merged, I think we'll be able to delegate adding / refactoring built-in datasets much more easily.

### DIFF
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -74,6 +74,8 @@ things if your data happens to meet these assumptions:
     the time. Rest assured, :class:`~.datasets.hdf5.H5PYDataset` has an
     option to load things into memory which we will be covering soon.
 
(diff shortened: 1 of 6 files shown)


vdumoulin:
@bartvm @dwf With this tutorial, I think we'll have all we need to start assigning the task of refactoring an existing built-in dataset or add a new commonly-used dataset to Fuel.

bartvm:
Looks good to me! The only thing I would consider adding is basically turning the entire tutorial in a series of unit tests (basically just copy and paste the tutorial I guess?). However, we can refer that to a later date if you prefer merging this soon.

vdumoulin:
I'd prefer merging sooner. I'll create an issue for that.

vdumoulin:
I opened the issue (#87) for the unit tests, I'll merge now and we'll address this later.

PR #76, opened 2015-04-14
Reduce size of converted BinarizedMNIST dataset

vdumoulin:


vdumoulin:
Agreed, I'll add a unit test to make sure this gets caught.

vdumoulin:
Turns out the tests were skipped because I forgot to change `'binarized_mnist'` to `'binarized_mnist.hdf5'` in the call to `skip_if_not_available`.

I added a couple goodies while I was at it. I reduced the tests' run time by keeping the dataset off memory and only loading a small subset of it to do the tests (which will be necessary for very large datasets), and I added checksum verification for that subset, which should increase our confidence that we haven't messed up the data conversion. (There's a nice typo in my commit message: I typed 'sumcheck' instead of 'checksum'. I'll leave it there for posterity.)

I think that's something we should encourage for unit tests, maybe I should add a section mentioning good unit testing practices for new built-in datasets in my tutorial?

vdumoulin:
@dwf Tests pass.

PR #78, opened 2015-04-14
Add axes semantics in H5PYDataset

vdumoulin:


tests/test_binarized_mnist.py
```diff
@@ -40,6 +40,16 @@ def test_binarized_mnist_test():
...
 
 
+def test_binarized_mnist_axes():
+    skip_if_not_available(datasets=['binarized_mnist.hdf5'])
+
+    dataset = BinarizedMNIST('train', load_in_memory=False)
+    assert all(
+        dim_label == label for (dim_label, label) in
+        zip(dataset.axis_label_dict['features'],
+            ('batch', 'channel', 'axis_0', 'axis_1')))
```
lamblin: `zip` will stop at the shortest sequence, so the test would pass if `dataset.axis_label_dict['features']` was an empty list or tuple.

fuel/utils.py
```diff
@@ -87,3 +88,24 @@ def __getstate__(self):
...
+    if axis_label == 'b':
+        return 'batch'
+    elif axis_label == 'c':
+        return 'channel'
+    elif axis_label == 't':
+        return 'time'
+    elif re.match('^[0-9]+$', axis_label):
+        return 'axis_' + axis_label
+    else:
+        return axis_label
```
lamblin: Should we complain if it is not a string or a basestring? For instance, if someone calls `expand_axis_label(0)` instead of `'0'`?

fuel/utils.py
```diff
@@ -87,3 +88,24 @@ def __getstate__(self):
...
+        Abbreviated axis label
+
+    """
+    if axis_label == 'b':
+        return 'batch'
+    elif axis_label == 'c':
+        return 'channel'
+    elif axis_label == 't':
+        return 'time'
+    elif re.match('^[0-9]+$', axis_label):
```
bartvm: I'm not sure we need to add numbers to the axes. My idea is that these labels add *semantics*, and "axis 0" is about as non-semantic as it gets.For images the labels could be `w` for width and `h` for height, etc. Otherwise, if there are no semantics, then they should jsut be `f` for features, and this label can be repeated.

vdumoulin: You make a good point. @lamblin what do you think?

lamblin: Sure, 'w' and 'h' for images makes sense.
'f' for 'features' works as well, or we could just let it blank.

vdumoulin:
@lamblin @bartvm Your comments have been addressed. Tests are passing.

bartvm:
Is there any specific reason we insist on one-letter abbreviations by the way? Otherwise using the full words, `batch`, `height`, etc. might be a more explicit and clear, and save people from having to dig into the code or documentation to find out what `b` or `h` stands for.

vdumoulin:
It's mostly out of habit, I guess. I'm used to seeing convolution formats being described as `bc01` or `b01c` or `c01b`, so initially it's what I wanted to reflect in the code.

By the way, the only place where axis labels *can* be abbreviated is within the HDF5 file itself: `H5PYDataset` always expands these abbreviations when extracting axis label information from the HDF5 file.

Whether or not it's useful enough to keep is another story; I wouldn't necessarily be against removing this feature altogether and establish a convention for axis labels that systematically uses full words.

lamblin:
I agree with @vdumoulin, it was mostly a question of habit.
The Zen of Python says "Explicit is better than implicit", so let's use full words.

vdumoulin:
@bartvm @lamblin I removed label abbreviation. Tests are passing (0.05% decrease in test coverage aside).

fuel/datasets/hdf5.py
```diff
@@ -233,6 +233,18 @@ def split_dict(self):
         return self._split_dict
 
     @property
+    def axis_label_dict(self):
```
dwf: Personally I'd just call this `.axis_labels`.

vdumoulin: Fair enough. I'll change that.

vdumoulin:
@dwf I changed `axis_label_dict` for `axis_labels`. The tests are passing except for a 0.05% test coverage decrease. I looked into why it decreased, and it turns out I added two lines in `fuel.converters.binarized_mnist` (which is not yet covered by unit tests) which causes the coverage to decrease.

bartvm:
LGTM

tests/test_hdf5.py
```diff
@@ -56,6 +56,21 @@ def test_h5py_dataset_split_parsing():
     os.remove('tmp.hdf5')
 
 
+def test_h5py_dataset_axis_labels():
+    h5file = h5py.File(name='tmp.hdf5', mode="w")
```
bartvm: @vdumoulin Could you wrap this into a `try-finally` block? Right now `tmp.hdf5` won't be removed if the test fails for any reason.

vdumoulin: Done.

Issue #80, opened 2015-04-17
H5PYDataset should load lazily (and probably not call load() in constructor)


dwf:
Isn't that the point of `do_not_pickle_attributes` after all?

vdumoulin:
Fixed by #83.

PR #83, opened 2015-04-23
[WIP] Add axis labels to data streams

vdumoulin:
I'm starting the PR early to get comments on the approach I've chosen (in particular from @lamblin @dwf @bartvm).

The `axis_labels` attribute (defaulting to `None`) is introduced in `Dataset` and `AbstractDataStream`. Like in `H5PYDataset`, this attribute should be a dict mapping source name to tuple of axis labels.

By convention, if a source doesn't appear in `axis_labels`, it means there's no axis label for this source (or, equivalently, its label is `''`). `None` is treated as an empty dict.

`Transformer` subclasses replace the `axis_labels` attribute of their wrapped data stream by `None` by default, because we can't know what a transformation will do to the axis semantics. Subclasses can override that property; as an example, I've implemented the property for `ForceFloatX`, which trivially passes its wrapped data stream's `axis_labels` attribute.

### DIFF
--- a/fuel/datasets/base.py
+++ b/fuel/datasets/base.py
@@ -46,7 +46,7 @@ class Dataset(object):
     """
     provides_sources = None
 
-    def __init__(self, sources=None):
+    def __init__(self, sources=None, axis_labels=None):
         if not self.provides_sources:
             raise ValueError("dataset does not have `provides_sources`")
         if sources is not None:
@@ -54,6 +54,7 @@ class Dataset(object):
                                       for source in sources):
                 raise ValueError("unable to provide requested sources")
             self.sources = sources
+        self._axis_labels = axis_labels
 
     @property
     def sources(self):
@@ -64,6 +65,14 @@ class Dataset(object):
     @sources.setter
     def sources(self, sources):
         self._sources = sources
+
+    @property
+    def axis_labels(self):
+        return self._axis_labels
+
+    @axis_labels.setter
+    def axis_labels(self, axis_labels):
+        self._axis_labels = axis_labels
 
     @property
     def example_iteration_scheme(self):
--- a/fuel/datasets/base.py
+++ b/fuel/datasets/base.py
@@ -54,7 +54,7 @@ class Dataset(object):
                                       for source in sources):
                 raise ValueError("unable to provide requested sources")
             self.sources = sources
-        self._axis_labels = axis_labels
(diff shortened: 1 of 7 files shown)


vdumoulin:
As a side note, I'm sneaking a fix for #80 in this PR.

fuel/datasets/base.py
```diff
@@ -66,6 +67,14 @@ def sources(self, sources):
         self._sources = sources
 
     @property
+    def axis_labels(self):
+        return self._axis_labels
```
bartvm: This property doesn't seem to add much over a normal attribute.

bartvm:
LGTM. A bit of documentation with MNIST as an example would be nice at one point, but we can leave that for a future PR.

fuel/datasets/base.py
vdumoulin: Agreed. I set it as an attribute instead.

vdumoulin:
I opened an issue (#99) for adding documentation and addressed your inline comment. If I merge when the tests pass, is it alright with you?

bartvm:
Yes, although right now the tests are failing because `MarkovChainDataset` doesn't call super. I think it should probably be compulsory to call `super`, so updating `MarkovChainDataset` would be the way to go then.

vdumoulin:
I'm on it.

vdumoulin:
Tests pass on both sides, I'll merge.

dwf:
@vdumoulin @bartvm is there anything automatic that can be done for `DataStream`, if the dataset it is wrapping has axis information? Should it essentially strip off `batch`, since it seems to default to an example-wise stream?

Currently there's no way to add axis information to a `Batch` stream, other than setting the attribute after construction (see #134).

dwf:
@vdumoulin @bartvm Actually, even that doesn't work, since it's a read-only property! FML

bartvm:
I think this should probably have the same approach as was discussed for `num_examples` and `num_batches`: each `Transformer` and `DataStream` should have an `axis_labels` property that determines the axis labels of the stream based on the wrapped stream or dataset, and its own configuration. For example, for `DataStream` it depends on the iteration scheme used whether there is a batch axis, right? And for more complicated `Transformer`s the labels could change entirely (i.e. flattening).

vdumoulin:
Agreed.

PR #90, opened 2015-04-29
Introduce default batch iterator and default transformers

vdumoulin:
Following an offline discussion with @lamblin, I propose to introduce a notion of "default transformers" to `Dataset` subclasses.

The problem we're trying to solve is the following: suppose that we have a dataset of 8-bit-per-channel images that we want to add to Fuel's built-in datasets. Here are three approaches for storing this data on disk:

* Store the data as `float32`, with pixel values between 0.0 and 1.0. Files are four times larger than they could be, which can be problematic for very large datasets, but it avoids complications on the implementation side.
* Store the data as `uint8`. This leads to minimal file sizes, which could be important for very large datasets. However, since each pixel value is represented as an integer between 0 and 255, some preprocessing  (dividing by 255.0, casting as `float32`) is necessary so that the data is ML-friendly. Two approaches can be taken:
    - Document the data format properly, and leave it to the user to do the conversion with `Transformer`s. This is quite simple on the implementation side, but it's not very user-friendly, since most of the time users have more boilerplate code to write.
    - Let the dataset define a default transformation to the data, and define a canonical way for the users to access an iterator that will apply this default transformation.

We concluded that the best approach would be the latter one, as it combines compactness and user-friendliness.

In order to keep it simple for developers to implement new datasets, we figured that the best place to put this code was in the `Dataset` abstract class. Here's how it works:
* We introduced a `apply_default_transformers` method that takes a data stream as input and applies a series of `Transformer`s to it. ~~By default, it simply returns the data stream passed as argument. Subclasses can override this method to define default transformations to apply, such as dividing by 255.0 and casting to `float32'~~. In addition, there's now a `default_transformers` property which is a tuple of tuples containing

    - the class to instantiate,
    - the args other than the data stream to pass to the constructor, and
    - the kwargs to pass to the constructor.

The method will instantiate each transformer sequentially by passing it the previous transformer as the data stream argument, starting from the data stream received as input.
* ~~In addition to `example_iteration_scheme`, we defined a `batch_iteration_scheme` attribute which subclasses can set to implement a sane iteration scheme default.~~
* ~~We introduced a `get_batch_stream` method which takes an iteration scheme as argument, defaulting to `None`. If no iteration scheme is passed, `self.batch_iteration_scheme` is used. The method then instantiates a `DataStream` using the chosen iteration scheme and calls `self.apply_default_transformers` to apply default transformations.~~
* A `default_stream` class method was added to `DataStream` as the preferred way of instantiating a data stream with the default transformers applied.


vdumoulin:
I think the PR is now in a state where I'd be comfortable to merge if it looks good to all parties involved.

fuel/datasets/cifar10.py
```diff
@@ -33,23 +34,12 @@ class CIFAR10(H5PYDataset):
...
-        if which_set not in ('train', 'test'):
-            raise ValueError("CIFAR10 only has a train and test set")
-
         kwargs.setdefault('load_in_memory', True)
-        self.which_set = which_set
-
-        super(CIFAR10, self).__init__(self.data_path, which_set,
-                                      **kwargs)
+        super(CIFAR10, self).__init__(self.data_path, which_set, **kwargs)
+        self.default_transformers += (
```
bartvm: Maybe set this as a class attribute? I guess it will never be different between instances, and it might be useful at one point to do something like `apply_transformers(my_own_dataset, CIFAR10.default_transformers)` i.e. to be able to access the default transformers from a dataset without having to instantiate it.

bartvm:
Except for my comment about it being useful to have access to the default transformers of a dataset without the need to instantiate it (so that you can easily apply the same set of transformations to other datasets), the rest of it looks good to me.

fuel/datasets/cifar10.py
vdumoulin: It might be hard to do consistently, since datasets sometimes rely on constructor arguments to define the default transformers (e.g. `H5PYDataset` sets `Flatten` as the default transformer if `flatten is True`, which is only known when it is instantiated).

bartvm: Mmm... But is `flatten` really a "default transformer" then? I imagined "default transformers" as being a completely fixed set of transformers for a given dataset, whereas flattening in this case is optional and chosen by the user.

I guess I would still prefer the solution then where CIFAR-10 sets its default transformers as a class attribute, and `H5PYDataset` extends that tuple if necessary in the constructor. But maybe we should reconsider having "flatten" as  keyword argument entirely. It can either be part of the default transformers, or not, but I guess that it's arguably feature envy if we let the dataset choose what transformers will be applied through a keyword argument; it blurs the separation between data loading and transformation that data sets and transformers was supposed to introduce.

vdumoulin: Actually, I'm wondering if we even need things like the `flatten` constructor argument in `H5PYDataset` anymore. It's nothing more than a convenience flag for applying a transformer.

Is it reasonable to expect that users that want to flatten topological data do so explicitly on their own when constructing the data stream?

If so, then we could restrict the use default transformers to cases where the data is stored in a certain way but it makes more sense to use it under another representation, which would work well as a class attribute.

vdumoulin: Oops, didn't see your comment when posting mine. I guess we agree, then.

bartvm: Yes, I think it's completely reasonable, and in cases where it should be done 90% of the time, we can always just add it to the default transformers (and then for the 10% that doesn't want to use it they can apply the transformers manually). I think I'd much prefer the class attribute approach, keeps the dataset / transformer divide a bit clearer.

vdumoulin:
There's one thing that I'll have to fix before it's ready: in Python 2.7, pickling fails because I'm passing around instancemethod objects in `SingleMapping` subclasses.

vdumoulin:
I've rebased and the tests pass.

The only other thing I've changed is creating a new `fuel.transformers.defaults` module which contains functions for creating commonly-used default transformers (such as mapping `uint8` pixel values to the unit interval).

vdumoulin:
@bartvm Do you have other comments to formulate regarding this PR?

bartvm:
Nope, LGTM!

PR #91, opened 2015-04-29
Fix H5PYDataset bug for requesting an unordered list of indices

vdumoulin:


vdumoulin:
@dwf @bartvm This is a quick one for fixing a bug reported in the Fuel mailing list, would any of you mind to review it?

fuel/datasets/hdf5.py
```diff
@@ -316,11 +316,14 @@ def _out_of_memory_get_data(self, state=None, request=None):
             if isinstance(request, slice):
                 request = slice(request.start + subset.start,
                                 request.stop + subset.start, request.step)
+                data = state[source_name][request]
             elif isinstance(request, list):
                 request = [index + subset.start for index in request]
+                sorted_request = numpy.sort(request)
+                sorted_data = state[source_name][sorted_request, ...]
+                data = sorted_data[numpy.argsort(numpy.argsort(request))]
```
bartvm: Doing 3 sorts seems a bit crazy. I guess you could two:
```python
indices = numpy.argsort(request)
reverse_indices = numpy.argsort(indices)
sorted_data = state[source_name][request[indices], ...]
data = sorted_data[reverse_indices]
```
Maybe there's a way of doing it with only 1 as well?

bartvm:
So this approach is clean in the sense that it is transparent to the user, but it's not very efficient, because you end up sorting the indices (twice, to get a reverse mapping) and then need to shuffle the data in place after reading.

An alternative/supplementary solution would be to create a `sorted_indices=True` keyword argument for the `ShuffledScheme`. That way the only thing that is done is that the indices are sorted once. Given that the order of examples in your minibatches doesn't matter (which it doesn't in 90% of the cases) this is the only thing you really need to do.

vdumoulin:
Fair enough.

bartvm:
For usability I guess that the following might be a good idea: By default `sorted_indices=False` for the shuffled scheme, but `sort_indices=True` for `H5PYDataset`. This means that it would work exactly like you would expect, but it's a bit on the slow side. Power users who bother to read the code/documentation can then set `sorted_indices=True` and `sort_indices=False` for a speedup.

vdumoulin:
I pushed @bartvm's suggested fix.

fuel/datasets/hdf5.py
dwf: What if you pre-allocated `data` and did

`data[indices] = state[source_name][request[indices], ...]`

vdumoulin:
I don't know about the `sort_indices` /  `sorted_indices` differentiation.

I made a quick benchmark:

```python
import time

import numpy
from fuel.schemes import ShuffledScheme


dataset = IndexableDataset(numpy.random.uniform((100000, 100)))

start_time = time.time()
for i in xrange(1000000):
    stream = DataStream(
        dataset=dataset,
        iteration_scheme=ShuffledScheme(100000, 10000, sorted_indices=False))
stop_time = time.time()
print 'Unsorted indices: {} seconds'.format(stop_time - start_time)

start_time = time.time()
for i in xrange(1000000):
    stream = DataStream(
        dataset=dataset,
        iteration_scheme=ShuffledScheme(100000, 10000, sorted_indices=True))
stop_time = time.time()
print 'Sorted indices:   {} seconds'.format(stop_time - start_time)
```

On my machine, the timings are roughly 9.76 seconds for the unsorted indices and 9.78 seconds for the sorted indices. This is with a batch size of 10,000.

Wouldn't we be splitting hair for almost no speedup?

dwf:
Try it with lots more small batches. If we still don't see a difference then that's quite something.

vdumoulin:
Oh, I made the stupidest mistake: I have to **iterate** over my streams if I want to see something...

vdumoulin:
Alright, I updated the example code above. With a batch size of B, the timings are:

* B=10: 27.03 / 27.86 (1.03 times slower)
* B=100: 16.90 / 18.58 (1.10 times slower)
* B=1000: 15.02 / 17.78 (1.12 times slower)
* B=10000: 15.48 / 17.99 (1.16 times slower)

That's when nothing is done with the batches; with fprop and backprop the slowdowns will likely be more negligible.

bartvm:
I'm a bit confused as to what your point is. To summarize, there are two options:

* Sorting the indices as part of the iteration scheme
* Sorting the indices in `H5PYDataset`, and then unsorting the data

The first one is fast (as your benchmarks show) but requires user intervention to stop `H5PYDataset` from crashing. The second one is the one I am more worried about performance-wise, because it needs 2/3 calls to sort, and might end up copying large amounts of data when re-shuffling the batch itself.

My suggestion was to implement both i.e. keep the original implementation of the PR as well as the current one; the first one being user-friendly, the second one being fast. To see whether that is worth it would require benchmarking the original implementation as well. If it turns out the original implementation has a negligible slowdown (which I doubt) then there might not be a need for the faster version.

vdumoulin:
My point is: let's leave sorting the indices in `ShuffledScheme` on by default, as it is fast, and power-users that either want unsorted indices or faster computation can turn that off manually.

bartvm:
Aah, okay, gotcha. I'm ambivalent... On the one hand, it seems like unexpected behavior to sort indices, and although minor, it could be a noticeable slowdown for small, very fast models (5-10%). On the other hand, otherwise a lot of users might be doing unnecessarily slow iterations over HDF5 datasets.

Maybe see what other people think, I'm undecided :)

vdumoulin:
Another option would be to simply refuse to get data when indices are not sorted, so that at least users get a meaningful error message, and implement something like `batchwise_shuffled_sequential` in Pylearn2 that `H5PYDataset` users can use (and which can incidentally be set as the default iteration mode in the class itself).

vdumoulin:
Or we raise a warning if indices aren't sorted but do the sorting/unsorting gymnastics anyways.

bartvm:
`batchwise_shuffled_sequential` is definitely something we should implement, because of the performance benefits of being able to read a contiguous chunk of data, but I think we should still have the option of reading actual shuffled examples (as in this PR). When training with large batches on non-shuffled data, I imagine that the two schemes could give slightly different results (because the examples within a batch would be highly correlated you'd get a worse estimate of the true gradient).

vdumoulin:
I re-introduced unordered indexing in H5PYDataset. It will try to index the dataset as usual, and if it fails it will display a warning message once explaining that things might be slower and do the unordered indexing gymnastics.

dwf:
Did you see my comment above on how to do all of it with only one sort?

vdumoulin:
I did, this is what I implemented.

fuel/datasets/hdf5.py
```diff
@@ -316,11 +322,21 @@ def _out_of_memory_get_data(self, state=None, request=None):
             if isinstance(request, slice):
                 request = slice(request.start + subset.start,
                                 request.stop + subset.start, request.step)
+                data = state[source_name][request]
             elif isinstance(request, list):
                 request = [index + subset.start for index in request]
+                try:
+                    data = state[source_name][request]
+                except TypeError:
```
bartvm: I feel we should do one of two things here: Set an internal flag `sort_needed` on the first failure, and don't try again afterwards, or allow a keyword argument to be set that does the same thing.

The problem here is that I'm very worried about the speed of the current solution. Try and except-statements can be quite slow, and are usually discouraged if you know that they are triggered often (as could be the case here). I'm also not sure at what point h5py complains, but if that is halfway through the reading, it could be very, very slow.

vdumoulin: Right. The problem is that at this level in the code `H5PYDataset` can't know for sure whether two consecutive calls of `get_data` originate from the same data stream.

We could tie the `sort_needed` flag to a specific `state` argument, but we would need to compare objects using the `is` operator instead of `==`, as different `h5py.File` instances with the same path/`FileID` compare equal.

vdumoulin: I think your original idea was the better one: sort indices by default in `H5PYDataset` and have the user explicitly set `sort_indices` to `False` if that's the desired behaviour.

dwf: Please document that prominently, both in a comment adjacent to the code
and in relevant class/method docstrings. That's the sort of "whoopsie"
feature that can catch someone off guard.

>

vdumoulin:
After thinking of it, I like @bartvm's original idea better, so this is what I implemented: by default, `ShuffledScheme.sorted_indices` is `False`, and `H5PYDataset` has a `sort_indices` constructor argument (defaulting to `True`) that explicitly controls whether indices are sorted in `get_data`.

vdumoulin:
I pushed a quick fix for #92, since I'm already playing with `H5PYDataset` as part of this PR.

bartvm:
Cool, this looks good to me as soon as the tests pass!

vdumoulin:
Tests are passing.

PR #94, opened 2015-05-01
Make doctests work with Python 2/3 string outputs

### LINKED ISSUE #87: Turn 'how to create a new dataset' tutorial into unit tests

vdumoulin:
This allows to remove a bunch of `doctest: +SKIP` statements and improve overall test coverage.

Fixes #87.

### DIFF
--- a/doctests/__init__.py
+++ b/doctests/__init__.py
@@ -14,7 +14,7 @@ from tests import skip_if_not_available
 
 
 class Py23DocChecker(OutputChecker):
-    """Single-source Python 2/3 output checker
+    """Single-source Python 2/3 output checker.
 
     For more information, see the `original blog post`_.
 
--- a/doctests/__init__.py
+++ b/doctests/__init__.py
@@ -24,9 +24,9 @@ class Py23DocChecker(OutputChecker):
 
     """
     def check_output(self, want, got, optionflags):
-        if sys.version_info[0] > 2:
-            want = re.sub("u'(.*?)'", "'\\1'", want)
-            want = re.sub('u"(.*?)"', '"\\1"', want)
+        if sys.version_info[0] < 3:
+            got = re.sub("u'(.*?)'", "'\\1'", got)
+            got = re.sub('u"(.*?)"', '"\\1"', got)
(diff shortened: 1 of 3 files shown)


bartvm:
I'd prefer it the other way around, actually, because the documentation is supposed to be Python 3. I guess that applying the regular expression to `got` instead of `want` would allow you to make everything work too, without the need to put `'u` in the docstrings everywhere?

doctests/__init__.py
```diff
@@ -5,11 +5,31 @@
...
+    .. _original blog post:
+       https://dirkjan.ochtman.nl/writing/2014/07/06/
+       single-source-python-23-doctests.html
+
+    """
+    def check_output(self, want, got, optionflags):
+        if sys.version_info[0] > 2:
+            want = re.sub("u'(.*?)'", "'\\1'", want)
+            want = re.sub('u"(.*?)"', '"\\1"', want)
+        return doctest.OutputChecker.check_output(self, want, got, optionflags)
```
bartvm: Why not `OutputChecker.check_output` (it's imported after all) or, even better, `super(Py23DocChecker, self).check_output(want, got, optionflags)`

vdumoulin: It's just a copy-paste from the blog post, but sure, I can do that.

vdumoulin: I'll have to settle for `Outpuchecker.check_output`, as `OutputChecker` is an old-style class and subclassing it prevents me from using `super`.

bartvm:
Cool, LGTM.

Too bad that you need to duplicate all the `code-block`s into hidden doctests though. Maybe we can make a ticket for it, and implement a custom `DocTestParser` sometime in the future. It should be possible to basically add a `:doctest:` directive to `code-block`s, that will be picked up by a custom `DocTestParser`.

vdumoulin:
Yeah, it kind of bothers me as well. I think I might have something for that, though, let met give it a try.

vdumoulin:
Nevermind, looks like `doctest` only recognizes interpreter-like commands.

vdumoulin:
Shall I go ahead and merge?

bartvm:
Yeah, go ahead.

PR #95, opened 2015-05-01
MNIST refactoring.

### LINKED ISSUE #84: Refactor MNIST to subclass H5PYDataset
See these two tutorials (in order) for a crash-course on how to implement a new built-in dataset in Fuel:

* [How to use H5PYDataset to load data in Fuel](http://fuel.readthedocs.org/en/latest/h5py_dataset.html)
* [How to implement a built-in dataset in Fuel](http://fuel.readthedocs.org/en/latest/new_dataset.html)

Thrandis:
Fixes #84. @vdumoulin, let me know if you want some changes.

### DIFF
--- a/fuel/converters/mnist.py
+++ b/fuel/converters/mnist.py
@@ -6,6 +6,7 @@ import h5py
 import numpy
 
 from fuel.converters.base import fill_hdf5_file
+
 MNIST_IMAGE_MAGIC = 2051
 MNIST_LABEL_MAGIC = 2049
 
@@ -55,6 +56,7 @@ def mnist(input_directory, save_path, dtype=None):
     h5file['features'].dims[2].label = 'height'
     h5file['features'].dims[3].label = 'width'
     h5file['targets'].dims[0].label = 'batch'
+    h5file['targets'].dims[1].label = 'index'
 
     h5file.flush()
     h5file.close()
@@ -121,7 +123,7 @@ def read_mnist_labels(filename):
 
     Returns
     -------
-    labels : :class:`~numpy.ndarray`, shape (nlabels,)
+    labels : :class:`~numpy.ndarray`, shape (nlabels, 1)
         A one-dimensional unsigned byte array containing the
         labels as integers.
 
@@ -131,4 +133,5 @@ def read_mnist_labels(filename):
         if magic != MNIST_LABEL_MAGIC:
             raise ValueError("Wrong magic number reading MNIST label file")
         array = numpy.frombuffer(f.read(), dtype='uint8')
+    array = array.reshape(array.size, 1)
     return array
--- a/fuel/converters/mnist.py
+++ b/fuel/converters/mnist.py
@@ -49,7 +49,6 @@ def mnist(input_directory, save_path, dtype=None):
             ('train', 'targets', train_labels),
             ('test', 'features', test_features),
             ('test', 'targets', test_labels))
-    print train_features.shape
     fill_hdf5_file(h5file, data)
     h5file['features'].dims[0].label = 'batch'
     h5file['features'].dims[1].label = 'channel'
--- a/fuel/converters/mnist.py
+++ b/fuel/converters/mnist.py
@@ -1,0 +1,135 @@
+import gzip
+import os
+import struct
+
+import h5py
+import numpy
+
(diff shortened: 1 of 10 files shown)


fuel/converters/mnist.py
```diff
@@ -0,0 +1,134 @@
+import gzip
+import os
+import struct
+
+import h5py
+import numpy
+
+from fuel.converters.base import fill_hdf5_file
```
vdumoulin: Could you add a blank line between this line and the next one?

fuel/datasets/mnist.py
```diff
@@ -1,155 +1,41 @@
...
-import numpy
 
 from fuel import config
-from fuel.datasets import IndexableDataset
-from fuel.utils import do_not_pickle_attributes
-MNIST_IMAGE_MAGIC = 2051
-MNIST_LABEL_MAGIC = 2049
+from fuel.datasets import H5PYDataset
+
+logger = logging.getLogger(__name__)
```
vdumoulin: I know a logger is also being created in `binarized_mnist.py` without being used, but could you remove it from this file since it's not being used?

Also, while you're at it, would you mind removing it from `binarized_mnist.py` as well?

fuel/converters/mnist.py
```diff
@@ -0,0 +1,134 @@
...
+    train_features = read_mnist_images(train_feat_path, dtype)
+    train_lab_path = os.path.join(input_directory,
+                                  'train-labels-idx1-ubyte.gz')
+    train_labels = read_mnist_labels(train_lab_path)
+    test_feat_path = os.path.join(input_directory,
+                                  't10k-images-idx3-ubyte.gz')
+    test_features = read_mnist_images(test_feat_path, dtype)
+    test_lab_path = os.path.join(input_directory,
+                                 't10k-labels-idx1-ubyte.gz')
+    test_labels = read_mnist_labels(test_lab_path)
```
vdumoulin: You should reshape this so that it has shape (10000, 1) (and also do the same for train labels). This will fix some of the Travis failures, where a test in Blocks expects targets to be column matrices.

This will also allow you to properly tag the nature of this data by adding the `index` label to `h5file['targets'].dims[1]`.

vdumoulin:
Travis build is failing. There's a test in `test_serialization.py` that you'll need to adapt, because it assumes that MNIST has the old interface.

There's also a doctest failing, because the dtype of MNIST data pre-refactoring was float, which is not the case anymore.

You should override `H5PYDataset.get_data` so that it looks like this:

``` python
    rval = list(super(MNIST, self).get_data(state=state, request=request))
    if 'features' in self.sources:
        i = self.sources.index('features')
        rval[i] = (rval[i] / 255.).astype(config.floatX) 
    return tuple(rval)
```

It's pretty inelegant, but I have something better coming up and for the moment it'll do.

Thrandis:
Hey, thanks for the comments! If I remove the logger, flake8 is not happy:
./datasets/binarized_mnist.py:2:1: F401 'logging' imported but unused
./datasets/mnist.py:2:1: F401 'logging' imported but unused
So, shall I leave them?

bartvm:
Just remove `import logging` as well

Thrandis:
Damn that was obvious... Thanks @bartvm!

Thrandis:
Oh! Seems that I also need to change the file mnist.py in the examples of blocks? Am I right?

vdumoulin:
Yes, see [this](https://github.com/bartvm/fuel/issues/21).

vdumoulin:
I cleared the Travis cache. The HDF5 dataset will be regenerated the next time you push a commit, which should resolve some test failures you were experiencing.

vdumoulin:
There's a failing doctest that you can fix by changing line 76 of `docs/index.rst` to
``` python
>>> mnist = MNIST(which_set='train', flatten=('features',))

vdumoulin:
Also, you can now remove [these lines](https://github.com/Thrandis/fuel/blob/ccw/.travis.yml#L19-L33) from `.travis.yml`.

Thrandis:
forgot the comma after features ...

Thrandis:
@vdumoulin can you launch travis again, I added the missing commas! Thanks :-)

vdumoulin:
Just restarted it.

PR #96, opened 2015-05-03
WIP: Toronto Face Database

jbornschein:
For now is rather a quick "I need TFD 48x48"-hack. But I would extend it to be a polished dataset for the 48x48 and 96x96 TFD.

I'm not sure how to best deal with the 5 different folds defined for this dataset. But I did not look into the 'splits' interface yet. Does it support fancy-indexing instead of simple slices to specify splits?


fuel/converters/toronto_face_database.py
```diff
@@ -0,0 +1,55 @@
...
+
+    """
+    h5file = h5py.File(save_path, mode="w")
+
+    tfd = loadmat(os.path.join(input_directory, 'TFD_48x48.mat'))
+    
+    folds = tfd['folds']
+    features = tfd['images'].reshape([-1, 1, 48, 48])
+    labs_ex = tfd['labs_ex']
+    labs_id = tfd['labs_id']
```
vdumoulin: I'd be more explicit with the choice of source name. Something like `'expression_targets'` and `'identity_targets'` would work for me.

fuel/converters/toronto_face_database.py
```diff
@@ -0,0 +1,55 @@
...
+            ('valid', 'labs_ex' , labs_ex[valid1_mask]),
+            ('valid', 'labs_id' , labs_id[valid1_mask]),
+            ('test', 'features', features[test1_mask]),
+            ('test', 'labs_ex' , labs_ex[test1_mask]),
+            ('test', 'labs_id' , labs_id[test1_mask]),
+    )
+    fill_hdf5_file(h5file, data)
+
+    for i, label in enumerate(('batch', 'channel', 'height', 'width')):
+        h5file['features'].dims[i].label = label
```
vdumoulin: You should also add axis labels for the expression and identity sources.

vdumoulin:
The 'split' interface doesn't support fancy indexing yet, but it shouldn't be overly complicated to implement. I'll have a look at that this week if I have time.

fuel/converters/toronto_face_database.py
jbornschein: I agree -- done.

fuel/converters/toronto_face_database.py
jbornschein: done.

fuel/converters/toronto_face_database.py
```diff
@@ -0,0 +1,73 @@
+import os
+
+import h5py
+
```
vdumoulin: Remove this blank line.

fuel/converters/toronto_face_database.py
```diff
@@ -0,0 +1,73 @@
...
+    Parameters
+    ----------
+    input_directory : str
+        Directory in which the required input files reside.
+    output_fname : str
+        Where to save the converted dataset.
+
+    """
+    with h5py.File(output_fname, mode="w") as h5file:
+
```
vdumoulin: Remove blank line.

fuel/datasets/toronto_face_database.py
```diff
@@ -0,0 +1,29 @@
...
+    which_set : 
+
+    """
+    filename = 'toronto_face_database.hdf5'
+    default_transformers = uint8_pixels_to_floatX(('features',))
+
+    def __init__(self, which_set, **kwargs):
+        kwargs.setdefault('load_in_memory', True)
+        super(TorontoFaceDatabase, self).__init__(
+            self.data_path,
```
vdumoulin: Can you put the args and kwargs on the same line as `self.data_path`?

fuel/datasets/toronto_face_database.py
```diff
@@ -0,0 +1,29 @@
+# -*- coding: utf-8 -*-
+import os
+
+from fuel import config
+from fuel.datasets import H5PYDataset
+from fuel.transformers.defaults import uint8_pixels_to_floatX
+
+
+class TorontoFaceDatabase(H5PYDataset):
+    u"""
```
vdumoulin: Could you add a description like for other datasets (e.g. MNIST, CIFAR10)?

fuel/converters/toronto_face_database.py
```diff
@@ -0,0 +1,73 @@
...
+from scipy.io import loadmat
+
+from fuel.converters.base import fill_hdf5_file, check_exists
+
+ALL_FILES = ['TFD_48x48.mat']
+
+
+@check_exists(required_files=ALL_FILES)
+def convert_toronto_face_database(input_directory, output_fname):
```
vdumoulin: The argument names here should be `directory` and `output_file`.

vdumoulin:
Is the URL for TFD publicly available? If so, we should make a downloader for it. If not, it may still be interesting to have a downloader and have its subparser require the URL as a command-line argument.

vdumoulin:
Could you elaborate on how the `folds` variable in the `TFD_48x48.mat` file is organized?

jbornschein:
No, AFAIK there is no publicly available URL to download it. 

It seems the established way to get access to is to contact the author/collector Joshua M Susskind (http://aclab.human.cornell.edu/people/josh/). I will contact him and ask him what he recommends we should do. 

The folds matrix in `TFD_48x48.mat` is of shape (102236, 5) and of type uint8. For each image there is one row; and it defines 5 different folds (== 5 columns). 
A 0 element indicates the image is unlabeled; a 1 indicates it should be used as a training example, a 2 indicates it should be used as a validation set example and a 3 indicates it should be used as a test set example. In other words: The set of unlabeled examples is the same for all 5 folds, but each fold assigns the labeled examples differently to the train/valid/test sets. I did not implement the different folds (and I'm not yet sure what the best approach would be).

I think there is another open question with this dataset: There is a version with 48x48 images, and a version with 96x96 images. There are obviously multiple options how we could handle this -- I currently think we should choose one these:

1) Use two different dataset classes (`TorontoFaceDataset48` and `TorontoFaceDataset96`), two different hdf5 files and two different conversion scripts.

2) Use two different dataset classes (`TorontoFaceDataset48` and `TorontoFaceDataset96`) but have them both load the data from one common `.hdf5` file which is created by a single conversion script which merges the `TFD_48x48.mat` and `TFD_96x96.mat` files Josh distributes into one.

vdumoulin:
If the author is comfortable with the idea, it would be great to have it publicly available. If not, like I said, I would still make a downloader script but require that the user enters the URL as a command-line argument for the sake of consistency.

We're increasingly trying to stay as close as possible to the original data, so we'd definitely need to keep the two datasets separate.

I'm facing a similar issue for SVHN, where format 1 contains uncropped images and format 2 contains cropped images. I've decided to keep both formats under the same name, but have the format be an argument to the downloader, the converter and the dataset. You can have a look at my PR (#133) to see how I've implemented it.

fuel/datasets/toronto_face_database.py
```diff
@@ -0,0 +1,50 @@
...
+    u"""Toronto Face Database.
+
+    The Toronto Face Dataset [TFD] contains face images from 8 different
+    sources subsumed into a common dataset. To homogenize the images they
+    where all centered and cropped using a face detector.
+
+    In order to use this dataset you need to get the permissions/licenses
+    from all sources:
+
+    ======== ================================================================ ===== ==================================
```
vdumoulin: Just so it fits in the docstring, you could reformat this as

```
* JACFEE  http://www.humintell.com/for-use-in-research/
          ($175, get the Standard Expressor version)
* POFA    http://www.paulekman.com/product/
          pictures-of-facial-affect-pofa/ ($175)
* KDEF    http://www.emotionlab.se/resources/kdef
* CK+     http://www.pitt.edu/~emotion/ck-spread.htm
* MMI     http://www.mmifacedb.com/
* MSFDE   http://psychophysiolab.com/msfde/terms.php
* NimSti  http://www.macbrain.org/resources.htm
* RadBoun http://www.socsci.ru.nl:8180/RaFD2/RaFD?p=register
```

dwf:
I don't think it's a matter of the author's wishes but of license
agreements, sadly.

vdumoulin:
Alright. Let's go with the URL-required-in-downloader option, then.

vdumoulin:
It's a pretty sad thing for TFD that it is so license-ridden. It's a useless barrier to research and innovation.

jbornschein:
Yes -- and I guess it's also annoying for Joshua to distribute his processed files only to persons who can show that they have acquired the  necessary licenses.

vdumoulin:
Yes, it must be.

vdumoulin:
@jbornschein I'm doing a cleanup of the open PRs. Would you still like to go ahead with this one?

jbornschein:
Oh, I totally forgot about this one. 

I'll take a closer look and either finish it or close the PR within the next week.

PR #97, opened 2015-05-04
Automatic tagging of fuel-generated HDF5 files

### LINKED ISSUE #72: HDF5 file format versioning
We should introduce versioning into the HDF5-based format we save, so that we can detect what set of conventions were being followed at that point in time, and read and interpret files that were not written with the current version of the code. 

My personal feeling is that format version need not be tied to (or rely upon) Fuel package versioning. This is the approach adopted by NumPy for NPY, and it has worked well.

vdumoulin:
Fixes #72.

This PR aims at introducing a way to keep trace of how an HDF5 was created using Fuel. It introduces a `interface_version` class attribute in the `H5PYDataset` class which needs to be changed every time the interface changes. It also introduces three attributes tagged in each file `fuel-convert` creates:

* `h5py_interface_version`: self-explanatory
* `fuel_version`: self-explanatory
* `fuel_convert_args`: which command-line arguments were passed to `fuel-convert` in order to create this dataset

### DIFF
--- a/bin/fuel-convert
+++ b/bin/fuel-convert
@@ -36,7 +36,7 @@ if __name__ == "__main__":
     h5file = h5py.File(args.output_file, 'a')
     interface_version = H5PYDataset.interface_version.encode('utf-8')
     h5file.attrs['h5py_interface_version'] = interface_version
-    fuel_convert_version = converters.module_version.encode('utf-8')
+    fuel_convert_version = converters.__version__.encode('utf-8')
     h5file.attrs['fuel_convert_version'] = fuel_convert_version
     command = [os.path.basename(sys.argv[0])] + sys.argv[1:]
     h5file.attrs['fuel_convert_command'] = ' '.join(command).encode('utf-8')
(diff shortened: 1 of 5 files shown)


vdumoulin:
Shortly after creating the PR, I realized that using the Fuel version might be a pain, since it would have to change every time a converter is added or modified. I added a `converters` module version instead.

vdumoulin:
I just rebased. I took that opportunity to add a `fuel-info` script which pretty-prints the metadata contained in a Fuel-converted HDF5 dataset.

dwf:
Why not __version__?

dwf:
I would say use `__version__` but other than that, yes, I think that this is a very good idea (I think I suggested it a while back).

vdumoulin:
Yeah, there's that...

dwf:
To be clear I mean add a `fuel.converters.__version__`, not use `fuel.__version__`.

vdumoulin:
Yay, the tests pass!

PR #98, opened 2015-05-04
CIFAR10 refactoring

ejls:
I've kept the format the same (the images are float32 between 0 and 255, the labels are uint8). Let me know if I missed something.

### DIFF
--- a/fuel/datasets/cifar10.py
+++ b/fuel/datasets/cifar10.py
@@ -1,17 +1,10 @@
 import os
-from collections import OrderedDict
-
-import numpy
-import six
-from six.moves import cPickle
 
 from fuel import config
-from fuel.datasets import IndexableDataset
-from fuel.utils import do_not_pickle_attributes
+from fuel.datasets import H5PYDataset
 
 
-@do_not_pickle_attributes('indexables')
-class CIFAR10(IndexableDataset):
+class CIFAR10(H5PYDataset):
     """The CIFAR10 dataset of natural images.
 
     This dataset is a labeled subset of the ``80 million tiny images''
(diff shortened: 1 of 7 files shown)


vdumoulin:
I would keep the dtype as `uint8` and automatically convert to `float32` and scale between 0 and 1 in `get_data`, like explained [here](https://github.com/bartvm/fuel/pull/95#issuecomment-98373974).

I would also store images in `bc01` instead of as flattened vectors for the sake of clarity: it will allow you to tag axes so that users can easily know which representation is used. You'll have to revert to using the `flatten` argument instead of `reshape`.

Once these two comments are addressed, the Scrutinizer errors are fixed and the Travis build passes, it'll be ready to merge.

vdumoulin:
Oh, one more thing: you can safely remove [these lines](https://github.com/ejls/fuel/blob/master/.travis.yml#L34-L44) now that `CIFAR10` is available through `fuel-download` and you're making use of it in `.travis.yml`.

vdumoulin:
I cleared the Travis cache, once you push a fix for the failing Scrutinizer tests the Travis tests should pass.

fuel/datasets/cifar10.py
```diff
@@ -40,46 +35,29 @@ class CIFAR10(IndexableDataset):
...
+    @property
+    def data_path(self):
+        return os.path.join(config.data_path, self.filename)
+
+    def get_data(self, state=None, request=None):
+        rval = list(super(CIFAR10, self).get_data(state=state,
+                                                  request=request))
+        if 'features' in self.sources:
+            i = self.sources.index('features')
+            rval[i] = rval[i].astype(theano.config.floatX)
```
vdumoulin: Use `config.floatX` here.

vdumoulin:
There's a Scrutinizer error remaining.

fuel/datasets/cifar10.py
```diff
@@ -40,46 +33,29 @@ class CIFAR10(IndexableDataset):
...
-        image_shape = (3072,) if self.flatten else (3, 32, 32)
-        images = numpy.zeros((num_examples,) + image_shape,
-                             dtype=config.floatX)
-        labels = numpy.zeros((num_examples, 1), dtype='uint8')
-        for i, fname in enumerate(self.files[self.which_set]):
-            with open(os.path.join(base_path, fname), 'rb') as f:
-                if six.PY3:
-                    batch = cPickle.load(f, encoding='latin1')
-                else:
-                    batch = cPickle.load(f)
```
vdumoulin: In order to fix the current test failure under Python 3, you should have a look at the above code.

dwf:
Is there any way to get the original uint8 pixels?

vdumoulin:
Not ATM, but I'm working on a PR (#90) that should fix that. Basically, I'd like to introduce an attribute of datasets that's just a sane default transformer to apply. Users would be free to apply that default transformer or use the data as-is if it's what they want.

ejls:
@vdumoulin  thanks for the walkthrough.

I've added a parameter to select the dtype… Maybe I shouldn't have ? Currently, the pixels are NOT scaled in [0,1] since the old code wasn't doing it, should I change this as well?

vdumoulin:
@ejls Here's my intention for this PR:
* Converter code stores images as `uint8` between 0 and 255
* Dataset code (more precisely the `get_data` method) scales batches between 0 and 1 and casts as `fuel.config.floatX`.

I have something coming up that will remove the need for hacky `get_data` overriding and provide more flexibility, but that'll get fixed later.

fuel/datasets/cifar10.py
```diff
@@ -40,46 +33,31 @@ class CIFAR10(IndexableDataset):
...
-    def __init__(self, which_set, flatten=True, **kwargs):
+    def __init__(self, which_set, flatten=True, dtype=config.floatX,
+                 **kwargs):
         if which_set not in ('train', 'test'):
             raise ValueError("CIFAR10 only has a train and test set")
 
+        kwargs.setdefault('load_in_memory', True)
         self.which_set = which_set
-        self.flatten = flatten
+        self.reshape = flatten
```
vdumoulin: Why not use `self.flatten`?

fuel/datasets/cifar10.py
```diff
@@ -40,46 +33,31 @@ class CIFAR10(IndexableDataset):
...
+        self.dtype = dtype
```
vdumoulin: No need to make this an attribute. For the moment it's sufficient that you systematically use `fuel.config.floatX`, so you can just use that directly further down.

ejls:
There is already [these lines](https://github.com/bartvm/fuel/blob/master/fuel/datasets/hdf5.py#L109), so I suppose there is actually no need for flatten, I changed the test so that it uses the `H5PYDataset.__init__` parameters instead.
So it looks like this :
``` 
-    train = CIFAR10('train', load_in_memory=False)
+    train = CIFAR10('train', flatten=('features',),
+                    load_in_memory=False)
-    test = CIFAR10('test', flatten=False, load_in_memory=False)
+    test = CIFAR10('test', load_in_memory=False)
```

@vdumoulin can you clear the cache before I amend the commit? I changed the converter (again).

vdumoulin:
What exactly have you changed in the converter?

ejls:
https://github.com/ejls/fuel/blob/master/fuel/converters/cifar10.py#L39-L42 and https://github.com/ejls/fuel/blob/master/fuel/converters/cifar10.py#L56-L59

vdumoulin:
I just reset the cache.

By the way, it makes reviewing easier if you leave some traces of what you're doing. I get that you want to keep the commit history clean, but this is something that can be done right before merging, when I give my OK.

dwf:
In the case of CIFAR10 (and other colour images datasets) I'd say the sane
default is to cast to floatx and subtract the mean. Mapping to the unit
interval is useful for MNIST but kinda weird for RGB.

fuel/datasets/cifar10.py
```diff
@@ -40,46 +33,31 @@ class CIFAR10(IndexableDataset):
...
-    files = {
-        'train': [os.path.join('cifar-10-batches-py',
-                               'data_batch_{}'.format(i))
-                  for i in range(1, 6)],
-        'test': ['cifar-10-batches-py/test_batch']
-    }
+    filename = 'cifar10.hdf5'
 
-    def __init__(self, which_set, flatten=True, **kwargs):
+    def __init__(self, which_set, flatten=True, dtype=config.floatX,
```
vdumoulin: You should remove the `flatten` kwarg, since `H5PYDataset` already has one (which accepts a tuple of source names; that means you'll have to adapt your unit tests).

vdumoulin:
@dwf At the moment it can't easily be done. Like I said, there's a PR to improve this side of things, so things are bound to change anyways. For this PR I'd be happy that pixel values are mapped to the unit interval, keeping in mind that this is rather temporary.

fuel/datasets/cifar10.py
```diff
@@ -40,46 +33,31 @@ class CIFAR10(IndexableDataset):
...
+    @property
+    def data_path(self):
+        return os.path.join(config.data_path, self.filename)
+
+    def get_data(self, state=None, request=None):
+        rval = list(super(CIFAR10, self).get_data(state=state,
+                                                  request=request))
+        if 'features' in self.sources:
+            i = self.sources.index('features')
+            rval[i] = rval[i].astype(self.dtype)
```
vdumoulin: You should remove `self.dtype` from the constructor and instead use `config.floatX` here. Also, please divide by 255.0 before casting.

The two lines below should be removed, as `H5PYDataset` should take care of the `flatten` constructor argument by itself.

vdumoulin:
@ejls Thank you, looks good to merge.

PR #100, opened 2015-05-06
First attempt to move to path for states.

### LINKED ISSUE #65: H5PYDataset does not play well with DataStream for serialization

Thrandis:
Here is a first draft of my code to fix #65. The role of open() seems a little weird, because even if the file has not be opened, we can still get data using get_file_handle(). Same for close(). What do you think @vdumoulin?

I also moved from dataset.close(handle) to dataset.close(), since we don't need to pass the handle anymore, is it okay?

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -120,7 +120,7 @@ class H5PYDataset(Dataset):
 
     """
     _ref_counts = defaultdict(int)
-    _file_handles = defaultdict()
+    _file_handles = {}
 
     def __init__(self, path, which_set, subset=None, load_in_memory=False,
                  flatten=None, driver=None, sort_indices=True, **kwargs):
@@ -227,19 +227,21 @@ class H5PYDataset(Dataset):
     @property
     def split_dict(self):
         if not hasattr(self, '_split_dict'):
-            handle = self.get_file_handle()
+            self._out_of_memory_open()
+            handle = self._file_handle
             split_array = handle.attrs['split']
             self._split_dict = H5PYDataset.parse_split_array(split_array)
-            self.release_file_handle()
+            self._out_of_memory_close()
         return self._split_dict
 
     def load_axis_labels(self):
-        handle = self.get_file_handle()
+        self._out_of_memory_open()
+        handle = self._file_handle
         axis_labels = {}
         for source_name in handle:
             axis_labels[source_name] = tuple(
                 dim.label for dim in handle[source_name].dims)
-        self.release_file_handle()
+        self._out_of_memory_close()
         return axis_labels
 
     @property
@@ -270,23 +272,9 @@ class H5PYDataset(Dataset):
             self._subsets = subsets
         return self._subsets
 
-    def get_file_handle(self):
-        if self.path not in self._file_handles.keys():
-            handle = h5py.File(name=self.path, mode="r", driver=self.driver)
-            self._file_handles[self.path] = handle
-        self._ref_counts[self.path] += 1
-        return self._file_handles[self.path]
-
-    def release_file_handle(self):
-        self._ref_counts[self.path] -= 1
-        if not self._ref_counts[self.path]:
-            del self._ref_counts[self.path]
-            self._file_handles[self.path].close()
-            del self._file_handles[self.path]
-
     def load(self):
         if self.load_in_memory:
-            handle = self.get_file_handle()
+            handle = self._file_handle
             data_sources = []
             for source_name, subset in zip(self.sources, self.subsets):
                 data = handle[source_name][subset]
@@ -294,7 +282,6 @@ class H5PYDataset(Dataset):
                     data = data.reshape((data.shape[0], -1))
                 data_sources.append(data)
             self.data_sources = data_sources
-            self.release_file_handle()
         else:
             self.data_sources = None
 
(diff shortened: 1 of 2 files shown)


fuel/datasets/hdf5.py
```diff
@@ -119,7 +119,8 @@ class H5PYDataset(Dataset):
         indices are ordered.
 
     """
-    ref_counts = defaultdict(int)
+    _ref_counts = defaultdict(int)
+    _file_handles = defaultdict()
```
vdumoulin: No need to make this a `defaultdict`. The reason why `_ref_counts` is a `defaultdict` is to simplify the logic in the defunct `_get_file_id` method, but for `_file_handles` we gain nothing.

fuel/datasets/hdf5.py
```diff
@@ -300,20 +306,12 @@ def open(self):
...
     def _out_of_memory_open(self):
-        file_id = self._get_file_id()
-        state = h5py.File(name=file_id, mode="r", driver=self.driver)
-        self.ref_counts[state.id] += 1
-        return state
+        self.get_file_handle()
+        return self.path
 
-    def close(self, state):
+    def close(self):
```
vdumoulin: This method is part of the interface defined in `Dataset` (see its definition [here](https://github.com/bartvm/fuel/blob/master/fuel/datasets/base.py#L130-L132)), and as such we should keep the `state` argument.

vdumoulin:
You do have a point: passing the file path as the handle is pretty reduntant, and we could just as well pass `None`.

Here's an alternative:
* We put the code from `get_file_handle` back in `_out_of_memory_open` (except that it doesn't return anything) and we remove the `get_file_handle` method.
* The `open` method calls `self._out_of_memory_open()` if `self.load_in_memory` is `False` and returns `None` no matter what.
* We introduce the following property:
``` python
@property
def _file_handle(self):
    if not self.path in self._file_handles:
        raise IOError('no open handle for file {}'.format(self.path))
    return self._file_handles[self.path]
```
* File handles are accessed like so:
``` python
handle = self._file_handle
```
* We put the implementation of `release_file_handle` into `out_of_memory_close` and remove the `release_file_handle` method.

vdumoulin:
It would also be good to add a unit test that fails currently but passes with your changes.

Try adding a test like
``` python
# ...
dataset = H5PYDataset(...)
stream = DataStream(dataset)
cPickle.dumps(stream)
```

Thrandis:
@vdumoulin: The problem now is that when we instantiate a new H5PYDataset, the function split_dict() is called (through available_splits()), but the dataset is not opened yet.
Is it okay to call open() at this moment (or _out_of_memory_open()), and immediately close it afterwards?

vdumoulin:
Yes, that would be acceptable: you're opening the file to read the metadata and then closing it.

Thrandis:
I still have some tests that fails. It seems that I have trouble getting the data or when there is two different H5PYDataset on the same hdf5 file.

vdumoulin:
Which tests, exactly?

fuel/datasets/hdf5.py
```diff
@@ -275,12 +270,11 @@ def subsets(self):
                 if num_examples != subset.stop - subset.start:
                     raise ValueError("sources have different lengths")
             self._subsets = subsets
-            self._out_of_memory_close(handle)
         return self._subsets
 
     def load(self):
         if self.load_in_memory:
-            handle = self._out_of_memory_open()
```
vdumoulin: In this specific case you should open and close: we're reading the file from disk and into memory. This is probably why your tests are failing.

Thrandis:
test_hdf5.test_h5py_dataset_split
  -> ValueError: 'test' split is not provided by this dataset. Available splits are (u'train',).
test_hdf5.test_h5py_dataset_out_of_memory
 -> ValueError: 'test' split is not provided by this dataset. Available splits are (u'train',).
test_hdf5.test_h5py_flatten_in_memory
 -> AssertionError: Arrays are not equal
test_hdf5.test_h5py_flatten_out_of_memory
 -> AssertionError: Arrays are not equal

vdumoulin:
Weird. Could you fix what I've outlined just to see if it solves the issue?

Thrandis:
It don't solve the issuse ... :-(

vdumoulin:
Calling `self._out_of_memory_open` and `self._out_of_memory_close` in `load` seems to solve at least part of the failures for me.

Thrandis:
Mhhh strange. For me the 4 tests don't pass using `self._out_of_memory_open` and `self._out_of_memory_close`.

vdumoulin:
You need to open and close in `load`.

I think I found why the tests fail: in `test_dataset_datastream_pickles`, you're not closing the data stream, which leaves a dangling file handle in the `_file_handles` dict. Since we're re-using `'tmp.hdf5'` everywhere in the tests, it causes some issues.

The solution is to make sure that `close` gets called no matter what, by putting it in the `finally` block. (By the way, it would be a good thing to do that for all tests in that file as well.)

Thrandis:
Great! I'll change that!

vdumoulin:
Looks good, thanks a lot!

Thrandis:
@vdumoulin thanks for your help ;-)

PR #103, opened 2015-05-10
Test coverage improvements

vdumoulin:
Test coverage was dropping pretty fast, so I figured I'd do something about it.

It's still work in progress, but it may be a good idea to merge it in its current state (this PR is already getting quite big to review, sorry about that!) and continue in another PR.

### DIFF
--- a/bin/fuel-convert
+++ b/bin/fuel-convert
@@ -23,4 +23,6 @@ if __name__ == "__main__":
             default=os.path.join(os.getcwd(), '{}.hdf5'.format(name)))
         subparser_fn(subparser)
     args = parser.parse_args()
-    args.func(args)
+    args_dict = vars(args)
+    func = args_dict.pop('func')
+    func(**args_dict)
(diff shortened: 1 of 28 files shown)


vdumoulin:
General tip to whoever will review this PR: it may be faster to review it commit by commit, as I didn't do lots of back-and-forth edits between commits and they're incremental.

bartvm:
Looks good to me at first sight, but I have to say that towards the end of the PR I found myself skimming more than reading... I'll try and give it a second pass later today.

vdumoulin:
Thanks a lot for having a look! @dwf also said he'd give it a look.

dwf:
So I think my comments boil down to two major things:
* Tempfile module where appropriate 
* Use
```python
args_dict = vars(args)
args_dict.pop('func')
args.func(**args_dict)
```

and 

```python
def my_converter(foo, bar, baz):
    ...
```
rather than
```python
args.func(args)
```
and
```python
def my_converter(args):
    foo = args.foo
    bar = args.bar
    baz = args.baz
```

vdumoulin:
@dwf I addressed your comments in the four most recent commits. Let's wait and see what Travis has to say.

dwf:
LGTM. I don't see any reason not to merge.

Issue #86, opened 2015-04-27
Add CIFAR100 to built-in datasets


vdumoulin:
See these two tutorials (in order) for a crash-course on how to implement a new built-in dataset in Fuel:

* [How to use H5PYDataset to load data in Fuel](http://fuel.readthedocs.org/en/latest/h5py_dataset.html)
* [How to implement a built-in dataset in Fuel](http://fuel.readthedocs.org/en/latest/new_dataset.html)

vdumoulin:
@julianser This is your ticket.

vdumoulin:
Fixed by #124.

PR #105, opened 2015-05-11
Added CIFAR100 to built-in datasets, issue #86.

julianser:
@vdumoulin

### DIFF
--- a/fuel/downloaders/cifar100.py
+++ b/fuel/downloaders/cifar100.py
@@ -1,0 +1,22 @@
+from fuel.downloaders.base import default_downloader
+
+
+def cifar100(subparser):
+    """Sets up a subparser to download the CIFAR-100 dataset file.
+
+    The CIFAR-109 dataset file is downloaded from Alex Krizhevsky's
+    website [ALEX].
+
+    .. [ALEX] http://www.cs.toronto.edu/~kriz/cifar-100-python.tar.gz
+
+    Parameters
+    ----------
+    subparser : :class:`argparse.ArgumentParser`
+        Subparser handling the `cifar10` command.
+
+    """
+    url = 'http://www.cs.toronto.edu/~kriz/cifar-100-python.tar.gz'
+    filename = 'cifar-100-python.tar.gz'
+    subparser.set_defaults(
+        func=default_downloader, urls=[url], filenames=[filename])
+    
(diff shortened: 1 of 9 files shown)


fuel/downloaders/cifar100.py
```diff
@@ -0,0 +1,22 @@
+from fuel.downloaders.base import default_downloader
+
+
+def cifar100(subparser):
+    """Sets up a subparser to download the CIFAR-100 dataset file.
+
+    The CIFAR-109 dataset file is downloaded from Alex Krizhevsky's
```
vdumoulin: Typo: `CIFAR-109`

fuel/downloaders/cifar100.py
```diff
@@ -0,0 +1,22 @@
...
+    website [ALEX].
+
+    .. [ALEX] http://www.cs.toronto.edu/~kriz/cifar-100-python.tar.gz
+
+    Parameters
+    ----------
+    subparser : :class:`argparse.ArgumentParser`
+        Subparser handling the `cifar10` command.
```
vdumoulin: Typo: `cifar10`

vdumoulin:
Can you also add unit tests like [here](https://github.com/bartvm/fuel/blob/master/tests/test_cifar10.py)?

vdumoulin:
If my comments are addressed and the Scrutinizer errors are fixed, we'll be good to merge (Travis is under maintenance, I'll manually restart the tests if need be).

vdumoulin:
The current test failure is caused by the dataset not being present in the Travis build bot. You can solve it by adding a block of code similar to [this code block](https://github.com/bartvm/fuel/blob/master/.travis.yml#L44-L52) right below in `.travis.yml`.

Also, in order to prevent the tests from failing if the dataset isn't present, you can add a `skip_if_not_available` statement at the beginning, like [here](https://github.com/bartvm/fuel/blob/master/tests/test_mnist.py#L15).

vdumoulin:
There are still [PEP8 errors](https://scrutinizer-ci.com/g/bartvm/fuel/inspections/90643fdc-8e20-4ff7-96b8-fc9c6c01720f) to fix, but the Travis failures look unrelated to your PR. I would merge your PR as soon as you fix the PEP8 errors and I'll deal with the unrelated test failures afterwards if need be.

vdumoulin:
There's been a slight change in how converters work and how converters and downloaders register to their respective module:

http://fuel.readthedocs.org/en/latest/new_dataset.html#download-code

http://fuel.readthedocs.org/en/latest/new_dataset.html#conversion-code

julianser:
I've made a new (cleaner) PR.

dwf:
FYI there's never really a need to create a new PR, you can just push to
the same remote branch with --force.

PR #110, opened 2015-05-14
test unpack picklable

mducoffe:


mducoffe:
I have tested with NUMPY_AVAILABLE = True/False in iter_dispatch from picklable itertools. When NUMPY_AVAILABLE is False, the test fails

mducoffe:
@vdumoulin first impression ?

tests/test_serialization.py
```diff
@@ -40,3 +41,14 @@ def test_in_memory():
...
+
+def test_unpack_picklable():
+    data = numpy.arange(10)
+    stream = Batch(
+        DataStream(IterableDataset(data)), iteration_scheme=ConstantScheme(2))
+    wrapper = Unpack(stream)
+    epoch = wrapper.get_epoch_iterator()
+    # Pickle the epoch
+    with tempfile.NamedTemporaryFile(delete=False) as f:
+        cPickle.dump(epoch, f)
```
vdumoulin: You can remove the need for a temporary file by using `cPickle.dumps`, which returns the pickled string for a given object:

``` python
cPickle.dumps(epoch)
```

would be sufficient.

vdumoulin:
Beside my comment everything looks fine.

Your location choice is not a bad idea at all, but I'd put this test in `tests/test_transformers.py` along with the other Unpack unit test. You can reformat the two tests as a test class and factor out the common logic in a `setUp` method (see `TestMerge` for a good example on how to do that).

vdumoulin:
Mission accomplished. Merging.

PR #112, opened 2015-05-14
Added RenameStream transformer.

### LINKED ISSUE #12: RenameStream
From https://github.com/bartvm/blocks/issues/267

Something that @arasmus ran into: When combining multiple data streams, the names of the sources might clash. I guess a general solution for this would be to create a `RenameStream` which just renames the sources and passes on the data directly. That way you can put than into the yet-to-be-written data stream-chainer.

Thrandis:
Fixes #12

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -4,7 +4,7 @@ from multiprocessing import Process, Queue
 import numpy
 from picklable_itertools import chain, ifilter, izip, imap, repeat, starmap
 from picklable_itertools.extras import partition
-from six import add_metaclass
+from six import add_metaclass, iteritems
 
 from fuel import config
 from fuel.streams import AbstractDataStream
@@ -592,21 +592,22 @@ class MultiProcessing(Transformer):
         return data
 
 
-class RenameStream(Transformer):
+class Rename(Transformer):
     """Renames the sources of the stream.
 
     Parameters
     ----------
-    data_stream : :class:`DataStream` or :class:`Transformer`
+    data_stream : :class:`DataStream` or :class:`Transformer`.
         The data stream.
     names : dict
-        A dictionnary mapping the old and new names of the sources to rename.
+        A dictionnary mapping the old and new names of the sources
+        to rename.
 
     """
     def __init__(self, data_stream, names):
-        super(RenameStream, self).__init__(data_stream)
+        super(Rename, self).__init__(data_stream)
         sources = list(self.data_stream.sources)
-        for old, new in names.iteritems():
+        for old, new in iteritems(names):
             if old not in sources:
                 raise KeyError("%s not in the sources of the stream" % old)
             else:
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -590,3 +590,31 @@ class MultiProcessing(Transformer):
         if data == StopIteration:
             raise StopIteration
         return data
+
+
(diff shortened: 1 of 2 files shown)


fuel/transformers/__init__.py
```diff
@@ -590,3 +590,31 @@ def get_data(self, request=None):
         if data == StopIteration:
             raise StopIteration
         return data
+
+
+class RenameStream(Transformer):
```
vdumoulin: You did exactly as the ticket suggested, but I'd change the name to simply `Rename`. The `Stream` suffix looks redundant.

fuel/transformers/__init__.py
```diff
@@ -590,3 +590,31 @@ def get_data(self, request=None):
...
+    """Renames the sources of the stream.
+
+    Parameters
+    ----------
+    data_stream : :class:`DataStream` or :class:`Transformer`
+        The data stream.
+    names : dict
+        A dictionnary mapping the old and new names of the sources to rename.
```
vdumoulin: This is too long, according to PEP8. Break that in two lines of < 75 characters, it'll fix the Scrutinizer failure.

fuel/transformers/__init__.py
```diff
@@ -590,3 +590,31 @@ def get_data(self, request=None):
...
+
+    """
+    def __init__(self, data_stream, names):
+        super(RenameStream, self).__init__(data_stream)
+        sources = list(self.data_stream.sources)
+        for old, new in names.iteritems():
```
vdumoulin: Travis fails because `dict.iteritems` exists only in Python 2.x. Use [six.iteritems](https://pythonhosted.org/six/#six.iteritems) for 2/3 cross-compatibility.

fuel/transformers/__init__.py
```diff
@@ -590,3 +590,31 @@ def get_data(self, request=None):
...
+            if old not in sources:
+                raise KeyError("%s not in the sources of the stream" % old)
+            else:
+                sources[sources.index(old)] = new
+        self.sources = tuple(sources)
+
+    def get_data(self, request=None):
```
vdumoulin: Right now this method isn't covered by any unit test. Here's a suggestion: what's particular about this implementation of `get_data` is that it should not alter what its child iterator returns in any way. You could test for that.

vdumoulin:
Great job! Once my comments are addressed it'll be ready to merge.

vdumoulin:
I restarted the stalled Travis build. Could you add one last check in your unit test? `get_data` raises a `ValueError` if `request` is not `None`, but right now this isn't covered by any check.

Thrandis:
There you go! Thanks for your patience :-)

vdumoulin:
No problem at all, thank you very much for your commitment! :)

By the way, I was able to pinpoint where to add tests by looking at the Coveralls report. It's really a great tool to identify which cases aren't covered by actual unit tests. You can also run it locally this way:

``` bash
coverage run -p --source=fuel -m nose2.__main__ -v tests
coverage run -p --source=fuel -m nose2.__main__ -v doctests
coverage combine
coverage html
```

This will produce an `htmlcov` directory containing a coverage report. Just point to `htmlcov/index.html` in your web browser and you'll be able to see a module-by-module breakdown of test coverage. You'll also be able to click on a particular module and see what's covered and what's not.

Merging.

Thrandis:
Oh that's great! I'll use that next time!

PR #113, opened 2015-05-15
Gracefully handle parser error.

dwf:
Right now, fuel-convert with no argument spits out something kinda ugly.


dwf:
@vdumoulin

vdumoulin:
Weird, on my end it fails gracefully without having to add this block of code.

vdumoulin:
Nevermind, some old `fuel-convert` script was masking the issue. LGTM.

PR #114, opened 2015-05-15
Better converter errors on non-existent files.

dwf:
@vdumoulin

### DIFF
--- a/bin/fuel-convert
+++ b/bin/fuel-convert
@@ -6,6 +6,7 @@ import sys
 import h5py
 
 from fuel import converters
+from fuel.converters.base import MissingInputFiles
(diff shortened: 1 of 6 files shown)


dwf:
Fixed + squashed the typo from before.

vdumoulin:
I'm hesitant on being too strict, but could you make sure that what you introduced is covered by a unit test?

dwf:
@vdumoulin 95e1511

dwf:
Woopsie, need `from __future__ import print_function`.  cd1d750

dwf:
On second thought, squashed as 96302f7.

vdumoulin:
vdumoulin's seal of approval

PR #116, opened 2015-05-16
Downloaders refactor & non-public URL support

dwf:
* Refactor downloaders in the same way that converters was recently refactored.
* Allow specification of url_prefix to default_downloader.
* Incidentally also fix a similar error to #114's target.


vdumoulin:
vdumoulin's seal of approval

PR #118, opened 2015-05-16
Improve download function

### LINKED ISSUE #117: More robust download function
Some datasets are hundreds of gigabytes. Currently the default downloader provides no notion of progress, let alone support for resumption.

Surely there's a library out there that we can offload all of this to...

vdumoulin:
This PR adds a progress bar. Partially fixes #117.

### DIFF
--- a/fuel/downloaders/base.py
+++ b/fuel/downloaders/base.py
@@ -101,7 +101,7 @@ def default_downloader(directory, urls, filenames, url_prefix=None,
             if os.path.isfile(f):
                 os.remove(f)
     else:
-        print('Downloading', ', '.join(filenames), '\n')
+        print('Downloading ' + ', '.join(filenames) + '\n')
         for url, f, n in zip(urls, files, filenames):
             if not url:
                 if url_prefix is None:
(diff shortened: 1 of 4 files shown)


vdumoulin:
@dwf

fuel/downloaders/base.py
```diff
@@ -11,6 +12,22 @@ class NeedURLPrefix(Exception):
     pass
 
 
+@contextmanager
+def progress_bar(name, maxval):
```
dwf: nice use of context managers

vdumoulin:
I'm debating adding an additional dependency (`httmock`) to mock HTTP requests. It allows to speed up the tests as well as craft responses that target specific cases in download helpers. Any thoughts?

dwf:
Probably a good call (running tests offline is kind of important for working on the road, and the UCI people probably do not appreciate our Travis bot hitting their server frequently), but save that for another PR.

bartvm:
For tests you can add whatever you want I think, because the user won't
have to install them. Just make sure to add them to req-travis-pip.txt
after https://github.com/mila-udem/fuel/pull/120 got merged

bartvm:
Oh, and add them as an optional dependency in `setup.py`

dwf:
IIRC there's some sort of "tests_require" thing in setup.py, no?

bartvm:
https://github.com/mila-udem/blocks/blob/master/setup.py#L40-L44

PR #120, opened 2015-05-16
Separate Fuel and Blocks tests

bartvm:


vdumoulin:
For the sake of curiosity, other than being better organized, does this change anything?

bartvm:
Nope, but I'm trying to clean things up a bit, because we're going to have multiple repos (`blocks-examples`, etc.) and in the ideal case (need to see if that's feasible) each repo runs the tests of all the other repos that depend on it, which means Fuel would have to run Fuel, Blocks, Blocks Extras and Blocks Examples... That's a lot of tests, and I'd prefer not to have endless `.travis.yml` files in each repo, so I'm trying to do some cleaning up, although I still haven't found the ideal way of doing this.

vdumoulin:
I see. Thanks for the explanation.

bartvm:
Rebased and squashed. If tests pass, could someone have a look and merge?

vdumoulin:
LGTM. I'll merge if the tests pass.

bartvm:
They already did :) (the last commit didn't add any code, just a change to the README).

vdumoulin:
Great, merging!

PR #122, opened 2015-05-17
Mock HTTP responses in downloader tests

vdumoulin:
With this the Fuel test suite should be able to run offline as well.

Plus it speeds up test execution, which is always a good thing. :)

### DIFF
--- a/docs/new_dataset.rst
+++ b/docs/new_dataset.rst
@@ -242,7 +242,6 @@ You can now use the Iris dataset like you would use any other built-in dataset:
     :hide:
     >>> import mock
     >>> import os
-    >>> import sys
     >>> from picklable_itertools import chain
     >>> from six.moves import range
     >>> from fuel.downloaders.base import default_downloader
@@ -260,7 +259,6 @@ You can now use the Iris dataset like you would use any other built-in dataset:
     >>> fill_downloader_subparser(subparsers.add_parser('iris'))
     >>> args = parser.parse_args(['iris'])
     >>> args_dict = vars(args)
-    >>> args_dict['fd'] = sys.stdout
     >>> func = args_dict.pop('func')
     >>> content = b''
     >>> for i in range(50):
--- a/docs/new_dataset.rst
+++ b/docs/new_dataset.rst
@@ -240,18 +240,11 @@ You can now use the Iris dataset like you would use any other built-in dataset:
 
 .. doctest::
     :hide:
-    >>> from httmock import all_requests, HTTMock
-    >>> content = b''
-    >>> for i in range(50):
-    ...    content += b'0.0,0.0,0.0,0.0,Iris-setosa\n'
-    >>> for i in range(50):
-    ...    content += b'0.0,0.0,0.0,0.0,Iris-versicolor\n'
-    >>> for i in range(50):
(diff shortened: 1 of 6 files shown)


vdumoulin:
@dwf following up on my previous PR.

vdumoulin:
I'm still in the process of figuring the best way to mock out HTTP responses, don't merge this PR yet.

vdumoulin:
@dwf I'm ready for merciless scrutiny.

fuel/downloaders/base.py
```diff
@@ -55,6 +69,9 @@ def download(url, file_handle, chunk_size=1024):
         URL to download.
     file_handle : file
         Where to save the downloaded URL.
+    fd : open file handle, optional
```
dwf: fd is a bit of a confusing name.  I'm also not sure about the choice of stderr. Seems like that be reserved for, well, errors, warnings, that sort of thing... @bartvm?

vdumoulin: The reason for that is I re-used the same name as in `progressbar2`, as well as the same default value. I have nothing against giving it a better name and a saner default. Ideas?

dwf: The reason it's confusing to me is that it's usually used for file handles,
but there's already a file handle argument. Maybe that's enough to tip off
someone to read the docstring, I dunno.

vdumoulin: Maybe something like `progressbar_fd` then?

vdumoulin: After some thought, the only reason I have this kwarg is so that I don't litter the test logs with progress bar outputs (nose buffers stdout and nose2 can do the same with the -B flag). I just changed the `fd` argument to `ProgressBar` to be `sys.stdout` and we'll use that as default.

vdumoulin:
Ping @dwf.

docs/new_dataset.rst
```diff
@@ -257,7 +260,25 @@ You can now use the Iris dataset like you would use any other built-in dataset:
...
+    ... def call_download(func, args_dict, mock_requests):
+    ...     mock_response = mock.Mock()
+    ...     mock_response.iter_content = mock.Mock(
+    ...         side_effect = lambda s: chain(
+    ...             (content[s * i: s * (i + 1)] for i in range(length // s)),
+    ...             (content[(length // s) * s:],)))
+    ...     mock_response.headers = {'content-length': length}
+    ...     mock_requests.get.return_value = mock_response
+    ...     func(**args_dict)
+    >>> call_download(func, args_dict) # doctest: +ELLIPSIS
```
dwf: Hmm, so, the point of doctests is to be documentation as well as tests. I'm not sure the documentation is actually meaningful at this point... I'd suggest refactoring it into some sort of utility code that minimizes the amount of extra mock-related stuff necessary for the reader to read.

dwf: And it might be worth just not mocking it in the doctests, I dunno, see what @bartvm thinks.

dwf:
Otherwise, LGTM

docs/new_dataset.rst
dwf: Just saw the :hide:. Alrighty, merging.

PR #124, opened 2015-05-19
New commit to fix PEP8 formatting in issue 6 and to merge with new ch…

julianser:
…anges in Fuel. @vdumoulin

### DIFF
--- a/fuel/converters/cifar100.py
+++ b/fuel/converters/cifar100.py
@@ -91,3 +91,4 @@ def fill_subparser(subparser):
         Subparser handling the `cifar100` command.
     """
     subparser.set_defaults(func=convert_cifar100)
+
--- a/fuel/converters/cifar100.py
+++ b/fuel/converters/cifar100.py
@@ -4,7 +4,7 @@ import tarfile
 import h5py
 import numpy
 import six
-from six.moves import range, cPickle
+from six.moves import cPickle
 
 from fuel.converters.base import fill_hdf5_file, check_exists
 
@@ -84,6 +84,7 @@ def convert_cifar100(directory, output_file):
 
 def fill_subparser(subparser):
     """Sets up a subparser to convert the CIFAR100 dataset files.
+
     Parameters
     ----------
     subparser : :class:`argparse.ArgumentParser`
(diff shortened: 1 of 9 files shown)


dwf:
https://scrutinizer-ci.com/g/mila-udem/fuel/inspections/16765e0a-d74d-4630-9c77-ce182423c4ac

fuel/converters/cifar100.py
```diff
@@ -0,0 +1,93 @@
...
+    h5file.close()
+
+
+def fill_subparser(subparser):
+    """Sets up a subparser to convert the CIFAR100 dataset files.
+
+    Parameters
+    ----------
+    subparser : :class:`argparse.ArgumentParser`
+        Subparser handling the `cifar100` command.
```
vdumoulin: You're missing a blank line after this line, this is why Scrutinizer complains.

vdumoulin:
Currently the tests fail because the CIFAR100 dataset isn't downloaded when Travis builds.

Add `cifar100` to [this line](https://github.com/mila-udem/fuel/blob/master/.travis.yml#L40), and we should be okay.

julianser:
@vdumoulin I think there is a merge mistake. I am looking into it...

vdumoulin:
I think it may just be a punctual failure. I restarted the tests, we'll see what happens.

julianser:
Yes, it must be, because I just ran:

fuel-download cifar100
fuel-convert cifar100

They both worked fine...

vdumoulin:
The tests now pass. Go figure...

The only thing left to do now is to make sure we maintain test coverage. There are two things to test: the download script and the conversion script.

This involves creating tests like the [cifar10 downloader test](https://github.com/mila-udem/fuel/blob/master/tests/test_downloaders.py#L100-L109) and the [cifar10 converter test](https://github.com/mila-udem/fuel/blob/master/tests/test_converters.py#L220-L279).

You can pretty much copy-paste those tests and adapt them for CIFAR100.

vdumoulin:
Looking back at the converter test, it's more involved than adapting the CIFAR10 test. Let's leave the coverage as it is and make the converter test as part of another PR.

Once the tests pass, I'll merge.

PR #126, opened 2015-05-19
Add ConcatenatedScheme.

dwf:


vdumoulin:
LGTM. Merge at will when the tests pass.

dwf:
Coveralls isn't working but it's small enough that you can tell by inspection that all new lines are covered by tests, and the tests passed on Travis.

PR #133, opened 2015-05-22
Implement SVHN dataset (format 2)

vdumoulin:
Here's an implementation of the SVHN dataset, format 2. Some of the infrastructure for implementing format 1 has also been put in place.

### DIFF
--- a/fuel/converters/svhn.py
+++ b/fuel/converters/svhn.py
@@ -210,16 +210,16 @@ def convert_svhn_format_1(directory, output_file):
         split_dict = {
             'train': {
                 'features': train_interval, 'targets': train_interval,
-                'bbox_height': train_interval, 'bbox_width': train_interval,
-                'bbox_left': train_interval, 'bbox_top': train_interval},
+                'bbox_heights': train_interval, 'bbox_widths': train_interval,
+                'bbox_lefts': train_interval, 'bbox_tops': train_interval},
             'test': {
                 'features': test_interval, 'targets': test_interval,
-                'bbox_height': test_interval, 'bbox_width': test_interval,
-                'bbox_left': test_interval, 'bbox_top': test_interval},
+                'bbox_heights': test_interval, 'bbox_widths': test_interval,
+                'bbox_lefts': test_interval, 'bbox_tops': test_interval},
             'extra': {
                 'features': extra_interval, 'targets': extra_interval,
-                'bbox_height': extra_interval, 'bbox_width': extra_interval,
-                'bbox_left': extra_interval, 'bbox_top': extra_interval}}
+                'bbox_heights': extra_interval, 'bbox_widths': extra_interval,
+                'bbox_lefts': extra_interval, 'bbox_tops': extra_interval}}
         h5file.attrs['split'] = H5PYDataset.create_split_array(split_dict)
     finally:
         h5file.flush()
--- a/fuel/converters/svhn.py
+++ b/fuel/converters/svhn.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 import os
 
 import h5py
--- a/fuel/converters/svhn.py
+++ b/fuel/converters/svhn.py
@@ -1,0 +1,131 @@
+import os
+
+import h5py
+from scipy.io import loadmat
+
+from fuel.converters.base import fill_hdf5_file, check_exists
+
+
+FORMAT_1_FILES = ['{}.tar.gz'.format(s) for s in ['train', 'test', 'extra']]
+FORMAT_1_TRAIN_FILE, FORMAT_1_TEST_FILE, FORMAT_1_EXTRA_FILE = FORMAT_1_FILES
+FORMAT_2_FILES = ['{}_32x32.mat'.format(s) for s in ['train', 'test', 'extra']]
+FORMAT_2_TRAIN_FILE, FORMAT_2_TEST_FILE, FORMAT_2_EXTRA_FILE = FORMAT_2_FILES
+
+
+@check_exists(required_files=FORMAT_1_FILES)
+def convert_svhn_format_1(directory, output_file):
+    """Converts the SVHN dataset (format 1) to HDF5.
+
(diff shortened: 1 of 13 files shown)


vdumoulin:
Before implementing the format 1 converter, I'd like to figure out the following: how should we deal with variable-length examples in HDF5 files?

In the past, I've solved that problem by storing a dataset of variable-sized images as a ragged array of variable-length vectors. In order to unflatten the images, I also stored an array with their shapes.

We could implement a similar solution in Fuel and automagically reshape the flat images through the `Dataset.default_transformers` class attribute.

Comments? Suggestions?

bartvm:
I think the approach of storing vectors and shapes separately is a good one, but I'm wondering whether we should implement that at a lower level than transformers, simply because the data is nonsensical without the shape information. We could implement it as part of `H5PYDataset`, and perform the reshaping in the `get_data` method?

vdumoulin:
Would you be comfortable merging this PR right away (after reviewing, of course) and have format 1 be implemented in another PR?

vdumoulin:
I went ahead and implemented the converter for SVHN format 1. A couple remarks:

* Although the converter works, you won't be able to open this dataset with `H5PYDataset` until #141 is merged. However, converting using this branch and opening under the `vlen_h5pydataset` branch works; I've tried it successfully.
* The `digitStruct.mat` file contained in the archives is a version 7.3 file, and it is a pain to work with. In theory, it's HDF5 under the hood, but in practice it's obfuscated with region selections of region selections of region selections. It's also pretty slow: parsing this file for the `extra` split takes around 20-25 minutes on my machine.
* Overall, converting SVHN format 1 takes around 55-60 minutes on my machine. It's far from blazing fast, but it's not painfully slow either.
* Memory-wise, the converted dataset takes 5.9G.

vdumoulin:
@bartvm Now that variable-length data is supported in `H5PYDataset`, would you mind having a look at this PR? The _format 1_ version of SVHN would be a useful reference implementation of a built-in dataset with variable-length data.

fuel/converters/svhn.py
```diff
@@ -0,0 +1,298 @@
...
+                        f.extract(member, path=TMPDIR)
+                        bar.update(i)
+            return num_examples
+
+        examples_per_split = {}
+        for split in splits:
+            examples_per_split[split] = extract_tar(split)
+
+        num_examples = sum(examples_per_split.values())
+        split_intervals = {}
```
bartvm: Maybe something like
```python
examples_per_split = [extract_tar(split) for split in splits]
intervals = zip([0] + examples_per_split, examples_per_split + [num_examples])
split_intervals = dict(zip(splits, intervals))
```
would be a bit shorter?

fuel/converters/svhn.py
```diff
@@ -0,0 +1,298 @@
...
+
+        def get_boxes_and_labels(split):
+            boxes_and_labels = []
+            num_ex = examples_per_split[split]
+            progress_bar_context = progress_bar(
+                '{} digitStruct'.format(split), num_ex)
+            path = split_structs[split]
+            with h5py.File(path, 'r') as f, progress_bar_context as bar:
+                for image_number in range(num_ex):
+                    bbox_group = f['digitStruct']['bbox'][image_number, 0]
```
bartvm: A `f_ds = f['digitStruct']` would make the rest of this block a bit less verbose.

fuel/datasets/svhn.py
```diff
@@ -0,0 +1,55 @@
...
+       Unsupervised Feature Learning*, NIPS Workshop on Deep Learning
+       and Unsupervised Feature Learning, 2011.
+
+    .. [LBBH] Yann LeCun, Léon Bottou, Yoshua Bengio, and Patrick Haffner,
+       *Gradient-based learning applied to document recognition*,
+       Proceedings of the IEEE, November 1998, 86(11):2278-2324.
+
+    Parameters
+    ----------
+    which_format : 1 or 2
```
bartvm: `{1, 2}`

fuel/datasets/svhn.py
```diff
@@ -0,0 +1,55 @@
...
+        SVHN format 1 contains the full numbers, whereas SVHN format 2
+        contains cropped digits. **Note: at the moment, only format 2
+        is available through Fuel.**
+    which_set : 'train', 'test' or 'extra'
```
bartvm: Same here, should be `{'traing', 'test', 'extra'}` according to NumPy docstring standards I think

fuel/datasets/svhn.py
```diff
@@ -0,0 +1,55 @@
...
+    .. [LBBH] Yann LeCun, Léon Bottou, Yoshua Bengio, and Patrick Haffner,
+       *Gradient-based learning applied to document recognition*,
+       Proceedings of the IEEE, November 1998, 86(11):2278-2324.
+
+    Parameters
+    ----------
+    which_format : 1 or 2
+        SVHN format 1 contains the full numbers, whereas SVHN format 2
+        contains cropped digits. **Note: at the moment, only format 2
+        is available through Fuel.**
```
bartvm: This isn't true anymore after this PR, right?

bartvm:
It's not easy code to read if you're not familiar with the SVHN data format, but overall it looks good to me and well-tested, so good to merge as far as I'm concerned.

vdumoulin:
Even if you *are* familiar with the SVHN data format, the way things are stored inside the format 1 HDF5 file is very convoluted. I basically had to reverse-engineer through trial-and-error to parse it, and mocking out the file for testing was a nightmare.

I'll add comments to make the layout clearer in the format 1 converter tomorrow morning.

bartvm:
Cool, after that, feel free to press the merge button.

vdumoulin:
It looks like Coveralls failed to communicate the results back to Github (Coveralls'  web interface indicates a coverage increase for the latest commit). The other tests pass, I'll merge.

bartvm:
Yeah, some Blocks PRs are having the same thing... *sigh*

PR #141, opened 2015-05-25
Add variable-length data support in H5PYDataset

### LINKED ISSUE #136: Variable length data in Fuel
Guys, I would like to know how you envisage handling datasets with variable-length examples in Fuel, e.g. speech, handwriting, music data (I do not mention text, because we sort of support text files). Do you think they should be also converted to HDF5? There is some sort of support of variable length data in HDF5, but I haven't tried it so far. In both machine translation and speech projects I participated we used tables with concatenated sequences and offset-length tables. This works, but it is quite different from what `HDF5Dataset` currently expects.

I am asking without doing too much of prior research, because I would like to know if you have already thought it through.

vdumoulin:
Fixes #136.

Left to implement before it's ready to merge: a tutorial explaining all this.

### DIFF
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -299,6 +299,112 @@ what you requested, and nothing more.
 >>> print(data.shape)
 (80, 10)
 
+Variable-length data
+--------------------
+
+:class:`~.datasets.hdf5.H5PYDataset` also supports variable length data. Let's
+update the image features to reflect that:
+
+>>> train_image_features = [
+...     numpy.random.randint(10, size=(3, 5, 5)).astype('float32')
+...     for i in range(90)]
+>>> test_image_features = [
+...     numpy.random.randint(10, size=(3, 6, 6)).astype('float32')
+...     for i in range(10)]
+
+In this new example, training images measure 5x5 pixels whereas test images
+measure 6x6 pixels. Image sizes could also vary within the training and test
+sets; this particular arrangement is simply for illustrative purposes.
+
+First, we put the vector features and the targets inside the HDF5 file as
+before:
+
+>>> f = h5py.File('dataset.hdf5', mode='w')
+>>> f['vector_features'] = numpy.vstack(
+...     [numpy.load('train_vector_features.npy'),
+...      numpy.load('test_vector_features.npy')])
+>>> f['targets'] = numpy.vstack(
+...     [numpy.load('train_targets.npy'),
+...      numpy.load('test_targets.npy')])
+>>> f['vector_features'].dims[0].label = 'batch'
+>>> f['vector_features'].dims[1].label = 'feature'
+>>> f['targets'].dims[0].label = 'batch'
+>>> f['targets'].dims[1].label = 'index'
+
+We now have to put the variable-length images inside the HDF5 file. We can't
+do that directly, since HDF5 and h5py don't support multi-dimensional ragged
+arrays. However, there *is* support for 1D ragged arrays. Instead, we'll
+flatten the images before putting them in the HDF5 file:
+
+>>> all_image_features = train_image_features + test_image_features
+>>> dtype = h5py.special_dtype(vlen=numpy.dtype('uint8'))
+>>> image_features = f.create_dataset('image_features', (100,), dtype=dtype)
+>>> image_features[...] = [image.flatten() for image in all_image_features]
+>>> image_features.dims[0].label = 'batch'
+
+If you're feeling lost, have a look at the `dedicated tutorial on
+variable-length data`_.
+
+The images are now in the HDF5 file, but that doesn't help us unless we can
+recover their original shape. For that, we'll create a dimension scale that
+we'll attach to the ``'image_features'`` dataset using the name ``'shapes'``
+(use this *exact* name):
+
+>>> image_features_shapes = f.create_dataset(
+...     'image_features_shapes', (100, 3), dtype='uint8')
+>>> image_features_shapes[...] = numpy.array(
+...     [image.shape for image in all_image_features])
+>>> image_features.dims.create_scale(image_features_shapes, 'shapes')
+>>> image_features.dims[0].attach_scale(image_features_shapes)
+
+We'd also like to tag those variable-length dimensions with semantic
+information. We'll create another dimension scale that we'll attach to the
+``'image_features'`` dataset using the name ``'shape_labels'``
+(use this *exact* name):
+
+>>> image_features_shape_labels = f.create_dataset(
+...     'image_features_shape_labels', (3,), dtype='S7')
+>>> image_features_shape_labels[...] = [
+...     'channel'.encode('utf8'), 'height'.encode('utf8'),
+...     'width'.encode('utf8')]
+>>> image_features.dims.create_scale(
+...     image_features_shape_labels, 'shape_labels')
+>>> image_features.dims[0].attach_scale(image_features_shape_labels)
+
+The :class:`~.datasets.hdf5.H5PYDataset` class will handle things from
+there on. When image features are loaded, it will retrieve their shapes and
+do the reshape automatically.
+
+Lastly, we create the split dictionary exactly as before:
+
+>>> split_dict = {
+...     'train': {'vector_features': (0, 90), 'image_features': (0, 90),
+...               'targets': (0, 90)},
+...     'test': {'vector_features': (90, 100), 'image_features': (90, 100),
+...              'targets': (90, 100)}}
+>>> f.attrs['split'] = H5PYDataset.create_split_array(split_dict)
+>>> f.flush()
+>>> f.close()
+
+That's it. Now let's try to retrieve some data:
+
+>>> train_set = H5PYDataset(
+...     'dataset.hdf5', which_set='train', sources=('image_features',))
+>>> print(train_set.axis_labels['image_features'])
+('batch', 'channel', 'height', 'width')
+>>> handle = train_set.open()
+>>> images, = train_set.get_data(handle, slice(0, 10))
+>>> print(images[0].shape)
+(3, 5, 5)
+>>> train_set.close(handle)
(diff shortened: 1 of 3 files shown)


vdumoulin:
@rizar You might be interested in this PR. Could you have a look at the [tutorial](https://github.com/vdumoulin/fuel/blob/vlen_h5pydataset/docs/h5py_dataset.rst#variable-length-data) I just wrote and tell me if things look clear on your side?

Once this PR is merged, I'll finish implementing SVHN format 1, which will serve as a reference implementation for new variable-length built-in datasets.

docs/h5py_dataset.rst
```diff
@@ -299,6 +299,112 @@ what you requested, and nothing more.
...
+measure 6x6 pixels. Image sizes could also vary within the training and test
+sets; this particular arrangement is simply for illustrative purposes.
+
+First, we put the vector features and the targets inside the HDF5 file as
+before:
+
+>>> f = h5py.File('dataset.hdf5', mode='w')
+>>> f['vector_features'] = numpy.vstack(
+...     [numpy.load('train_vector_features.npy'),
+...      numpy.load('test_vector_features.npy')])
```
lamblin: Are you reloading the fixed-length data here?
Otherwise, the call to `numpy.vstack` will fail if the lengths are different.

lamblin: Nevermind, I did not see the point of `vector_features`, but now I understand it is a different part of the data.

docs/h5py_dataset.rst
```diff
@@ -299,6 +299,112 @@ what you requested, and nothing more.
...
+If you're feeling lost, have a look at the `dedicated tutorial on
+variable-length data`_.
+
+The images are now in the HDF5 file, but that doesn't help us unless we can
+recover their original shape. For that, we'll create a dimension scale that
+we'll attach to the ``'image_features'`` dataset using the name ``'shapes'``
+(use this *exact* name):
+
+>>> image_features_shapes = f.create_dataset(
+...     'image_features_shapes', (100, 3), dtype='uint8')
```
lamblin: If images are larger than 255 along one dimension, we would need a bigger dtype, I think. Is it worth mentioning as a note?

lamblin:
The tutorial looks clear to me, thanks!

docs/h5py_dataset.rst
vdumoulin: Good point. I'll change the dtype for `'int32'`. It'll be safer is someone happens to copy-paste this piece of code, and I won't have to add an awkward disclaimer.

docs/h5py_dataset.rst
```diff
@@ -299,6 +299,112 @@ what you requested, and nothing more.
...
+('batch', 'channel', 'height', 'width')
+>>> handle = train_set.open()
+>>> images, = train_set.get_data(handle, slice(0, 10))
+>>> print(images[0].shape)
+(3, 5, 5)
+>>> train_set.close(handle)
+>>> test_set = H5PYDataset(
+...     'dataset.hdf5', which_set='test', sources=('image_features',))
+>>> handle = test_set.open()
+>>> images, = test_set.get_data(handle, slice(0, 10))
```
rizar: Is there a support for different shapes within one split? What would `get_data` request return in this case?

rizar:
@vdumoulin , I am starting digging, so far it looks good but I have to catch up with all these awesome developments that have happened recently in Fuel thanks to you :)

fuel/datasets/hdf5.py
```diff
@@ -328,35 +362,55 @@ def _file_handle(self):
 
     def get_data(self, state=None, request=None):
         if self.load_in_memory:
-            return self._in_memory_get_data(state=state, request=request)
+            rval, rval_shapes = self._in_memory_get_data(state, request)
         else:
-            return self._out_of_memory_get_data(state=state, request=request)
+            rval, rval_shapes = self._out_of_memory_get_data(state, request)
+        for i, shapes in enumerate(rval_shapes):
```
rizar: maybe `for data, shapes in zip(rval, rval_shapes)`?

rizar: Oh I see, you use it for an assignment further.

Perhaps, a bit of  refactoring/renaming would be good here, `i`, `j`, `rv` and `sh` is a bit too much of abbreviations and one-letter variables.

fuel/datasets/hdf5.py
```diff
@@ -328,35 +362,55 @@ def _file_handle(self):
...
+                for j, (rv, sh) in enumerate(zip(rval[i], shapes)):
+                    rval[i][j] = rv.reshape(sh)
+        return tuple(rval)
 
     def _in_memory_get_data(self, state=None, request=None):
         if state is not None or request is None:
             raise ValueError
-        return tuple(data_source[request] for data_source in self.data_sources)
+        rval = [data_source[request] for data_source in self.data_sources]
+        rval_shapes = [shapes[request] if shapes is not None else None
```
rizar: To be honest, I am not a big fan of `rval`. What about `data` instead?

fuel/datasets/hdf5.py
```diff
@@ -229,6 +230,14 @@ def parse_split_array(split_array):
                 split_dict[split][source] = (start, stop, comment)
         return split_dict
 
+    @staticmethod
+    def unordered_fancy_index(request, indexable):
```
rizar: Perhaps a docstring for this function?

docs/h5py_dataset.rst
vdumoulin: Absolutely! I was a bit lazy yesterday, I'll rework the tutorial to make this clearer. In that case, as in the current case, `get_data` would return a 1D numpy array with dtype `object`.

fuel/datasets/hdf5.py
vdumoulin: Agreed. I'll try to come up with better names.

fuel/datasets/hdf5.py
vdumoulin: Sure!

fuel/datasets/hdf5.py
vdumoulin: Sure!

vdumoulin:
@rizar Thank you for your time. I addressed your comments in the last three commits.

fuel/datasets/hdf5.py
```diff
@@ -326,35 +360,55 @@ def _file_handle(self):
 
     def get_data(self, state=None, request=None):
         if self.load_in_memory:
-            return self._in_memory_get_data(state=state, request=request)
+            data, data_shapes = self._in_memory_get_data(state, request)
         else:
-            return self._out_of_memory_get_data(state=state, request=request)
+            data, data_shapes = self._out_of_memory_get_data(state, request)
+        for i, shapes in enumerate(data_shapes):
```
rizar: That's much better. I would also get rid of the intermediate variable `shapes`, because the difference between `shapes` and `data_shapes` is not immediately obvious. You can call old `data_shapes` `shapes`, and instead of old `shapes` simply use `shapes[i]`. I would even go further to say that mixture of `zip` and `enumerate` looks a bit weird, and maybe it would be better to use only indices here, i.e. `data[i][j] = data[i][j].reshape(shapes[i][j])`.

vdumoulin: Done!

vdumoulin:
@rizar I added the docstring I forgot to add the other day. I think this PR would be ready to be merged. Would you mind having another look to see if it's ready to be merged according to you?

rizar:
I will read it one more time tomorrow, and then you can merge. This will be a great feature!

fuel/datasets/hdf5.py
```diff
@@ -227,6 +228,34 @@ def parse_split_array(split_array):
                 split_dict[split][source] = (start, stop, comment)
         return split_dict
 
+    @staticmethod
+    def unordered_fancy_index(request, indexable):
+        """Safe list indexing.
+
+        Some objects, such as h5py datasets, only support list indexing
+        if the list is ordered.
```
rizar: I am not 100% sure, but I think that one should say "sorted list" not "ordered list". Because all list are ordered, `OrderedDict` is ordered, but only sorted lists contain their elements in ascending order.

fuel/datasets/hdf5.py
```diff
@@ -242,8 +271,16 @@ def load_axis_labels(self):
         handle = self._file_handle
         axis_labels = {}
         for source_name in handle:
-            axis_labels[source_name] = tuple(
-                dim.label for dim in handle[source_name].dims)
+            vlen_source = (len(handle[source_name].dims) == 1
```
rizar: Copy-paste detected: you have these conditions for a source to be considered a variable-length source in at least two places in your code. Maybe `def is_variable_length(source_name)` is worth adding to the class?

rizar: Also a sanity check could be done, that if `'shapes' in handle.dims[0]`, then `len(handle.dims)` must be one.

rizar: Sorry for spamming: seems like `source_name in self.sources_to_reshape` would be equivalent to `is_variable_length(source_name)` I suggest above.

rizar:
Yeah, it looks good to me. Although I confess I did not really check tests. There are a few refactoring suggestions that I would recommend to address before merge.

vdumoulin:
@rizar Thank you for your valuable feedback. I changed the name 'unordered' for 'unsorted' and refactored `'sources_to_reshape'` as `'vlen_sources'` in addition to using it whenever possible.

rizar:
Cool, really good job!

One last thing needed to use this dataset class for speech recognition are splits which are not slices, i.e. a way to define arbitrary sequence of indices as a subset of a dataset.

vdumoulin:
Thanks! Yes, this is also part of future plans. I know you're not the only one who wants this feature, so this is on the top of my list of things to do.

PR #144, opened 2015-05-29
[WIP] Index subset support for H5PYDataset

### LINKED ISSUE #101: More flexibility to validation set selection
Currently I don't see any clean way to do traditional k-fold or leave-one-out cross-validation or any more flexible division of the training data to the validation set. There is only this (start, end) pair available and it is not flexible enough for this purpose (not possible to define two ranges for training and leave in-between samples for validation). This should be a very typical approach in machine learning in general, but with DL datasets are so big that people don't afford to do this. Still, I'd like fuel to support it. What do you think?

I would be personally happy with just adding a shuffle switch that would randomize the order of samples before division. But of course it would need to be internally randomized only once when opened to avoid train and validation from overlapping.

So this is what I traditionally do:
```
train = MNIST("train", end=50000) # first 50k samples
valid = MNIST("train", start=50000) # the last 10k samples
test = MNIST("test")
```

One work-around would be to only create one dataset from "train" and use Transformers somehow, but this data division is so important part of the process that I think fuel should support it directly. Especially because it is pretty easy to implement and support.

### LINKED ISSUE #102: Add support for lists of indices in H5PYDataset subsets.
We should be able to create a split that is a list of disjoint indices, as well as specify the subset of an existing split as a list of disjoint indices.

vdumoulin:
Fixes #101. Fixes #102.

This PR adds support for two types of index subsets:
* At the HDF5 level, it allows the user to specify a split as a list of indices.
* At the dataset level, it allows the user to specify a subset of a split as a list of indices.

The current implementation is a draft. In particular, I had to do some pretty ugly things to the H5PYDataset interface in order to support HDF5-level index split. Basically, if `start` and `stop` are negative, H5PYDataset will look for a `split_subset` attribute in the sources (where `split` is the name of the split), which is a reference to an HDF5 dataset that contains the indices for that split/source pair.

There are two things I'd like to polish before I feel like this PR is ready to be merged:
* I think it's time to rethink the H5PYDataset interface to integrate HDF5-level index splits in a less hacky way.
* I'd like to document the interface itself in the online docs.

@rizar I think this might interest you.

### DIFF
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -229,25 +229,29 @@ Playing with H5PYDataset datasets
 
 Let's explore what we can do with the dataset we just created.
 
-The simplest thing is to load it by giving its path and a split name:
+The simplest thing is to load it by giving its path and a tuple of split names:
 
->>> train_set = H5PYDataset('dataset.hdf5', which_set='train')
+>>> train_set = H5PYDataset('dataset.hdf5', which_set=('train',))
 >>> print(train_set.num_examples)
 90
->>> test_set = H5PYDataset('dataset.hdf5', which_set='test')
+>>> test_set = H5PYDataset('dataset.hdf5', which_set=('test',))
 >>> print(test_set.num_examples)
 10
+
+Passing more than one split name would cause the splits to be concatenated.
+The available data sources would be the intersection of the sources provided
+by each split.
 
 You can further restrict which examples are used by providing a ``slice`` object
 as the ``subset`` argument. *Make sure that its* ``step`` *is either 1 or*
 ``None`` *, as these are the only two options that are supported*.
 
 >>> train_set = H5PYDataset(
-...     'dataset.hdf5', which_set='train', subset=slice(0, 80))
+...     'dataset.hdf5', which_set=('train',), subset=slice(0, 80))
 >>> print(train_set.num_examples)
 80
 >>> valid_set = H5PYDataset(
-...     'dataset.hdf5', which_set='train', subset=slice(80, 90))
+...     'dataset.hdf5', which_set=('train',), subset=slice(80, 90))
 >>> print(valid_set.num_examples)
 10
 
@@ -279,7 +283,7 @@ We can request data as usual:
 We can also request just the vector features:
 
 >>> train_vector_features = H5PYDataset(
-...     'dataset.hdf5', which_set='train', subset=slice(0, 80),
+...     'dataset.hdf5', which_set=('train',), subset=slice(0, 80),
 ...     sources=['vector_features'])
 >>> handle = train_vector_features.open()
 >>> data, = train_vector_features.get_data(handle, slice(0, 10))
@@ -299,7 +303,7 @@ In :class:`~.datasets.hdf5.H5PYDataset`, this is accomplished via the
 what you requested, and nothing more.
 
 >>> in_memory_train_vector_features = H5PYDataset(
-...     'dataset.hdf5', which_set='train', subset=slice(0, 80),
+...     'dataset.hdf5', which_set=('train',), subset=slice(0, 80),
 ...     sources=['vector_features'], load_in_memory=True)
 >>> data, = in_memory_train_vector_features.data_sources
 >>> print(type(data)) # doctest: +ELLIPSIS
@@ -345,7 +349,7 @@ we arbitrarily chose ``-1`` for both.
 Let's check that the training and test set do contain even and odd examples:
 
 >>> train_set = H5PYDataset(
-...     'dataset.hdf5', which_set='train', sources=('vector_features',))
+...     'dataset.hdf5', which_set=('train',), sources=('vector_features',))
 >>> handle = train_set.open()
 >>> print(
 ...     numpy.array_equal(
@@ -356,7 +360,7 @@ Let's check that the training and test set do contain even and odd examples:
 True
 >>> train_set.close(handle)
 >>> test_set = H5PYDataset(
-...     'dataset.hdf5', which_set='test', sources=('vector_features',))
+...     'dataset.hdf5', which_set=('test',), sources=('vector_features',))
 >>> handle = test_set.open()
 >>> print(
 ...     numpy.array_equal(
@@ -463,7 +467,7 @@ That's it. Now let's kick the tires a little. The axis labels appear as they
 should:
 
 >>> train_set = H5PYDataset(
-...     'dataset.hdf5', which_set='train', sources=('image_features',))
+...     'dataset.hdf5', which_set=('train',), sources=('image_features',))
 >>> print(train_set.axis_labels['image_features'])
 ('batch', 'channel', 'height', 'width')
 
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -248,8 +248,7 @@ The available data sources would be the intersection of the sources provided
 by each split.
 
 You can further restrict which examples are used by providing a ``slice`` object
-as the ``subset`` argument. *Make sure that its* ``step`` *is either 1 or*
-``None`` *, as these are the only two options that are supported*.
+or a list of indices as the ``subset`` argument.
 
 >>> train_set = H5PYDataset(
 ...     'dataset.hdf5', which_sets=('train',), subset=slice(0, 80))
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -62,11 +62,16 @@ things if your data happens to meet these assumptions:
 
   1. ``split`` : string identifier for the split name
   2. ``source`` : string identifier for the source name
-  3. ``start`` : start index (inclusive) of the split in the source array
-  4. ``stop`` : stop index (exclusive) of the split in the source array
-  4. ``indices`` : reference pointing to a dataset containing indices for this split/source pair
-  5. ``available`` : boolean, ``False`` if this split is not available for this source
-  6. ``comment`` : comment string
+  3. ``start`` : start index (inclusive) of the split in the source
+     array, used if ``indices`` is a null reference.
+  4. ``stop`` : stop index (exclusive) of the split in the source
+     array, used if ``indices`` is a null reference.
+  5. ``indices`` : h5py.Reference, reference to a dataset containing
+     subset indices for this split/source pair. If it's a null
+     reference, ``start`` and ``stop`` are used.
+  6. ``available`` : boolean, ``False`` is this split is not available
+     for this source
+  7. ``comment`` : comment string
 
 .. tip::
 
--- a/docs/h5py_dataset.rst
+++ b/docs/h5py_dataset.rst
@@ -57,15 +57,15 @@ things if your data happens to meet these assumptions:
   names.
 * Data sources are not explicitly split into separate HDF5 datasets or separate
   HDF5 files. Instead, splits are defined in the ``split`` attribute of the root
-  group. It's expected to be a 1D numpy array of compound ``dtype`` with six
+  group. It's expected to be a 1D numpy array of compound ``dtype`` with seven
   fields, organized as follows:
 
   1. ``split`` : string identifier for the split name
   2. ``source`` : string identifier for the source name
   3. ``start`` : start index (inclusive) of the split in the source array
   4. ``stop`` : stop index (exclusive) of the split in the source array
-  5. ``available`` : boolean, ``False`` if this split is not available for this
-     source
+  4. ``indices`` : reference pointing to a dataset containing indices for this split/source pair
+  5. ``available`` : boolean, ``False`` if this split is not available for this source
   6. ``comment`` : comment string
 
 .. tip::
@@ -139,7 +139,9 @@ attribute of the root group.
 ...     dtype=numpy.dtype([
 ...         ('split', 'a', 5),
 ...         ('source', 'a', 15),
-...         ('start', numpy.int64, 1), ('stop', numpy.int64, 1),
+...         ('start', numpy.int64, 1),
+...         ('stop', numpy.int64, 1),
+...         ('indices', h5py.special_dtype(ref=h5py.Reference)),
 ...         ('available', numpy.bool, 1),
 ...         ('comment', 'a', 1)]))
 >>> split_array[0:3]['split'] = 'train'.encode('utf8')
@@ -151,18 +153,19 @@ attribute of the root group.
 >>> split_array[0:3]['stop'] = 90
 >>> split_array[3:6]['start'] = 90
 >>> split_array[3:6]['stop'] = 100
+>>> split_array[:]['indices'] = h5py.Reference()
 >>> split_array[:]['available'] = True
 >>> split_array[:]['comment'] = '.'.encode('utf8')
 >>> f.attrs['split'] = split_array
 
 We created a 1D numpy array with six elements. The ``dtype`` for this array
 is a compound type: every element of the array is a tuple of ``(str, str, int,
-int, bool, str)``. The length of each string element has been chosen to be the
-maximum length we needed to store: that's 5 for the ``split`` element
-(``'train'`` being the longest split name) and 15 for the ``source`` element
-(``'vector_features'`` being the longest source name). We didn't include any
-comment, so the length for that element was set to 1. Due to a quirk in pickling
-empty strings, we put ``'.'`` as the comment value.
+int, h5py.Reference, bool, str)``. The length of each string element has been
+chosen to be the maximum length we needed to store: that's 5 for the ``split``
+element (``'train'`` being the longest split name) and 15 for the ``source``
+element (``'vector_features'`` being the longest source name). We didn't
+include any comment, so the length for that element was set to 1. Due to a
+quirk in pickling empty strings, we put ``'.'`` as the comment value.
 
 .. warning::
 
@@ -176,6 +179,10 @@ data is available for some source/split combination: for instance, the test
 set may not be labeled, and the ``('test', 'targets')`` combination may not
 exist. In that case, you can set the ``available`` element for that combination
 to ``False``, and :class:`~.datasets.hdf5.H5PYDataset` will ignore it.
+
+Don't worry too much about ``indices``; we'll get back to that later. For the
+moment, all you need to know is that since our splits are congiguous, we don't
+need that feature and therefore put empty references.
 
 The method described above does the job, but it's not very convenient. An even
 simpler way of achieving the same result is to call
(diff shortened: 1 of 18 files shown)


vdumoulin:
Regarding the change in H5PYDataset interface, here's the interface as it stands at this point in the PR:

* The HDF5 file should have a 'split' attribute at the root node.
* This split attribute is a 1D array.
* Each element of the array is of compound dtype.
* Each element of the array contains `(split, source, start, stop, available, comment)` of dtypes `(str, str, int, int, bool, str)`. Their semantic is the following:
  - `split` : name of the split (e.g. train, test)
  - `source` : name of the source (e.g. features, targets)
  - `start` : starting index (inclusive) of the split in the source
  - `stop` : stopping index (exclusive) of the split in the source
  - `available` : whether or not this split/source pair is available
  - `comment` : any relevant comment regarding this split/source pair
* If `available` is `False`, `start` and `stop` are overlooked.
* If `start` and `stop` are both negative, H5PYDataset expects the `source` dataset to have an attribute named `{split}_subset`, which is a reference to another dataset containing the indices for this split/source pair.

Here's what I think would be better:

* The HDF5 file should have a 'split' attribute at the root node.
* This split attribute is a 1D array.
* Each element of the array is of compound dtype.
* Each element of the array contains `(split, source, subset, available, comment)` of dtypes `(str, str, h5py.special_dtype(ref=h5py.Reference), bool, str)`. Their semantic is the following:
  - `split` : name of the split (e.g. train, test)
  - `source` : name of the source (e.g. features, targets)
  - `subset` : [region reference](http://docs.h5py.org/en/latest/refs.html) pointing to the data for this split/source pair. If it's a null reference, this split/source pair is considered to be unavailable.
  - `comment` : any relevant comment regarding this split/source pair

vdumoulin:
Argh, forget about the region reference suggestion. As elegant as it looks, it can only be used to slice through the dataset, not as a 'virtual' dataset which can itself be sliced through. Using that would cause the whole split to be loaded in memory.

Instead, here's another idea:
* The HDF5 file should have a 'split' attribute at the root node.
* This split attribute is a 1D array.
* Each element of the array is of compound dtype.
* Each element of the array contains `(split, source, start, stop, indices, available, comment)` of dtypes `(str, str, int, int, h5py.special_dtype(ref=h5py.Reference), bool, str)`. Their semantic is the following:
  - split : name of the split (e.g. train, test)
  - source : name of the source (e.g. features, targets)
  - start : starting index (inclusive) of the split in the source
  - stop : stopping index (exclusive) of the split in the source
  - indices : object reference pointing to a dataset of indices for this split/source pair
  - available : bool, whether or not this split/source pair is available
  - comment : any relevant comment regarding this split/source pair
* If `available` is `False`, `start`, `stop` and `indices` are overlooked.
* If `available` is `True` and `indices` is not a null reference, `indices` is used and `start` and `stop` are overlooked.

rizar:
Sorry, I am not sure I will be able to review it before NIPS deadline. But right after it, I promise, I will!

vdumoulin:
No problem at all, I'll likely be pretty busy before NIPS myself.

vdumoulin:
Fixes #145. @dwf since I'm already playing in that area, I decided to go ahead and implement the logic necessary to concatenate splits.

Here are the caveats of this PR as it stands:

* The list of indices is kept in-memory, which could be problematic for large datasets whose splits are lists of indices.
* When concatenating multiple splits, no effort is made to check whether these splits are contiguous: a list containing all indices for these splits is created and this is what is used to index through the dataset, which may make accessing data slower.

dwf:
Are we treating sequential splits at all differently now anyway? I thought
slice support had been removed.

vdumoulin:
No, slice support is still there through the `subset` kwarg. Here's the logic behing choosing which data is considered for a given dataset:

* `which_set` is used to define a subset of all the data for a split or a list of splits.
* `subset` is used to specify a subset of that subset to use.

In both cases, subsets can be specified as a slice or as a list of indices.

dwf:
Sorry, I was thinking iteration logic. There I think we have done away with
slices for reasons of silent errors that I don't quite understand still.

bartvm:
The reason was that this would not give an error:
```python
dataset = IndexableDataset(np.arange(10))
stream = DataStream(dataset, iteration_scheme=SequentialScheme(100, 5))
```
Instead it would return `[0, ..., 4]`, `[5, ..., 9]`, and after that it would return `[]` 18 times, because in the background it's returning `np.arange(10)[10:15] = []`. We had a nasty bug at one point because of this. We were scared that checking for each iteration whether the slice is contained within the length of the dataset might slow things down too much (no idea if that's true), so we decided to replace slices with lists for now, but make support for slices optional later, by adding e.g. a `return_slice=True/False` flag to `SequentialScheme`.

That said, I'm happy to reintroduce slices as the default if someone can benchmark the slowdown of performing that check, or introducing that optional flag.

vdumoulin:
@rizar ping: when you have time this week could you have another look?

fuel/datasets/hdf5.py
```diff
@@ -132,8 +136,23 @@ class H5PYDataset(Dataset):
         that case, it is the user's responsibility to make sure that
         indices are ordered.
 
+    Attributes
+    ----------
+    sources : tuple of strings
```
rizar: Should these attributes be documented in the root `Dataset` class?

rizar:
@vdumoulin , honestly, I really don't have time now, since I am moving from Germany to Belarus and tomorrow I will devote myself entirely to finishing bureaucracy and packing bags. On the bright side I should be done by Thursday and I promise that this PR will be my highest priority on that day. Until that time I will try to read it and understand it.

Sorry for all these postponements!

vdumoulin:
No problem at all, I wanted to make sure to follow up with you on this PR, but there's no urgency on my end for it to be reviewed immediately. I wish you a pleasant moving and a bureaucracy experience that doesn't look like that:

Les douze travaux d'Astérix

docs/h5py_dataset.rst
```diff
@@ -57,15 +57,15 @@ things if your data happens to meet these assumptions:
...
   fields, organized as follows:
 
   1. ``split`` : string identifier for the split name
   2. ``source`` : string identifier for the source name
   3. ``start`` : start index (inclusive) of the split in the source array
   4. ``stop`` : stop index (exclusive) of the split in the source array
-  5. ``available`` : boolean, ``False`` if this split is not available for this
-     source
+  4. ``indices`` : reference pointing to a dataset containing indices for this split/source pair
+  5. ``available`` : boolean, ``False`` if this split is not available for this source
```
rizar: The reader at this  point might be confusing by apparent overlapping between `start`, `stop` and `indices` roles. Perhaps an explanation would help.

fuel/datasets/hdf5.py
```diff
@@ -106,16 +106,20 @@ class H5PYDataset(Dataset):
          array
       4. ``stop`` : stop index (exclusive) of the split in the source
          array
-      5. ``available`` : boolean, ``False`` is this split is not available
+      6. ``indices`` : h5py.Reference, reference to a dataset containing
```
rizar: Also here, I think, it should be  said that indices can be "null" which means that the `start` and `stop` attributes specify the subset

fuel/datasets/hdf5.py
```diff
@@ -106,16 +106,20 @@ class H5PYDataset(Dataset):
...
-      6. ``comment`` : comment string
+      7. ``comment`` : comment string
 
     Parameters
     ----------
     file_or_path : :class:`h5py.File` or str
         HDF5 file handle, or path to the HDF5 file.
-    which_set : str
-        Which split to use.
+    which_set : str or tuple of str
```
rizar: It is a reoccurring question in Blocks whether when a tuple or a list are possible arguments a single object also should be accepted. I guess that we are sort of converging to minimalism in the interface, which means that only an iterable should probably be accepter in this case. What's your opinion?

fuel/datasets/hdf5.py
```diff
@@ -106,16 +106,20 @@ class H5PYDataset(Dataset):
...
+        Which split(s) to use. If one than more split is requested,
+        the provided sources will be the intersection of provided
+        sources for these splits.
     subset : slice, optional
```
rizar: This is not related to this PR, but why is this `subset` option required? I thought such things should be done at the datastream level.

fuel/datasets/hdf5.py
```diff
@@ -146,23 +165,55 @@ def __init__(self, file_or_path, which_set, subset=None,
...
+                    "'{}' split is not provided by this ".format(split) +
+                    "dataset. Available splits are " +
+                    "{}.".format(available_splits))
+            split_provides_sources = set(
+                self.get_provided_sources(handle, split))
+            if provides_sources:
+                provides_sources &= split_provides_sources
+            else:
+                provides_sources = split_provides_sources
+        self.provides_sources = tuple(sorted(provides_sources))
```
rizar: I am trying to find where these attributes were set in the previous version of the code but so far I fail. Can you please help?

fuel/datasets/hdf5.py
```diff
@@ -266,98 +476,93 @@ def unsorted_fancy_index(request, indexable):
...
+            indices = defaultdict(list)
+            for split in self.which_set:
+                split_start_stop = self.get_start_stop(handle, split)
+                split_indices = self.get_indices(handle, split)
+                for source_name in self.sources:
+                    if source_name in split_indices:
+                        ind = split_indices[source_name]
+                    else:
+                        ind = list(range(*split_start_stop[source_name]))
+                    indices[source_name] = sorted(
```
rizar: Is this `sorted` call necessary? If we support unsorted index array, then it seems to me we could do without it.

fuel/datasets/hdf5.py
bartvm: I can imagine `subset` is useful if you want to load a subset of HDF5 into memory. Loading all of the dataset in memory might be unfeasible, and doing it on a data stream level would force you to read from disk (relying on HDF5's caching mechanism).

fuel/datasets/hdf5.py
bartvm: https://github.com/mila-udem/fuel/pull/144/files#diff-2e06746e89611307aa3b9d5fbc38faa3L300

fuel/datasets/hdf5.py
bartvm: Haven't read the code in detail, but HDF5 doesn't support unsorted index arrays, so it might have to do with that.

docs/h5py_dataset.rst
vdumoulin: Your understanding is correct. I'll make sure to make that clearer here.

fuel/datasets/hdf5.py
vdumoulin: I'm not very opinionated on that, but if it helps uniformize the interface in Blocks and Fuel I'd be for it. Since we're already in the process of changing the interface, I guess there's no harm in taking that decision now.

However, since a string is itself an iterable and the current interface expects a string, I'm wary of introducing silent bugs where the user passes `'train'` and the code understands that as five splits, `'t'`, `'r'`, `'a'`, `'i'` and `'n'`. Except for type checking and raising an informative error message if the user passes a string, I don't see a good way of making sure this doesn't happen. What are your thoughts on this?

fuel/datasets/hdf5.py
vdumoulin: I confirm: this is why it's in there.

fuel/datasets/hdf5.py
```diff
@@ -411,21 +616,27 @@ def _in_memory_get_data(self, state=None, request=None):
         return data, shapes
 
     def _out_of_memory_get_data(self, state=None, request=None):
+        if not isinstance(request, (slice, list)):
```
rizar: Do we still have iteration schemes that produce slices? They are not save, because range checks doesn't work for them. If we don't have schemes, I would consider drop the support for slices here.

fuel/datasets/hdf5.py
vdumoulin: @rizar You're right that `get_data` supports unsorted indices by default (this can be disabled by setting `sort_indices` to `False` for better performance), which makes this call redundant by default.

I can see a case where crashes might be hard to track: imagine that the user makes sure that indices within a request are sorted and sets `sort_indices` to `False`, but that split indices themselves are unsorted. From the user's point of view, everything is sorted, but under the hood we're accessing an unsorted list of indices in sorted order.

Would there be use cases where an unsorted split indices list is a useful thing to have, or can we assume that split indices should always be sorted?

fuel/datasets/hdf5.py
vdumoulin: I think you're right. I remember @bartvm talking about this in another PR. If it's true, then I'll remove support for slices.

fuel/datasets/hdf5.py
rizar: Indeed, type checking would be helpful at that point. Although judging from the frequency of me stumbling upon exactly this issue, it is often not done.

Ideally I think we should be consistent in Blocks regarding such things, and I https://github.com/mila-udem/blocks/blob/master/blocks/filter.py#L97 might serve as an example to follow.

fuel/datasets/hdf5.py
rizar: All right, this makes sense, thanks for the explanations.

fuel/datasets/hdf5.py
rizar: I  think that in practice the split indices might be unsorted. I can easily imagine yet another speech dataset with different canonical orders for all the utterances and for particular subsets... 

But your argument is valid. And also if we concatenate two sorted lists of indices, we might end up with unsorted batches at the splice.

I will think further what should be the right way to go here.

fuel/datasets/hdf5.py
rizar: That would be also cool for making the code simpler, because as now it have quite a lot of nested ifs.

rizar:
The PR is cool and fills an important gap! All the enhancements I can propose so far are in the comments to the code.

fuel/datasets/hdf5.py
bartvm: Although true that they aren't safe, I do believe that they are significantly faster for some HDF5 implementations. My suggestion was to re-implement support for slices, but perhaps make it optional through a flag. That way, new users won't get bitten by it, but power users have the option to speed things up easily.

dwf: It seems easy enough to identify conditions where a slice will return an empty batch and just raise an exception, no?

bartvm: The problem is, where do we do that? Not all datasets will support slices and/or indices, and there's no single base class for those that do I think, but we don't want to re-implement that check for every single dataset.

vdumoulin: There's no single base class for those that do, but there's an abstract `Dataset` class from which all datasets inherit. We could implement a static method in there so it's available to all datasets if need be.

rizar: Shouldn't this discussion be preceded by benchmarks of how much faster the slices are? Premature optimization - you know the continuation.

fuel/datasets/hdf5.py
rizar: The more I think about this issue the more it seems to me that your solution is a reasonable one. Just mention this procedure in the `which_set` docstring, and I guess it will be all right.

fuel/datasets/hdf5.py
```diff
@@ -411,21 +620,27 @@ def _in_memory_get_data(self, state=None, request=None):
...
+            raise ValueError()
         data = []
         shapes = []
         handle = self._file_handle
         for source_name, subset in zip(self.sources, self.subsets):
-            if isinstance(request, slice):
-                req = slice(request.start + subset.start,
-                            request.stop + subset.start, request.step)
+            if hasattr(subset, 'step'):
```
rizar: In this piece of code you use two ways of checking that smth is a slice: `hasattr(smth, 'step')` and `isinstance(smth, slice)`. If they are identical, which seems to be the case, only one should be used, I guess.

fuel/datasets/hdf5.py
vdumoulin: I made the change. I also changed the name from `which_set` to `which_sets` for two reasons:

1. It fits better with that this argument is now used for.
2. It decreases further the chance that someone will try to pass a string as `which_sets`.

fuel/datasets/hdf5.py
vdumoulin: No problem, you're right that the code should stay consistent. I made the change.

vdumoulin:
@rizar if I'm not mistaken the only thing left to address is whether or not we want to maintain support for slice objects.

I think this is something we can settle later on. I propose that we merge the PR when it's ready, and I'll do the benchmarking and remove support (if we decide so) as part of another issue (I just opened it: #164).

docs/h5py_dataset.rst
rizar: Seems like the numbering in this list is wrong, you have two items `4.`.

fuel/datasets/hdf5.py
rizar: Thanks! I am a bit surprised though that you chose the `hasattr` variant. May  I ask why you like it more than the `isinstance` one?

docs/h5py_dataset.rst
vdumoulin: You're right. I made the change.

fuel/datasets/hdf5.py
```diff
@@ -411,21 +626,27 @@ def _in_memory_get_data(self, state=None, request=None):
...
+                if hasattr(request, 'step'):
```
rizar: I would put an assertion that `subset.step` equals 1 here, even if it was checked before. I will help the reader to understand the code. Or maybe just a reminder comment.

fuel/datasets/hdf5.py
vdumoulin: My understanding is that duck typing is preferred to `isinstance` checks in Python.

rizar: Sounds reasonable, though in this particular case it is hard to expect any non-slice having the same interface as slice. Also your duck-typing snippet does not check `start` and `finish` attributes. 

But I am fine with merging `hasattr` for now.

fuel/datasets/hdf5.py
vdumoulin: Actually, I think that we could lift this restriction now, since we support lists of indices in `subset`.

rizar: True, but then I guess such a slice will have to be converted to a list of indices, right?

vdumoulin: Right. I'll implement the logic for that.

rizar: If you wish! For me as the reviewer a comment plus an issue created would be sufficient. But of course, the less restriction in the interface, the better.

vdumoulin: As long as it isn't making it more difficult for you to review I'm more than happy to make the change.

rizar: I will be happy to review it :)

vdumoulin: e3d45d7

rizar:
Cool, waiting for tests to pass and merge.

vdumoulin:
I suspect that I'll have to clear the cache and run the tests again. I'll manage that and ping you when the tests pass.

vdumoulin:
I am happy to announce that the tests pass.

rizar:
Great job!

PR #149, opened 2015-06-02
Improve IndexableDataset with non-ndarray sources.

dwf:
If the request is an instance of ~~collections.Container~~ collections.Iterable but not a numpy.ndarray, do not attempt to do fancy indexing; instead, perform the indexing in a loop. This allows one to use an IndexableDataset with list sources in conjunction with a BatchScheme.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -7,7 +7,7 @@ from six.moves import zip, range
 
 from fuel.datasets import Dataset, IterableDataset, IndexableDataset
 from fuel.streams import DataStream
-from fuel.schemes import BatchSizeScheme, ConstantScheme
+from fuel.schemes import ShuffledScheme, BatchSizeScheme, ConstantScheme
 from fuel.transformers import Mapping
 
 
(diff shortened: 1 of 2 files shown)


dwf:
In truth I think maybe collections.Iterable is a better fit here. Let me update.

tests/test_datasets.py
```diff
@@ -117,6 +117,18 @@ def test_value_error_get_data_none_request(self):
...
             ValueError, IndexableDataset([1, 2, 3]).get_data, None, None)
 
+    def test_batch_iteration_scheme_with_lists(self):
+        """Batch schemes should work with more than ndarrays."""
+        data = IndexableDataset(OrderedDict([('foo', list(range(50))),
+                                             ('bar', list(range(1, 51)))]))
+        stream = DataStream(data,
+                            iteration_scheme=ShuffledScheme(data.num_examples,
+                                                            5))
+        returned = [sum(batches, []) for batches in
```
bartvm: There's also `concat` in Toolz, but it's a test, so doesn't really matter.

PR #150, opened 2015-06-03
Reshape the targets of CIFAR10 

ronekko:
For the MNIST, the `fuel-convert` generates targets as 2D array and it can work well with Blocks.
But for the CIFAR10, the shape of the arrays of targets is `(num_examples, )` which is 1D array, and it seems inconsistent with Blocks.
The dimension was originally 2D (in both implementation and test), and at #98 it was modified as 1D.


bartvm:
Thanks! I'll let @vdumoulin have a quick look, but in principle I agree that targets should be 2D for consistency, and this probably just slipped through the net.

vdumoulin:
You're absolutely right, this should have been 2D from the start. Thanks a lot!

PR #151, opened 2015-06-04
Refactored HDF5 dataset

dmitriy-serdyuk:
I renamed it to `PytablesDataset`, refactored and added some documentation.

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -45,7 +45,6 @@ class PytablesDataset(Dataset):
         self.provides_sources = sources
         self.path = path
         self.data_node = data_node
-        self.open_file(path)
         self.start = start
         self.stop = stop
         self.nodes = None
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -54,7 +54,7 @@ class PytablesDataset(Dataset):
 
     def open_file(self, path):
         self.h5file = tables.open_file(path, mode="r")
-        node = self.h5file.getNode('/', self.data_node)
+        node = self.h5file.get_node('/', self.data_node)
 
         self.nodes = [getattr(node, source) for source in self.sources_in_file]
         if self.stop is None:
@@ -65,7 +65,7 @@ class PytablesDataset(Dataset):
         self.open_file(self.path)
 
     def close_file(self):
-        self.h5file.close_file()
+        self.h5file.close()
         del self._h5file
         del self._nodes
 
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -66,6 +66,8 @@ class PytablesDataset(Dataset):
 
     def close_file(self):
         self.h5file.close_file()
+        del self._h5file
+        del self._nodes
 
     def get_data(self, state=None, request=None):
         """ Returns data from HDF5 dataset.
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
CC @vdumoulin

vdumoulin:
I'll have a look after lunch, thanks!

bartvm:
What's the use case for this class right now? If possible, we should probably just stick with `h5py`, because this duplicates a lot of functionality, but I know there's some more advanced features that are only in PyTables...

fuel/datasets/hdf5.py
```diff
@@ -38,24 +45,27 @@ def __init__(self, sources, start, stop, path, data_node='Data',
...
+            self.stop = self.nodes[0].nrows
+        self.num_examples = self.stop - self.start
 
     def load(self):
         self.open_file(self.path)
 
-    def close(self):
-        self.h5file.close()
+    def close_file(self):
+        self.h5file.close_file()
```
vdumoulin: We should also be deleting `self._h5file` and `self.nodes`, so that they're properly reloaded if the dataset tries to access these attributes again in its lifetime.

dmitriy-serdyuk:
We have our datasets in pytables and I suspect that there can be people who still use this class. Also it is much more simple and flexible, it may be a good starting point to create an interface to one's own pytables dataset.

dmitriy-serdyuk:
Why the coverage decreased so much?

bartvm:
Seems like your test isn't being run: https://coveralls.io/builds/2730193/source?filename=fuel%2Fdatasets%2Fhdf5.py

vdumoulin:
I think we should leave it in, since as @dmitriy-serdyuk said, some people are using it. When we decided that we would be focusing on `H5PYDataset`, we also decided to leave other options to people, even if those options weren't as polished or maintained.

Regarding the name change, is it sufficient to announce it on the mailing list?

bartvm:
Yeah, an e-mail to the mailing list is enough.

vdumoulin:
There's some weird stuff going on: Travis says the tests pass, but looking at the logs I see errors.

dmitriy-serdyuk:
I cannot open Travis. I think they have problems again.

bartvm:
#152

bartvm:
Rebasing on current master should fix Travis's build status. It seems like there are other tests broken though.

dmitriy-serdyuk:
Seems, that nosetests just doesn't call `tearDown()`.

tests/test_hdf5.py
```diff
@@ -6,11 +6,11 @@
 from numpy.testing import assert_equal, assert_raises
 from six.moves import range, cPickle
 
-from fuel.datasets.hdf5 import Hdf5Dataset, H5PYDataset
+from fuel.datasets.hdf5 import PytablesDataset, H5PYDataset
 from fuel.streams import DataStream
 
 
-class TestHdf5Dataset(object):
+class TestPytablesDataset(object):
```
bartvm: Shouldn't this inherit from `unittest.TestCase`?

dmitriy-serdyuk:
The problem was that `setUp()` was failing at the end, so `tearDown()` was not called, it didn't close files, so next `setUp()` was failing...

tests/test_hdf5.py
vdumoulin: nosetests also recognizes test classes and test class methods by name (see [here](http://nose.readthedocs.org/en/latest/writing_tests.html#test-classes)).

vdumoulin:
Looks good, thanks!

PR #152, opened 2015-06-04
Fail within if-block

bartvm:


bartvm:
Sadly `master` actually seems broken, but we missed that because of Travis's exit code wasn't correct.

vdumoulin:
It's a quick fix, I'll make a PR.

PR #158, opened 2015-06-10
Refactor fuel-convert output flag

vdumoulin:
Some converters can produce more than one format, and the current solution for dynamically assigning a default filename was getting awkward.

The way this is currently implemented is through a single default filename for all possible formats (e.g. `'svhn_format_{}.hdf5'`) which is formatted depending on the value of some command line arguments. This is awkward to maintain and prone to bugs.

This PR aims at solving this by allowing both file and directory names to be specified via the `-o` flag and using a custom `argparse.Action` subclass to split that into an `output_directory` and an (optional) `output_filename` argument. Converters now accept both arguments and can define a sane default filename, and they return the path pointing to the converted dataset.

### DIFF
--- a/bin/fuel-convert
+++ b/bin/fuel-convert
@@ -10,16 +10,12 @@ from fuel.converters.base import MissingInputFiles
 from fuel.datasets import H5PYDataset
 
 
-class SplitOutputAction(argparse.Action):
+class CheckDirectoryAction(argparse.Action):
     def __call__(self, parser, namespace, values, option_string=None):
-        absolute_path = os.path.abspath(values)
-        if os.path.isdir(absolute_path):
-            output_directory = absolute_path
+        if os.path.isdir(values):
+            setattr(namespace, self.dest, values)
         else:
-            output_directory = os.path.dirname(absolute_path)
-            setattr(namespace, 'output_filename',
-                    os.path.basename(absolute_path))
-        setattr(namespace, self.dest, output_directory)
+            raise ValueError('{} is not a existing directory'.format(values))
 
 
 if __name__ == "__main__":
@@ -36,9 +32,8 @@ if __name__ == "__main__":
             name, parents=[parent_parser],
             help='Convert the {} dataset'.format(name))
         subparser.add_argument(
-            "-o", "--output", dest="output_directory",
-            help="where to save the dataset", type=str,
-            default=os.getcwd(), action=SplitOutputAction)
+            "-o", "--output-directory", help="where to save the dataset",
+            type=str, default=os.getcwd(), action=CheckDirectoryAction)
         subparser_fn(subparser)
 
     args = parser.parse_args()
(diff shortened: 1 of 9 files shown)


jbornschein:
Haven't looked in detail at the implementation yet -- but I think your approach makes a lot of sense. And from the users perspective it's even fully compatible with the current behavior.

bin/fuel-convert
```diff
@@ -24,9 +36,11 @@ if __name__ == "__main__":
             name, parents=[parent_parser],
             help='Convert the {} dataset'.format(name))
         subparser.add_argument(
-            "-o", "--output-file", help="where to save the dataset", type=str,
-            default=os.path.join(os.getcwd(), '{}.hdf5'.format(name)))
+            "-o", "--output", dest="output_directory",
+            help="where to save the dataset", type=str,
```
dwf: More docs pls.

vdumoulin: Would something like

    where to save the dataset. Can be either a file or a directory.
    If it's a directory, fuel-convert will save using a dataset-specific
    default file name.

work for you?

dwf: Yes.

bin/fuel-convert
```diff
@@ -10,6 +10,18 @@ from fuel.converters.base import MissingInputFiles
 from fuel.datasets import H5PYDataset
 
 
+class SplitOutputAction(argparse.Action):
+    def __call__(self, parser, namespace, values, option_string=None):
+        absolute_path = os.path.abspath(values)
+        if os.path.isdir(absolute_path):
```
dwf: I'm not totally sure about this. The behaviour then depends on what's on the filesystem, which is kind of strange... see what @rizar or @bartvm think.

rizar: I shared @dwf concerns. Would it be sufficient to determine that the input is a folder if it ends with '/' and a file otherwise?

bartvm: I'm a bit confused by how things would work after this PR. If the user overwrites the filename, that means the corresponding `Dataset` class won't be able to find the file at all, right?

@rizar Nope, a path to a directory doesn't need to end with a `/`, and even if you would require that, it depends on the operating system (e.g. `\` in Windows).

Can't we split this into two flags? One for the output directory, one for the output file name. That avoids the need for tricky argument parsing, and makes it easier to give a warning about using that option (e.g. that the `Dataset` might not be able to load things automatically anymore).

rizar: @bartvm , I know, but I suggest that user can explicitly indicate that a path points to a directory by ending it with the separator `os.path.sep`. I am fine with two arguments though.

vdumoulin: What I had in mind was something similar to `cp`'s interface: if you pass it a directory as the second argument it'll re-use the filename in the first argument and copy *inside* that directory.

If you pass a directory as the output path, `fuel-convert` will use the default file name provided by the converter. Otherwise, it'll use the user-provided filename.

@bartvm you're right that if the user overwrites the filename, the corresponding `Dataset` class won't be able to find the file. I must admit that I can't think of a good use case for choosing a filename other than the default one. If nobody else can find a use case for that, I'd be happy to simplify the interface and just allow the user to specify the output directory.

jbornschein:
I think it makes a lot of sense to let the user choose the output directory but not the actual filename. 

Would you simplify it in a way that a converter can create multiple datasets then? I don't think it's very important -- but it seems natural to allow it then.

rizar:
+1 for only choosing the directory

vdumoulin:
@jbornschein yes, in that case I think it would be straightforward to extend the logic to multiple files by changing the converter's return value from a string to a list of strings.

For convenience, you could add an `all` option to the subparser in your PR which loops over all formats and does the conversion.

bin/fuel-convert
vdumoulin: To follow up on this discussion, I just removed the option for the user to set the file name manually.

vdumoulin:
I've updated the code to allow multiple saved files. I think the PR would be ready for a final review, if anyone has other things to point out.

PR #161, opened 2015-06-13
Add stream ability to select subset of data sources

bouthilx:
If I understand correctly, we need to modify dataset object in order to get different subset of sources without having to reload all data. 

The idea is not to force dataset to load all provided sources and move sources selection to stream but rather add a stream ability to select subset of sources. This would be useful in cases where we manipulate a single dataset object for different training settings, e.g. unsupervised training followed by supervised training.

### DIFF
--- a/fuel/datasets/base.py
+++ b/fuel/datasets/base.py
@@ -149,7 +149,7 @@ class Dataset(object):
         pass
 
     @abstractmethod
-    def get_data(self, state=None, request=None):
+    def get_data(self, state=None, request=None, sources=None):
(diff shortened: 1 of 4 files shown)


bouthilx:
I'll fix the other datasets after we agree on proposed modifications.

dwf:
I think the Fuel-ish way to do this would be c to just have a
SourcesSubsetTransformer that wraps an existing stream.

bouthilx:
Oh right! It's much simpler. I'll make another PR.

PR #162, opened 2015-06-13
Add a SourcesSubsetTransformer

bouthilx:
SourcesSubsetTransformer selects a subset of given stream sources.

The order of stream.sources is not maintained. The output order of
SourcesSubsetTransformer is defined by the order of its given sources
parameter.

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -619,3 +619,36 @@ class Rename(Transformer):
             raise ValueError
         data = next(self.child_epoch_iterator)
         return data
+
+
+class FilterSources(Transformer):
+    """Selects a subset of the stream sources.
+
+    Order of data stream's sources is maintained. The order of sources
+    given as parameter to FilterSources does not matter.
+
+    Parameters
+    ----------
+    data_stream : :class:`AbstractDataStream` or :class:`Transformer`.
+        The data stream.
+    sources : tuple of strings
+        The names of the data sources returned by this transformer.
+        Must be a subset of the sources given by the stream.
+
+    """
+    def __init__(self, data_stream, sources):
+        super(FilterSources, self).__init__(data_stream)
+        if any(source not in data_stream.sources for source in sources):
+            raise ValueError("sources must all be contained in "
+                             "data_stream.sources")
+
+        # keep order of data_stream.sources
+        self.sources = [s for s in data_stream.sources if s in sources]
+
+    def get_data(self, request=None):
+        if request is not None:
+            raise ValueError
+
+        data = next(self.child_epoch_iterator)
+        return [d for d, s in izip(data, self.data_stream.sources)
+                if s in self.sources]
--- a/tests/test_transformers.py
+++ b/tests/test_transformers.py
@@ -12,7 +12,7 @@ from fuel.streams import DataStream
 from fuel.transformers import (
     Transformer, Mapping, SortMapping, ForceFloatX, Filter, Cache, Batch,
     Padding, MultiProcessing, Unpack, Merge, SingleMapping, Flatten,
-    ScaleAndShift, Cast, Rename)
+    ScaleAndShift, Cast, Rename, FilterSources)
 
 
 class FlagDataStream(DataStream):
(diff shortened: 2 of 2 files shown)


fuel/transformers/__init__.py
```diff
@@ -619,3 +619,29 @@ def get_data(self, request=None):
...
         data = next(self.child_epoch_iterator)
         return data
+
+
+class SourcesSubsetTransformer(Transformer):
+    """Selects a subset of the stream sources.
+
+    Parameters
+    ----------
+    data_stream : :class:`DataStream` or :class:`Transformer`.
```
vdumoulin: I think this would also work with `ServerDataStream` out of the box. What do you think of :class:`AbstractDataStream` as an alternative type for that argument?

bouthilx: Is it also the case for most Transformers above?

vdumoulin:
I find the name of the class a bit verbose (the `Transformer` prefix looks redundant to me). I think `SourcesSubset` would be better. Alternatively, `FilterSources` would reflect the method of the same name in `Dataset`. What do you think?

Other than that and my inline comment, LGTM.

fuel/transformers/__init__.py
vdumoulin: I think it is.

fuel/transformers/__init__.py
```diff
@@ -619,3 +619,29 @@ def get_data(self, request=None):
...
+        The data stream.
+    sources : tuple of strings
+        The names of the data sources returned by this transformer.
+        Must be a subset of the sources given by the stream.
+
+    """
+    def __init__(self, data_stream, sources):
+        super(SourcesSubsetTransformer, self).__init__(data_stream)
+        self.sources = sources
+        self._source_ids = [self.data_stream.sources.index(source)
```
vdumoulin: I suggest that before doing this we check that all sources are available in `data_stream` and raise an informative error message otherwise.

bouthilx:
It think SourcesSubset would be better if stream.sources order is not maintained. However, I personally prefer FilterSources semantic, so I would be inclined to adapt the code to keep stream.sources order fixed and rename the class.

bouthilx:
@vdumoulin I amended my commit

fuel/transformers/__init__.py
```diff
@@ -619,3 +619,33 @@ def get_data(self, request=None):
...
+        Must be a subset of the sources given by the stream.
+
+    """
+    def __init__(self, data_stream, sources):
+        super(FilterSources, self).__init__(data_stream)
+        if any(source not in data_stream.sources for source in sources):
+            raise ValueError("sources must all be contained in "
+                             "data_stream.sources")
+
+        # keep order or data_stream.sources
```
vdumoulin: Typo: `order or -> order of`.

fuel/transformers/__init__.py
```diff
@@ -619,3 +619,33 @@ def get_data(self, request=None):
             raise ValueError
         data = next(self.child_epoch_iterator)
         return data
+
+
+class FilterSources(Transformer):
+    """Selects a subset of the stream sources.
```
vdumoulin: I'd add a paragraph explaining that the data stream's source order is maintained.

vdumoulin:
@bouthilx sounds good. When my latest comments have been adressed we'll be ready to merge.

fuel/transformers/__init__.py
bouthilx: yes sure!

bouthilx:
@vdumoulin Done

dwf:
So there's a test failure, but I have no goddamned idea how that happened given the diff.

bartvm:
Let me restart the test, just to see if it's deterministic...

bouthilx:
@dwf @bartvm It passed. What happened? I missed it.

bartvm:
A mysterious error in a completely unrelated part of the library.
```
FAIL: tests.test_server.TestServer.test_pickling
  File "/home/travis/build/mila-udem/fuel/tests/test_server.py", line 42, in test_pickling
    assert_allclose(*data)
  File "/home/travis/miniconda/lib/python3.4/site-packages/numpy/testing/utils.py", line 1297, in assert_allclose
    verbose=verbose, header=header)
  File "/home/travis/miniconda/lib/python3.4/site-packages/numpy/testing/utils.py", line 665, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Not equal to tolerance rtol=1e-07, atol=0

(mismatch 100.0%)
 x: array([[[[0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0],...
 y: array([[[[0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0],
         [0, 0, 0, ..., 0, 0, 0],...

Ran 177 tests in 2.104s

FAILED (failures=1)
```

dwf:
Cosmic rays!

bouthilx:
alt text

vdumoulin:
Tests pass, everything is looking fine and we've got a nice cat picture. I'd say this is ready to be merged. Thanks a lot!

PR #163, opened 2015-06-14
Add tutorial for data processing server

### LINKED ISSUE #60: Document how to write a script that spawns a dataset server before training
Launching a combo dataset server/experiment script is going to be a pretty common use case. It'd be nice to include some sort of helper to make this as painless as possible.

vdumoulin:
Fixes #60.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -10,6 +10,7 @@ Welcome to Fuel's documentation!
    h5py_dataset
    new_dataset
    extending_fuel
+   server
    api/index
 
 Fuel is a data pipeline framework which provides your machine learning models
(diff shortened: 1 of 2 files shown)


vdumoulin:
@dwf since you were the one creating the ticket, would you mind having a look at this to tell me if this is what you had in mind?

bartvm:
This PR looks good to me actually, is there anything holding it back (besides the merge conflicts)? Else we should probably get it in. The Fuel server can be useful, but I still feel many people aren't aware of its existence or how to use it, this would help.

vdumoulin:
Nothing that I'm aware of is holding it back, I think it just fell between the cracks. I'll rebase.

vdumoulin:
Rebased, and the tests pass.

PR #165, opened 2015-06-15
[requires.io] dependency update on master branch

bartvm:


dwf:
Yeah... that didn't go so well.

dwf:
@bartvm Any idea how this got so botched?

bartvm:
Yeah, if Conda can't meet one of the requirements (new Pandas in this case) the entire command fails and it doesn't install anything. Travis happily goes on running tests though, which will consistently fail because even NumPy isn't available.

bartvm:
Problem in general is that Conda's repositories seem to be 1-2 days behind PyPI, which means that these PRs initially always fail :/

nouiz:
I received an email by requires.io for the hooks. I activated it.

PR #167, opened 2015-06-16
Move Fuel install before Blocks install to tests against current Fuel…

vdumoulin:
… branch.

@bartvm Is this what you meant by "move `pip install .` before the Blocks install section"?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,8 @@ before_install:
   - export FUEL_DATA_PATH=$TRAVIS_BUILD_DIR/data
 install:
   # Install all Python dependencies
-  - |
   - pip install . -r requirements.txt
+  - |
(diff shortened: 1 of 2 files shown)


vdumoulin:
Looks like it's more complicated than just moving the `pip install .` line before the Blocks install block: Fuel requires some packages that are installed within the Blocks install block

rizar:
Should we also start testing `blocks-examples` here?

vdumoulin:
We should first fix the `blocks-examples` test run time, but after that yes, it might be a good idea.

rizar:
But aren't they fixed by https://github.com/mila-udem/blocks-examples/pull/12?

vdumoulin:
What I mean is: since the LeNet example has been introduced, the test run time has jumped from ~10 minutes to ~38 minutes.

PR #170, opened 2015-06-16
Upgrade Fuel's handling of data path

### LINKED ISSUE #168: Expand "~" in data path
Hi,

I've started looking into Blocks and Fuel and noticed a small inconvenience when I tried to run the MNIST example. I had set the data path in `.fuelrc` to `~/.fuel/data`, but the `mnist.hdf5` file could not be found because I didn't provide the absolute path. It's not a big deal, but perhaps it's worthwhile to preprocess the data path with `os.path.expanduser`.

Best,
Sigurd

### LINKED ISSUE #169: Allow more than one location in FUEL_DATA_PATH

vdumoulin:
Fixes #168. Fixes #169.

The data path now supports a sequence of paths separated by an os-specific delimiter (':' for Linux and OSX, ';' for Windows) via `os.path.pathsep`.

It also expands '~' and references to environment variables via `os.path.expanduser` and `os.path.expandvars`.

### DIFF
--- a/tests/test_svhn.py
+++ b/tests/test_svhn.py
@@ -20,8 +20,7 @@ def test_svhn():
         f.attrs['split'] = H5PYDataset.create_split_array(split_dict)
         f.close()
         dataset = SVHN(which_format=2, which_sets=('train',))
-        assert_equal(dataset.data_path,
-                     os.path.join(config.data_path, 'svhn_format_2.hdf5'))
+        assert_equal(dataset.filename, 'svhn_format_2.hdf5')
     finally:
         config.data_path = data_path
         os.remove('svhn_format_2.hdf5')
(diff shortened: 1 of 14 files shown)


tests/test_svhn.py
```diff
@@ -20,8 +20,7 @@ def test_svhn():
         f.attrs['split'] = H5PYDataset.create_split_array(split_dict)
         f.close()
         dataset = SVHN(which_format=2, which_sets=('train',))
-        assert_equal(dataset.data_path,
-                     os.path.join(config.data_path, 'svhn_format_2.hdf5'))
+        assert_equal(dataset.filename, 'svhn_format_2.hdf5')
```
lamblin: It should probably be `dataset._filename` now.

vdumoulin: `dataset._filename` is `svhn_format_{}.hdf5`, which will be formatted via the `dataset.filename` property.

lamblin:
Coverage score seemed to go down only because the number of tested lines was reduced, not because of new untested lines, so I'm merging.

PR #174, opened 2015-06-18
Make missing directories when downloading datasets

janchorowski:
Makes missing directories when downloading datasets.

### DIFF
--- a/fuel/downloaders/base.py
+++ b/fuel/downloaders/base.py
@@ -81,6 +81,22 @@ def download(url, file_handle, chunk_size=1024):
             file_handle.write(chunk)
 
 
+def ensure_directory_exists(directory):
+    """Create directory (with parents) if does not exist, raise on failure.
+
+    Parameters
+    ----------
+    directory : str
+        The directory to create
+
+    """
+    try:
+        os.makedirs(directory)
+    except OSError as e:
+        if e.errno != os.errno.EEXIST:
+            raise
+
+
 def default_downloader(directory, urls, filenames, url_prefix=None,
                        clear=False):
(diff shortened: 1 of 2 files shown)


fuel/downloaders/base.py
```diff
@@ -117,6 +117,13 @@ def default_downloader(directory, urls, filenames, url_prefix=None,
                 os.remove(f)
     else:
         print('Downloading ' + ', '.join(filenames) + '\n')
+        try:
+            os.makedirs(directory)
+        except OSError as e:
```
vdumoulin: Another (IMHO) more straightforward way of implementing this would be

``` python
if not os.path.exists(directory):
    os.makedirs(directory)
```

What do you think?

vdumoulin:
This is a useful addition. Would you mind writing a small unit test for this piece of code?

fuel/downloaders/base.py
janchorowski: This one has a race condition in it (see e.g. http://stackoverflow.com/questions/18973418/os-mkdirpath-returns-oserror-when-directory-does-not-exist). Chances are low that it will happen here, but still it is best to not write fragile code

janchorowski:
Sure, I guess it will be best to extract this into a utility function and test it as such.

fuel/downloaders/base.py
vdumoulin: Sounds good. In that case, maybe use `errno.EEXIST` like in the linked example?

vdumoulin:
LGTM. I'll wait for the tests to pass and I'll merge. Thanks a lot!

Issue #146, opened 2015-06-01
Cross validation support


dwf:
We should provide support for splitting a (finite) dataset into K folds and producing K pairs of training sets and validation sets from these folds.

You may want the folds to be stratified such that each class is represented in each fold with roughly the same proportion as in the wider training set. I have some code for this.

bartvm:
This is pretty relevant to this discussion: https://github.com/mila-udem/fuel/issues/101#issuecomment-99866739

dwf:
Mhmm. I think we could at least provide convenient functions to use for generating subsets, if not convenience wrappers that generate various DataStreams.

bartvm:
Yeah, agreed, my point was mostly that I think we shouldn't need to introduce any new abstractions, or change the behaviour of existing classes, to make this work. A helper function that generates a series of iteration schemes and/or data streams should be enough.

vdumoulin:
@ejls This is your assigned ticket for this week.

vdumoulin:
Fixed by #176.

PR #176, opened 2015-06-22
Add basic support for cross-validation (see #146)

ejls:


fuel/schemes.py
```diff
@@ -232,3 +232,29 @@ def get_request_iterator(self):
...
         self.rng.shuffle(indices)
         return iter_(indices)
+
+
+def k_fold_cross_validation(n, k):
+    """Return pairs of schemes to be used for cross-validation.
+
+    Parameters
+    ----------
+    n : int
```
vdumoulin: I'd replace `n` with a more explicit argument name, like `num_examples`.

vdumoulin:
I think we can generalize this a little more to support subclasses of `BatchScheme` and `IndexScheme`:

``` python
def k_fold_cross_validation(scheme_class, num_examples, k, **kwargs):
   # ...
       valid = scheme_class(xrange(begin, end), **kwargs)
```

vdumoulin:
In the perspective that this PR is work-in-progress, here's another suggestion: I don't think we should support having `num_examples` not be a multiple of `k` by default. Silently accepting that may lead to incorrect results without the user's knowledge.

I propose that we add an optional `strict` kwarg, defaulting to `True`, which controls the following:
* If `strict is True`, then `num_examples % k != 0` causes a `ValueError` to be raised.
* If `strict is False`, then the return value of the generator is a tuple with three elements: the `train` scheme, the `valid` scheme, and the number of examples covered by the `valid` scheme.

ejls:
I've updated the PR to address your remarks.

fuel/schemes.py
```diff
@@ -232,3 +232,48 @@ def get_request_iterator(self):
...
         return iter_(indices)
+
+
+def k_fold_cross_validation(scheme_class, num_examples, k, strict=True,
+                            **kwargs):
+    """Return pairs of schemes to be used for cross-validation.
+
+    Parameters
+    ----------
+    scheme_class : class
```
vdumoulin: I'd be more specific here and specify that it's either an `IndexScheme` instance or a `BatchScheme` instance.

fuel/schemes.py
```diff
@@ -232,3 +232,48 @@ def get_request_iterator(self):
...
+        The type of the returned schemes. The constructor is called with an
+        iterator and `**kwargs` as arguments.
+    num_examples : int
+        The number of examples in the datastream.
+    k : int
+        The number of folds to return.
+    strict : bool, optional
+        If `True`, enforce that `num_examples` is divisible by `k` and so,
```
vdumoulin: Can you also specify what happens when it's `False`? I know it's specified in the `Return` section, but I prefer it to be repeated than missed by users.

fuel/schemes.py
```diff
@@ -232,3 +232,48 @@ def get_request_iterator(self):
...
+    -------
+    folds : generator
+        The generator returns `k` tuples, for each tuple, the first element
+        is the training set, the second element is the validation set.
+        If `strict` is set to `False`, the tuple has a third element: the
+        size of the validation set.
+
+    """
+    if strict and num_examples % k != 0:
+        raise ValueError("Trying to cross validate a dataset which size is " +
```
vdumoulin: I'd try giving more information to the user. Here's a suggestion:

``` python
"{} examples are not divisible in {} evenly-sized ".format(num_examples, k) +
"folds. To allow this, have a look at the `strict` argument."
```

tests/test_schemes.py
```diff
@@ -103,3 +104,43 @@ def test_concatenated_scheme():
...
     assert (list(sch.get_request_iterator()) ==
             ([10] * 5) + ([20] * 3) + [30])
+
+
+def test_k_fold_cross_validation():
+    # test raise when strict=True
+    cross = k_fold_cross_validation(SequentialExampleScheme, 10, 3)
+    assert_raises(ValueError, next, cross)
+
+    # test BatchScheme when strict=False
```
vdumoulin: I think there's a typo: you're testing an `IndexScheme` here and a `BatchScheme` further down.

fuel/schemes.py
```diff
@@ -232,3 +232,48 @@ def get_request_iterator(self):
...
+        If `True`, enforce that `num_examples` is divisible by `k` and so,
+        that all validation sets have the same size. Defaults to `True`.
+
+    Returns
+    -------
+    folds : generator
+        The generator returns `k` tuples, for each tuple, the first element
+        is the training set, the second element is the validation set.
+        If `strict` is set to `False`, the tuple has a third element: the
+        size of the validation set.
```
vdumoulin: Phrasing suggestion:

```
The generator returns `k` tuples. The first two elements of the tuple are
the training and validation iteration schemes. If `strict` is set to `False`,
the tuple has a third element corresponding to the size of the validation set.
```

fuel/schemes.py
```diff
@@ -232,3 +232,48 @@ def get_request_iterator(self):
...
+        iterator and `**kwargs` as arguments.
+    num_examples : int
+        The number of examples in the datastream.
+    k : int
+        The number of folds to return.
+    strict : bool, optional
+        If `True`, enforce that `num_examples` is divisible by `k` and so,
+        that all validation sets have the same size. Defaults to `True`.
+
+    Returns
```
dwf: We've been using "Yields" in other parts of the code for this section, where we describe what the generator yields on individual iterations.

fuel/schemes.py
```diff
@@ -232,3 +232,48 @@ def get_request_iterator(self):
         indices = list(self.indices)
         self.rng.shuffle(indices)
         return iter_(indices)
+
+
+def k_fold_cross_validation(scheme_class, num_examples, k, strict=True,
```
dwf: I don't know if we need "k_fold" in the name; maybe just call it "cross_validation" and change "k" to "folds" or "num_folds"?

ejls:
Thank you both for your remarks, I've updated the PR.

vdumoulin:
Fixes #176. LGTM, merging.

PR #177, opened 2015-06-22
limiting the number of excursion in an iterator

mducoffe:


dwf:
Can you provide an example of where you might use this? It seems like most schemes provide a way to limit the number of returned batches/examples automatically based on their constructor arguments.

mducoffe:
When evaluating on the validation set I would like to use a random subset
of the minibatches using this iterator and ShuffledScheme. Like that I can
have a rough idea of the performance when trying the hyperparameters

2015-06-22 16:12 GMT-04:00 David Warde-Farley <notifications@github.com>:

dwf:
Okay. The name could probably be better too -- it's unclear in what way it would be "limited". Perhaps "TruncatedEpochScheme"? @rizar might have an opinion.

vdumoulin:
I'd be okay with `TruncatedEpochScheme`.

fuel/schemes.py
```diff
@@ -232,3 +232,17 @@ def get_request_iterator(self):
         indices = list(self.indices)
         self.rng.shuffle(indices)
         return iter_(indices)
+
+@add_metaclass(ABCMeta)
+class LimitedScheme(IterationScheme):
+    """Iterate over the sequence given an iterationScheme object
+        but in a limited number of excursions defined by times
+    """
+    def __init__(self, iteration_scheme, times, *args, **kwargs):
```
vdumoulin: You can safely remove `*args, **kwargs`. It's used in other iterators (e.g. `ShuffledExampleScheme`) to pass arbitrary args and kwargs to the superclass' `__init__` method, but in this case you're not calling any other `__init__` method.

fuel/schemes.py
```diff
@@ -232,3 +232,17 @@ def get_request_iterator(self):
...
+        self.iteration_scheme = iteration_scheme
+        self.times = times
+        assert self.times >=1, "Error : you need a positiv number of mini batches"
+
+    def get_request_iterator(self):
+        it = self.iteration_scheme.get_request_iterator()
+        return iter_([it.next() for n in range(self.times)])
```
vdumoulin: Is there a reason why you're coercing `it.next() for n in range(self.times)` into a list?

dwf: @vdumoulin You actually would have to do that for the result to be
picklable (generators are not). But you're better off just using
`picklable_itertools.islice`.

vdumoulin: TIL, then!

vdumoulin:
@mducoffe I'm doing a cleanup of the open PRs. Would you still like to go ahead with this one?

PR #179, opened 2015-06-23
Example schemes support for H5PYDataset

rizar:
This a starter implementation of support for examples schemes in H5PYDataset. I am interested in opinion, @vdumoulin , @dwf , should we have it?

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -186,11 +186,11 @@ class H5PYDataset(Dataset):
         self.sort_indices = sort_indices
 
         self._parse_dataset_info()
-        self.example_iteration_scheme = SequentialExampleScheme(self.num_examples)
+        self.example_iteration_scheme = SequentialExampleScheme(
+            self.num_examples)
 
         kwargs.setdefault('axis_labels', self.default_axis_labels)
         super(H5PYDataset, self).__init__(**kwargs)
-
 
     def _parse_dataset_info(self):
         """Parses information related to the HDF5 interface.
@@ -646,6 +646,8 @@ class H5PYDataset(Dataset):
                 if hasattr(request, 'step'):
                     req = slice(request.start + subset.start,
                                 request.stop + subset.start, request.step)
+                elif isinstance(request, numbers.Integral):
+                    req = subset.start + request
                 else:
                     req = [index + subset.start for index in request]
             else:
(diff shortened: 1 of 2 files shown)


vdumoulin:
Short and sweet. I had a quick look and it looks good to me. I'll wait for the tests to pass and I'll have another look then.

rizar:
Thanks for quick feedback! It still requires a test, but having your okay I will write it soon.

dwf:
You should really, really use `numbers.Integral` for things like this, so that it catches numpy integer dtypes.

rizar:
I am struggling to see how the test failures are related to the change I've made. @vdumoulin , can you please take a look?

vdumoulin:
The "ValueError not raised" test failure arises because `True` is recognized as an integral number by `numbers.Integral`, and this is a value that's not supposed to make sense in the context of a request. I'm still working on the doctest failure.

vdumoulin:
The doctest issue is really subtle: you're accessing `self.num_examples` in setting `self.example_iteration_scheme` *before* the superclass' `__init__` method has been called, which means that `sources` hasn't been set yet and the dataset is using the default `self.provides_sources`. This has the side effect of loading all sources in memory, which makes the doctest fail.

The error is fixed if you move that statement after the `super` call.

rizar:
@vdumoulin , thanks for your help!

@dwf , can we have somehow get rid of `bool` as a representative of `numbers.Integral`?

rizar:
@vdumoulin , I decided that in `test_value_error_out_of_memory_get_data` I can use something less ambiguous as `True` and now tests pass. I have to see however why coverage dropped.

rizar:
@vdumoulin , any idea why this PR breaks `tests.test_server.TestServer.test_pickling`?

dwf:
Sorry, didn't see your comment until now.

I don't really understand what you said about `bool` though.

rizar:
I tried to ask if there exists `numbers.IntegralButNotBool` which includes `int` and `numpy.int` and etc. but not `bool`.

dwf:
Oh. I see. I do not think so, no. :(

dwf:
Oh man, what the hell Python:

In [9]: isinstance(False, int)
Out[9]: True

ಠ_ಠ

rizar:
Okay, then I guess I was right that I changed the test to use `dict()` instead.

vdumoulin:
@rizar What's the status on this PR?

rizar:
Thanks for the reminder, I am going to finish it, say, this week.

rizar:
Outdated, replaced by #282.

PR #182, opened 2015-06-27
Force sources attrib of FilterSources to be a tuple

janchorowski:
AbstractDataStream requires that sources be a tuple. When sources are a list one get errors on concatenation with tuples, e.g. when a FilterSources transformer is chained with a Mapping one (there is then a list to tuple concatenation at https://github.com/mila-udem/fuel/blob/master/fuel/transformers/__init__.py#L107-L108 )

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -643,7 +643,7 @@ class FilterSources(Transformer):
                              "data_stream.sources")
 
         # keep order of data_stream.sources
-        self.sources = [s for s in data_stream.sources if s in sources]
+        self.sources = tuple([s for s in data_stream.sources if s in sources])
 
     def get_data(self, request=None):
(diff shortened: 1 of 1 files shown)


fuel/transformers/__init__.py
```diff
@@ -643,7 +643,7 @@ def __init__(self, data_stream, sources):
                              "data_stream.sources")
 
         # keep order of data_stream.sources
-        self.sources = [s for s in data_stream.sources if s in sources]
+        self.sources = tuple([s for s in data_stream.sources if s in sources])
```
rizar: No need for square brackets.

rizar:
LGTM after square brackers are removed.

janchorowski:
OK, done

rizar:
Thanks for the fix!

PR #185, opened 2015-06-30
Iris dataset

laurent-dinh:


docs/new_dataset.rst
```diff
@@ -235,7 +235,7 @@ Try downloading and converting the data file:
     cd $FUEL_DATA_PATH
     fuel-download iris
     fuel-convert iris
-    fuel-download --clear iris
+    fuel-download iris --clear
```
vdumoulin: Nice catch!

fuel/converters/iris.py
```diff
@@ -0,0 +1,71 @@
...
+        based on `dtype` will be used.
+    dtype : str, optional
+        Either 'float32', 'float64', or 'bool'. Defaults to `None`,
+        in which case images will be returned in their original
+        unsigned byte format.
+    Returns
+    -------
+    output_paths : tuple of str
+        Single-element tuple containing the path to the converted dataset.
+    """
```
vdumoulin: We're going to need a little breathing room in here ;). Could you make sure there's a blank line between paragraphs and the docstring ends with a blank line as well?

fuel/converters/iris.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def convert_iris(directory, output_directory, output_filename='iris.hdf5'):
+    """Converts the Iris dataset to HDF5.
+    Converts the Iris dataset to an HDF5 dataset compatible with
+    :class:`fuel.datasets.Iris`. The converted dataset is
+    saved as 'iris.hdf5'.
+    This method assumes the existence of the file `iris.data`.
+    It assumes the existence of the following file:
+    * `iris.data`
```
vdumoulin: There's duplicate information in lines 14-16.

fuel/converters/iris.py
```diff
@@ -0,0 +1,71 @@
...
+    """
+
+    output_path = os.path.join(output_directory, output_filename)
+    h5file = h5py.File(output_path, mode='w')
+    classes = {'Iris-setosa': 0, 'Iris-versicolor': 1, 'Iris-virginica': 2}
+    data = numpy.loadtxt(
+        os.path.join(directory, 'iris.data'),
+        converters={4: lambda x: classes[x]},
+        delimiter=',')
+    numpy.random.shuffle(data)
```
vdumoulin: In order to favour reproducibility, we should create an rng using `numpy.random.RandomState` with some user-defined seed here. The converter function should need an additional `seed=<whatever_you_wand>` kwarg which you'll use as the random seed.

fuel/converters/iris.py
```diff
@@ -0,0 +1,71 @@
...
+        Directory in which input files reside.
+    output_directory : str
+        Directory in which to save the converted dataset.
+    output_filename : str, optional
+        Name of the saved dataset. Defaults to `None`, in which case a name
+        based on `dtype` will be used.
+    dtype : str, optional
+        Either 'float32', 'float64', or 'bool'. Defaults to `None`,
+        in which case images will be returned in their original
+        unsigned byte format.
```
vdumoulin: I think the `dtype` description is a leftover from some copy-paste.

As a matter of fact, do we really need this much flexibility, or would it be okay to use `float64` and add `ForceFloatX` as a default transformer?

fuel/converters/iris.py
```diff
@@ -0,0 +1,71 @@
...
+        delimiter=',')
+    numpy.random.shuffle(data)
+    features = data[:, :-1].astype('float32')
+    targets = data[:, -1].astype('uint8').reshape((-1, 1))
+    train_features = features[:100]
+    train_targets = targets[:100]
+    valid_features = features[100:120]
+    valid_targets = targets[100:120]
+    test_features = features[120:]
+    test_targets = targets[120:]
```
vdumoulin: I don't think there's a canonical split for Iris, right? In that case, we should just make one `'all'` split and let users manually select their desired split with `subset` or through an iteration scheme.

fuel/datasets/iris.py
```diff
@@ -0,0 +1,29 @@
...
+       taxonomic problems*, Annual Eugenics, 7, Part II, 179-188,
+       September 1936.
+    .. [UCI] https://archive.ics.uci.edu/ml/datasets/Iris
+    Parameters
+    ----------
+    which_sets : tuple of str
+        Which split to load. Valid values are 'train' and 'test',
+        corresponding to the training set (50,000 examples) and the test
+        set (10,000 examples).
+    """
```
vdumoulin: This also needs blank lines between paragraphs and at the end of the docstring.

fuel/datasets/iris.py
```diff
@@ -0,0 +1,29 @@
...
+    .. [IRIS] Ronald A. Fisher, *The use of multiple measurements in
+       taxonomic problems*, Annual Eugenics, 7, Part II, 179-188,
+       September 1936.
+    .. [UCI] https://archive.ics.uci.edu/ml/datasets/Iris
+    Parameters
+    ----------
+    which_sets : tuple of str
+        Which split to load. Valid values are 'train' and 'test',
+        corresponding to the training set (50,000 examples) and the test
+        set (10,000 examples).
```
vdumoulin: Could you change the description to reflect my earlier comment on the `'all'` split?

fuel/downloaders/iris.py
```diff
@@ -0,0 +1,19 @@
...
+def fill_subparser(subparser):
+    """Sets up a subparser to download the Iris dataset file.
+    The Iris dataset file `iris.data` is downloaded from the UCI
+    Machine Learning Repository [UCI].
+    .. [UCI] https://archive.ics.uci.edu/ml/datasets/Iris
+    Parameters
+    ----------
+    subparser : :class:`argparse.ArgumentParser`
+        Subparser handling the `mnist` command.
+    """
```
vdumoulin: Same comment about blank lines.

fuel/downloaders/iris.py
```diff
@@ -0,0 +1,19 @@
...
+
+def fill_subparser(subparser):
+    """Sets up a subparser to download the Iris dataset file.
+    The Iris dataset file `iris.data` is downloaded from the UCI
+    Machine Learning Repository [UCI].
+    .. [UCI] https://archive.ics.uci.edu/ml/datasets/Iris
+    Parameters
+    ----------
+    subparser : :class:`argparse.ArgumentParser`
+        Subparser handling the `mnist` command.
```
vdumoulin: Typo: `mnist`.

fuel/downloaders/mnist.py
```diff
@@ -4,7 +4,7 @@
 def fill_subparser(subparser):
     """Sets up a subparser to download the MNIST dataset files.
 
-    The following MNIST dataset files are downoladed from Yann LeCun's
+    The following MNIST dataset files are downloaded from Yann LeCun's
```
vdumoulin: Thanks :).

vdumoulin:
Thank you for your diligence!

In addition to my inline comments, it would be a good thing to test the converter. Since this dataset is very small, I'd go the same route as for `caltech101_silhouettes` and explicitly download the data, do the conversion and compare the converted file with the original dataset (you can have a look at the test for caltech in `tests.test_converters` to get inspired).

fuel/converters/iris.py
```diff
@@ -0,0 +1,71 @@
...
+    h5file['targets'].dims[1].label = 'index'
+
+    h5file.flush()
+    h5file.close()
+
+    return (output_path,)
+
+
+def fill_subparser(subparser):
+    subparser.set_defaults(func=convert_iris)
```
vdumoulin: We should also add an option (e.g. `--seed`) with some default value in the subparser, which will be passed to the converter.

vdumoulin:
I was about to comment that you removed the random shuffling of the example, since examples are grouped by class in the raw file, but now I'm wondering if we should leave it like it is for the sake of simplicity.

Shuffling can also be achieved at the scheme level like so:

``` python
import numpy

indices = numpy.arange(150)
numpy.random.shuffle(indices)
train_indices, test_indices = indices[:100], indices[100:]
train_scheme = SequentialScheme(examples=train_indices, batch_size=10)
test_scheme = SequentialScheme(examples=test_indices, batch_size=10)
```

PR #188, opened 2015-07-07
Transformer interface overhaul

### LINKED ISSUE #134: Should all Transformers be passing **kwargs to superclass constructor?
A whole bunch of Transformers don't pass on extra arguments to their superclasses. It also strikes me as a little weird that they all end up with a `batch_input` attribute that in many cases is ineffectual.

vdumoulin:
Fixes #134.

Things left to do:
* Determine if it is okay to assume that data streams with no iteration scheme will produce examples.
* At the moment, transformers don't work on batches if they produce examples, and vice versa. We might want to bake in some default behaviour.
* Test coverage.

### DIFF
--- a/fuel/datasets/text.py
+++ b/fuel/datasets/text.py
@@ -49,7 +49,8 @@ class TextFile(Dataset):
     >>> text_data = TextFile(files=['sentences.txt'],
     ...                      dictionary=dictionary, bos_token=None,
     ...                      preprocess=lower)
-    >>> for data in text_data.get_example_stream().get_epoch_iterator():
+    >>> from fuel.streams import DataStream
+    >>> for data in DataStream(text_data).get_epoch_iterator():
     ...     print(data)
     ([2, 0, 3, 0, 1],)
     ([2, 0, 4, 1],)
--- a/fuel/datasets/text.py
+++ b/fuel/datasets/text.py
@@ -49,8 +49,7 @@ class TextFile(Dataset):
     >>> text_data = TextFile(files=['sentences.txt'],
     ...                      dictionary=dictionary, bos_token=None,
     ...                      preprocess=lower)
-    >>> from fuel.streams import DataStream
-    >>> for data in DataStream(text_data).get_epoch_iterator():
+    >>> for data in text_data.get_example_stream().get_epoch_iterator():
     ...     print(data)
     ([2, 0, 3, 0, 1],)
     ([2, 0, 4, 1],)
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -36,6 +36,15 @@ class IterationScheme(object):
     def get_request_iterator(self):
         """Returns an iterator type."""
 
+    @property
+    def produces_batches(self):
+        """Whether requests produced by this scheme are batches."""
+        if not hasattr(self, '_produces_batches'):
+            raise AttributeError("the {} scheme ".format(self.__class__) +
+                                 "can't tell what type of requests it " +
+                                 "produces (examples or batches)")
+        return self._produces_batches
+
 
 @add_metaclass(ABCMeta)
 class BatchSizeScheme(IterationScheme):
@@ -47,6 +56,7 @@ class BatchSizeScheme(IterationScheme):
     that only provide the number of examples that should be in a batch.
 
     """
+    _produces_batches = True
 
 
 @add_metaclass(ABCMeta)
@@ -75,6 +85,8 @@ class BatchScheme(IterationScheme):
         multiple of `batch_size`.
 
     """
+    _produces_batches = True
+
     def __init__(self, examples, batch_size):
         if isinstance(examples, Iterable):
             self.indices = examples
@@ -111,6 +123,8 @@ class IndexScheme(IterationScheme):
     but where we want to return single examples instead of batches.
 
     """
+    _produces_batches = False
+
     def __init__(self, examples):
         if isinstance(examples, Iterable):
             self.indices = examples
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -111,8 +111,7 @@ class ConcatenatedScheme(IterationScheme):
 
     """
     def __init__(self, schemes):
-        if not all(scheme.requests_examples == schemes[0].requests_examples
-                   for scheme in schemes):
+        if not len(set(scheme.requests_examples for scheme in schemes)) == 1:
             raise ValueError('all schemes must produce the same type of '
                              'requests (batches or examples)')
         self.schemes = schemes
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -18,6 +18,12 @@ class IterationScheme(object):
     sequential batches, shuffled batches, etc. for datasets that choose to
     support them.
 
+    Attributes
+    ----------
+    requests_examples : bool
+        Whether requests produced by this scheme correspond to single
+        examples (as opposed to batches).
+
     Notes
     -----
     Iteration schemes implement the :meth:`get_request_iterator` method,
@@ -30,12 +36,6 @@ class IterationScheme(object):
 
     .. _iterator protocol:
        https://docs.python.org/3.3/library/stdtypes.html#iterator-types
-
-    Attributes
-    ----------
-    requests_examples : bool
-        Whether requests produced by this scheme correspond to single
-        examples (as opposed to batches).
 
     """
     @abstractmethod
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -33,8 +33,9 @@ class IterationScheme(object):
 
     Attributes
     ----------
-    produces_batches : bool
-        Whether requests produced by this scheme are batches.
+    requests_examples : bool
+        Whether requests produced by this scheme correspond to single
+        examples (as opposed to batches).
 
     """
     @abstractmethod
@@ -52,7 +53,7 @@ class BatchSizeScheme(IterationScheme):
     that only provide the number of examples that should be in a batch.
 
     """
-    produces_batches = True
+    requests_examples = False
 
 
 @add_metaclass(ABCMeta)
@@ -81,7 +82,7 @@ class BatchScheme(IterationScheme):
         multiple of `batch_size`.
 
     """
-    produces_batches = True
+    requests_examples = False
 
     def __init__(self, examples, batch_size):
         if isinstance(examples, Iterable):
@@ -110,7 +111,7 @@ class ConcatenatedScheme(IterationScheme):
 
     """
     def __init__(self, schemes):
-        if not all(scheme.produces_batches == schemes[0].produces_batches
+        if not all(scheme.requests_examples == schemes[0].requests_examples
                    for scheme in schemes):
             raise ValueError('all schemes must produce the same type of '
                              'requests (batches or examples)')
@@ -120,8 +121,8 @@ class ConcatenatedScheme(IterationScheme):
         return chain(*[sch.get_request_iterator() for sch in self.schemes])
 
     @property
-    def produces_batches(self):
-        return self.schemes[0].produces_batches
+    def requests_examples(self):
+        return self.schemes[0].requests_examples
 
 
 @add_metaclass(ABCMeta)
@@ -132,7 +133,7 @@ class IndexScheme(IterationScheme):
     but where we want to return single examples instead of batches.
 
     """
-    produces_batches = False
+    requests_examples = True
 
     def __init__(self, examples):
         if isinstance(examples, Iterable):
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -106,8 +106,8 @@ class ConcatenatedScheme(IterationScheme):
 
     Notes
     -----
-    All schemes must produce the same type of requests (batches or
-    examples).
+    All schemes being concatenated must produce the same type of
+    requests (batches or examples).
 
     """
     def __init__(self, schemes):
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -31,19 +31,15 @@ class IterationScheme(object):
     .. _iterator protocol:
        https://docs.python.org/3.3/library/stdtypes.html#iterator-types
 
+    Attributes
+    ----------
+    produces_batches : bool
+        Whether requests produced by this scheme are batches.
+
     """
     @abstractmethod
     def get_request_iterator(self):
         """Returns an iterator type."""
-
-    @property
-    def produces_batches(self):
-        """Whether requests produced by this scheme are batches."""
-        if not hasattr(self, '_produces_batches'):
-            raise AttributeError("the {} scheme ".format(self.__class__) +
-                                 "can't tell what type of requests it " +
-                                 "produces (examples or batches)")
-        return self._produces_batches
 
 
 @add_metaclass(ABCMeta)
@@ -56,7 +52,7 @@ class BatchSizeScheme(IterationScheme):
     that only provide the number of examples that should be in a batch.
 
     """
-    _produces_batches = True
+    produces_batches = True
 
 
 @add_metaclass(ABCMeta)
@@ -85,7 +81,7 @@ class BatchScheme(IterationScheme):
         multiple of `batch_size`.
 
     """
-    _produces_batches = True
+    produces_batches = True
 
     def __init__(self, examples, batch_size):
         if isinstance(examples, Iterable):
@@ -107,12 +103,25 @@ class ConcatenatedScheme(IterationScheme):
         A list of :class:`IterationSchemes`, whose request iterators
         are to be concatenated in the order given.
 
+    Notes
+    -----
+    All schemes must produce the same type of requests (batches or
+    examples).
+
     """
     def __init__(self, schemes):
+        if not all(scheme.produces_batches == schemes[0].produces_batches
+                   for scheme in schemes):
+            raise ValueError('all schemes must produce the same type of '
+                             'requests (batches or examples)')
         self.schemes = schemes
 
     def get_request_iterator(self):
         return chain(*[sch.get_request_iterator() for sch in self.schemes])
+
+    @property
+    def produces_batches(self):
+        return self.schemes[0].produces_batches
 
 
 @add_metaclass(ABCMeta)
@@ -123,7 +132,7 @@ class IndexScheme(IterationScheme):
     but where we want to return single examples instead of batches.
 
     """
-    _produces_batches = False
+    produces_batches = False
 
     def __init__(self, examples):
         if isinstance(examples, Iterable):
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -42,11 +42,8 @@ class AbstractDataStream(object):
         of examples).
 
     """
-    def __init__(self, iteration_scheme=None, produces_examples=None,
-                 axis_labels=None):
+    def __init__(self, iteration_scheme=None, axis_labels=None):
         self.iteration_scheme = iteration_scheme
-        if produces_examples is not None:
-            self.produces_examples = produces_examples
         self.axis_labels = axis_labels
 
     @property
@@ -131,6 +128,10 @@ class DataStream(AbstractDataStream):
         if dataset.axis_labels:
             kwargs.setdefault('axis_labels', dataset.axis_labels.copy())
         super(DataStream, self).__init__(**kwargs)
+        # A DataStream with no iteration scheme is considered an example stream
+        # by default
+        if not self.iteration_scheme:
+            self.produces_examples = True
         # If the data stream produces examples, remove 'batch' from axis labels
         if self.produces_examples and self.axis_labels:
             for source, labels in iteritems(self.axis_labels):
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -1,7 +1,7 @@
 from abc import ABCMeta, abstractmethod
 
 import zmq
-from six import add_metaclass
+from six import add_metaclass, iteritems
 
 from fuel.iterator import DataIterator
 from fuel.server import recv_arrays
@@ -42,8 +42,11 @@ class AbstractDataStream(object):
         of examples).
 
     """
-    def __init__(self, iteration_scheme=None, axis_labels=None):
+    def __init__(self, iteration_scheme=None, produces_examples=None,
+                 axis_labels=None):
         self.iteration_scheme = iteration_scheme
+        if produces_examples is not None:
+            self.produces_examples = produces_examples
         self.axis_labels = axis_labels
 
     @property
@@ -125,8 +128,14 @@ class DataStream(AbstractDataStream):
 
     """
     def __init__(self, dataset, **kwargs):
-        kwargs.setdefault('axis_labels', dataset.axis_labels)
+        if dataset.axis_labels:
+            kwargs.setdefault('axis_labels', dataset.axis_labels.copy())
         super(DataStream, self).__init__(**kwargs)
+        # If the data stream produces examples, remove 'batch' from axis labels
+        if self.produces_examples and self.axis_labels:
+            for source, labels in iteritems(self.axis_labels):
+                self.axis_labels[source] = tuple(
+                    label for label in labels if label != 'batch')
         self.dataset = dataset
         self.data_state = self.dataset.open()
         self._fresh_state = True
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -47,6 +47,9 @@ class AbstractDataStream(object):
         self.axis_labels = axis_labels
         if self.iteration_scheme:
             self.produces_examples = iteration_scheme.requests_examples
+        else:
+            # TODO: is this a valid assumption?
+            self.produces_examples = True
 
     def get_data(self, request=None):
         """Request data from the dataset or the wrapped stream.
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -37,11 +37,16 @@ class AbstractDataStream(object):
     sources : tuple of strings
         The names of the data sources returned by this data stream, as
         given by the dataset.
+    produces_examples : bool
+        Whether this data stream produces examples (as opposed to batches
+        of examples).
 
     """
     def __init__(self, iteration_scheme=None, axis_labels=None):
         self.iteration_scheme = iteration_scheme
         self.axis_labels = axis_labels
+        if self.iteration_scheme:
+            self.produces_examples = iteration_scheme.requests_examples
 
     def get_data(self, request=None):
         """Request data from the dataset or the wrapped stream.
@@ -150,6 +155,11 @@ class ServerDataStream(AbstractDataStream):
 
     Parameters
     ----------
+    sources : tuple of strings
+        The names of the data sources returned by this data stream.
+    produces_examples : bool
+        Whether this data stream produces examples (as opposed to batches
+        of examples).
     host : str, optional
         The host to connect to. Defaults to ``localhost``.
     port : int, optional
@@ -163,11 +173,16 @@ class ServerDataStream(AbstractDataStream):
         tell how many batches will actually be queued with a particular
         HWM. Defaults to 10. Be sure to set the corresponding HWM on the
         server's end as well.
+    axis_labels : dict, optional
+        Maps source names to tuples of strings describing axis semantics,
+        one per axis. Defaults to `None`, i.e. no information is available.
 
     """
-    def __init__(self, sources, host='localhost', port=5557, hwm=10):
-        super(ServerDataStream, self).__init__()
+    def __init__(self, sources, produces_examples, host='localhost', port=5557,
+                 hwm=10, axis_labels=None):
+        super(ServerDataStream, self).__init__(axis_labels=axis_labels)
         self.sources = sources
+        self.produces_examples = produces_examples
         self.host = host
         self.port = port
         self.hwm = hwm
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -67,6 +67,7 @@ class AbstractDataStream(object):
                                  self.iteration_scheme))
         self._produces_examples = value
 
+    @abstractmethod
     def get_data(self, request=None):
         """Request data from the dataset or the wrapped stream.
 
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -45,11 +45,27 @@ class AbstractDataStream(object):
     def __init__(self, iteration_scheme=None, axis_labels=None):
         self.iteration_scheme = iteration_scheme
         self.axis_labels = axis_labels
+
+    @property
+    def produces_examples(self):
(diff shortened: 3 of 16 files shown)


vdumoulin:
@rizar @dwf I'd like to conclude this saga before Friday, July 17, as this is the last day I'll be available to work on that until the summer school and I'll be presenting Fuel at the summer school.

Do you think it would be possible for you to review this by then?

dwf:
Once the checks pass I can review it pretty much immediately.

dwf:
Okay, tests pass, but coverage has dropped noticeably, and there are a few Scrutinizer errors. I guess I'll start reviewing.

fuel/schemes.py
```diff
@@ -95,13 +104,26 @@ class ConcatenatedScheme(IterationScheme):
         A list of :class:`IterationSchemes`, whose request iterators
         are to be concatenated in the order given.
 
+    Notes
+    -----
+    All schemes must produce the same type of requests (batches or
```
dwf: I think a better way of phrasing it is that "Each scheme must produce either batches or examples, but not a mix of both."

dwf: Nevermind, I didn't actually understand what this said.

fuel/schemes.py
```diff
@@ -95,13 +104,26 @@ class ConcatenatedScheme(IterationScheme):
...
+    examples).
+
     """
     def __init__(self, schemes):
+        if not all(scheme.requests_examples == schemes[0].requests_examples
```
dwf: Shorter: `len(set(s.requests_examples for s in schemes)) == 1`

fuel/streams.py
```diff
@@ -37,11 +37,19 @@ class AbstractDataStream(object):
...
+        of examples).
 
     """
     def __init__(self, iteration_scheme=None, axis_labels=None):
         self.iteration_scheme = iteration_scheme
         self.axis_labels = axis_labels
+        if self.iteration_scheme:
+            self.produces_examples = iteration_scheme.requests_examples
+        else:
+            # TODO: is this a valid assumption?
```
dwf: Definitely not a valid assumption. But I think it is fair to punt on this. Here is what I suggest (@rizar might have a different opinion):

```python
@property
def produces_examples(self):
    if self.iteration_scheme: 
        return self.iteration_scheme.requests_examples
    elif not hasattr(self, '_produces_examples'):
        raise ValueError("cannot infer type of stream for {} instance; set the "
                                   "produces_examples attribute to True (for example streams "
                                   "or False (for batch streams).".format(self.__class__.__name__))
    else:
        return self._produces_examples

@produces_examples.setter
def produces_examples(self, value):
    if self.iteration_scheme:
        raise ValueError("cannot set produces_examples on {} instance; "
                                   "determined by iteration scheme {}".format(
                                       self.__class__.__name__,
                                       self.iteration_scheme))
    self._produces_examples = value

File: fuel/streams.py [hunk 4]
```diff
@@ -150,6 +158,11 @@ class ServerDataStream(AbstractDataStream):
 
     Parameters
     ----------
+    sources : tuple of strings
+        The names of the data sources returned by this data stream.
+    produces_examples : bool
+        Whether this data stream produces examples (as opposed to batches
```
dwf: Ideally this should eventually be auto-negotiated. Please remember to file a ticket.

fuel/streams.py
```diff
@@ -163,11 +176,16 @@ class ServerDataStream(AbstractDataStream):
         tell how many batches will actually be queued with a particular
         HWM. Defaults to 10. Be sure to set the corresponding HWM on the
         server's end as well.
+    axis_labels : dict, optional
+        Maps source names to tuples of strings describing axis semantics,
+        one per axis. Defaults to `None`, i.e. no information is available.
```
dwf: Ditto.

fuel/schemes.py
vdumoulin: Is it clear enough as it is?

fuel/streams.py
vdumoulin: This is something @lamblin also mentioned to me. I just filed the ticket for that.

fuel/streams.py
vdumoulin: I like that.

vdumoulin: See 2970001 for an overview of what changes it entails.

The biggest change from the users' point of view is that they'll have to get used to call `IterableDataset(...).get_example_stream()` instead of `DataStream(IterableDataset(...))`, because the latter would force them to add an additional `stream.produces_example = True` statement afterwards. I think it's a good trade-off.

vdumoulin:
@dwf thank you for your time, I did not expect you to review this so quickly!

I'll look into bringing the coverage back up tomorrow.

dwf:
Not quite done with the review but figured I'd give you what feedback I
could asap. I'll try and finish tonight.

fuel/schemes.py
dwf: How about "All schemes being concatenated"

fuel/streams.py
dwf: I think we might be able to make that assumption for `DataStream`, actually, if it doesn't have an iteration scheme, but AbstractDataStream should make no assumptions I don't think, since that'll probably be what other kinds of streams inherit from (honestly, the more I think about it the more I think that `iteration_scheme` actually belongs in `DataStream` rather than `AbstractDataStream`, but let's leave that for now) -- @rizar or @bartvm might have comments on that too).

dwf: So I guess in concrete terms that would just mean setting `self.produces_examples = True` if there's no iteration scheme in the `DataStream` constructor.

dwf: Okay, going to look at the rest of this PR. :stuck_out_tongue:

fuel/transformers/__init__.py
```diff
@@ -8,12 +8,19 @@
...
 
 @add_metaclass(ABCMeta)
 class Transformer(AbstractDataStream):
     """A data stream that wraps another data stream.
 
+    Subclasses must define a `transform_batch` method (to act on batches),
+    a `transform_example` method (to act on individual examples), or
+    both methods. If a `transform_any` method is defined, it is used
+    for both example and batch inputs. This is useful if the example
+    and batch version of a transformation are the same.
```
dwf: Maybe say something about how if you are going from batches to examples or examples to batches, you should override `get_data` instead.

fuel/transformers/__init__.py
```diff
@@ -344,12 +431,20 @@ class Batch(Transformer):
...
-    def __init__(self, data_stream, iteration_scheme, strictness=0):
+    def __init__(self, data_stream, iteration_scheme, strictness=0, **kwargs):
+        if not data_stream.produces_examples:
+            raise ValueError('the wrapped data stream must produce examples, '
+                             'not batches of examples.')
+        if data_stream.axis_labels:
+            kwargs.setdefault(
+                'axis_labels',
+                dict([(source, ('batch',) + labels if labels else None) for
+                      source, labels in iteritems(data_stream.axis_labels)]))
```
dwf: Here you should check for `produces_examples` in kwargs, raise an appropriate error if it's there, and pass `produces_examples=False` to super (since it would be weird if the `Batch` transformer didn't always produce batches...)

dwf: In fact, forget what I said about raising an error: there's no need, since you'll get a complaint about duplicate values for keyword anyway. But you should have `produces_examples=False` here.

fuel/transformers/__init__.py
```diff
@@ -65,21 +73,30 @@ def get_epoch_iterator(self, **kwargs):
         return super(Transformer, self).get_epoch_iterator(**kwargs)
 
     def get_data(self, request=None):
```
dwf: I like the `transform_example`, `transform_batch` setup. Could you write short docstrings for each of those (I don't think we really need a Parameters section, even).

What I kind of don't like is how `transform_any` is effectively a "hidden" part of the protocol because there's no base class method (though it is mentioned in the docs). Though it's kind of ugly, it might be better to throw catch a specific class of Exception (like `class NeedsStreamType(NotImplementedError): pass`), catch it in this method, and then dispatch one or the other of `transform_batch` or `transform_example`.

We could also write a really simple `StreamAgnosticTransformer` (or even just `AgnosticTransformer`) class that has an abstract `transform_any` and then `transform_batch` and `transform_example` implementations that call it, and direct users to subclass that if their transformation for batches and examples is the same.

dwf: (With the current logic, such an `AgnosticTransformer` would still require examples in, examples out or batches in, batches out, but I think that's fine.)

fuel/transformers/__init__.py
vdumoulin: Will do.

fuel/transformers/__init__.py
vdumoulin: Not sure I agree: the `AbstractDataStream` setter for `produces_examples` will complain because there's an iteration scheme for this stream.

I think we should check that the iteration scheme requests batches, though.

dwf: Okay, right, I forgot that we need to account for the iteration scheme.

dwf: Probably good to write a comment to the effect that it gets its `produces_examples` value from `self.iteration_scheme`, and that we expect it to always be true here.

fuel/transformers/__init__.py
```diff
@@ -97,8 +114,9 @@ class Mapping(Transformer):
         data under source names `add_sources`.
 
     """
-    def __init__(self, data_stream, mapping, add_sources=None):
-        super(Mapping, self).__init__(data_stream)
+    def __init__(self, data_stream, produces_examples, mapping,
```
dwf: It seems like a pretty fringe use case to be having a `Mapping` that has a different value of `produces_examples` than its data stream, don't you think? What about getting rid of the constructor argument and just doing

```python
@property
def produces_examples(self):
    return self.data_stream.produces_examples
```

(if you go this route, here it would be good to double check that it's not passed as kwargs, as it would have no effect)

fuel/transformers/__init__.py
```diff
@@ -118,8 +136,15 @@ def get_data(self, request=None):
 
 
 @add_metaclass(ABCMeta)
-class SingleMapping(Transformer):
-    """Applies a single mapping to multiple sources.
+class SourcewiseTransformer(Transformer):
```
dwf: Similar comment about this class and `AgnosticTransformer` etc. I don't know, I'm interested in hearing what you think of that proposal first.

dwf: I would say that `transform_any_source` kinda breaks symmetry with `transform_source_batch` and `transform_source_example`.

fuel/transformers/__init__.py
```diff
@@ -441,41 +547,38 @@ def sources(self):
...
             for i, sequence_length in enumerate(lengths):
                 mask[i, :sequence_length] = 1
-            data_with_masks.append(mask)
-        return tuple(data_with_masks)
+            batch_with_masks.append(mask)
+        return tuple(batch_with_masks)
 
 
-class Merge(Transformer):
+class Merge(AbstractDataStream):
```
dwf: Why is this now an ADS?

fuel/transformers/__init__.py
```diff
@@ -499,13 +602,34 @@ class Merge(Transformer):
...
+                             'output (batches or examples)')
         self.data_streams = data_streams
+        self.produces_examples = self.data_streams[0].produces_examples
+
         if len(list(chain(*[data_stream.sources for data_stream
                             in data_streams]))) != len(sources):
             raise ValueError("wrong number of sources given")
         self.sources = sources
 
+    def close(self):
```
dwf: Hmm, why are these suddenly necessary?

fuel/transformers/__init__.py
```diff
@@ -636,19 +765,21 @@ class FilterSources(Transformer):
...
     """
-    def __init__(self, data_stream, sources):
-        super(FilterSources, self).__init__(data_stream)
+    def __init__(self, data_stream, sources, **kwargs):
         if any(source not in data_stream.sources for source in sources):
             raise ValueError("sources must all be contained in "
                              "data_stream.sources")
+        if data_stream.axis_labels:
+            kwargs.setdefault('axis_labels',
+                              dict([(source, labels) for (source, labels)
```
dwf: No need for square brackets here

dwf:
I think I've more or less now read the entire PR. Sorry it took me so long, it's a big one.

fuel/transformers/__init__.py
vdumoulin: The `close`, `reset` and `next_epoch` methods defined in `Transformer` assume that the transformer wraps only one data stream, so I would need to override those. Plus, `Merge` already overrides `get_epoch_iterator`, which pretty much removes all the benefits from inheriting from `Transformer`. The only advantage of doing so would be to keep an *is a* relationship with `Transformer`, and even then, it's not 100% clear that merging data streams should be counted as a transformation.

What do you think?

fuel/transformers/__init__.py
vdumoulin: See my reply about ADS. `Merge` has a list of data streams, so the default implementation in `Transformer` would crash.

fuel/transformers/__init__.py
vdumoulin: I like your idea of having an `AgnosticTransformer`. I'll implement it.

fuel/transformers/__init__.py
vdumoulin: Will do.

fuel/transformers/__init__.py
vdumoulin: This also makes sense. I'll implement it.

fuel/transformers/__init__.py
vdumoulin: I think what you're suggesting makes sense.

vdumoulin: Although we could simply call the superclass' constructor by passing `self.data_stream.produces_examples`. An error will be raised if someone tries to pass `produces_examples` as a kwarg.

dwf: Actually, we don't even need to do that (and shouldn't) because the
overrided property's non-existent setter will get called. So that will
catch it if someone tries to pass `produces_examples`.

vdumoulin:
@dwf I think I addressed all your comments.

You and I have a different definition of "a long time", because I'm pretty satisfied with the review time.

fuel/transformers/__init__.py
```diff
@@ -8,12 +8,24 @@
...
+from fuel.schemes import BatchSizeScheme
 
 
 @add_metaclass(ABCMeta)
 class Transformer(AbstractDataStream):
     """A data stream that wraps another data stream.
 
+    Subclasses must define a `transform_batch` method (to act on batches),
+    a `transform_example` method (to act on individual examples), or
+    both methods. If a `transform_any` method is defined, it is used
```
dwf: I think this doc is out of date now

fuel/transformers/__init__.py
```diff
@@ -130,45 +183,112 @@ class SingleMapping(Transformer):
...
 
-class Flatten(SingleMapping):
-    """Flattens selected sources along all but the first axis."""
+    Incoming sources will be treated as numpy arrays (i.e. using
+    `numpy.asarray`).
+
+    """
     def __init__(self, data_stream, **kwargs):
-        super(Flatten, self).__init__(data_stream, **kwargs)
+        super(Flatten, self).__init__(
```
dwf: It would be very nice to automatically infer the new axis labels here, since it's a case where we can.

fuel/transformers/__init__.py
```diff
@@ -130,45 +183,112 @@ class SingleMapping(Transformer):
...
+    def transform_source_batch(self, source_batch):
+        return numpy.asarray(source_batch).reshape((source_batch.shape[0], -1))
 
 
-class ScaleAndShift(SingleMapping):
+class ScaleAndShift(AgnosticSourcewiseTransformer):
     """Scales and shifts selected sources by scalar quantities.
 
+    Incoming sources will be treated as numpy arrays (i.e. using
+    `numpy.asarray`).
```
dwf: Does this cause any monkey business with dtypes if it's a batch of e.g. float32 arrays in a list?

fuel/transformers/__init__.py
```diff
@@ -180,15 +300,20 @@ class ScaleAndShift(SingleMapping):
     def __init__(self, data_stream, scale, shift, **kwargs):
         self.scale = scale
         self.shift = shift
-        super(ScaleAndShift, self).__init__(data_stream, **kwargs)
+        kwargs.setdefault('axis_labels', data_stream.axis_labels)
```
dwf: Is there a case where we want `axis_labels` to be overrideable?

The only case where I can think of that we would want to override is if scale and shift are arrays, and there could be some broadcasting going on. But right now the assumption is that they are scalars...

fuel/transformers/__init__.py
```diff
@@ -200,31 +325,28 @@ def __init__(self, data_stream, dtype, **kwargs):
         if dtype == 'floatX':
             dtype = config.floatX
         self.dtype = dtype
-        super(Cast, self).__init__(data_stream, **kwargs)
+        kwargs.setdefault('axis_labels', data_stream.axis_labels)
```
dwf: Same question. Seems like an even clearer case where we can force `axis_labels` in the arg list below. I guess we can   argue "caveat emptor" or something but allowing the constructor to do something nonsensical doesn't seem worthwhile.

fuel/transformers/__init__.py
```diff
@@ -200,31 +325,28 @@ def __init__(self, data_stream, dtype, **kwargs):
...
+    def transform_any_source(self, source_data):
+        return numpy.asarray(source_data, dtype=self.dtype)
 
 
-class ForceFloatX(Transformer):
+class ForceFloatX(AgnosticSourcewiseTransformer):
     """Force all floating point numpy arrays to be floatX."""
-    def __init__(self, data_stream):
+    def __init__(self, data_stream, **kwargs):
+        kwargs.setdefault('axis_labels', data_stream.axis_labels)
```
dwf: Same question.

fuel/transformers/__init__.py
```diff
@@ -238,8 +360,10 @@ class Filter(Transformer):
         Should return ``True`` for the samples to be kept.
 
     """
-    def __init__(self, data_stream, predicate):
-        super(Filter, self).__init__(data_stream)
+    def __init__(self, data_stream, predicate, **kwargs):
+        kwargs.setdefault('axis_labels', data_stream.axis_labels)
```
dwf: This one's less clear to me, I think setdefault is fine.

fuel/transformers/__init__.py
```diff
@@ -271,12 +395,17 @@ class Cache(Transformer):
         refilled when needed through the :meth:`get_data` method.
 
     """
-    def __init__(self, data_stream, iteration_scheme):
+    def __init__(self, data_stream, iteration_scheme, **kwargs):
```
dwf: No mention of axis labels here... I think it can be forced in the call to super().

Also, not sure how produces_examples should be handled, as I have no idea how this is used. Does it only emit example streams?

fuel/transformers/__init__.py
dwf: I think in cases of doubt like this one it's fine to leave it setdefault.

fuel/transformers/__init__.py
vdumoulin: I think the farthest we can go here is to set the axis label of the flattened axis to `None` or `''`, but other than that I agree.

fuel/transformers/__init__.py
dwf: (Also note here that they can still manually set the `axis_labels` attribute if they really know what they're doing, but I think it's good if the constructor forces something that's sane 99% of the time)

fuel/transformers/__init__.py
dwf: What do we typically call anonymous features in tabular datasets?
"features", I thought?

vdumoulin: I vaguely remember @bartvm advocating for an empty string instead of `'features'`, but I don't remember why.

I'm okay with `'features'`, personally.

vdumoulin: This made me realize one thing: `DataStream`'s `axis_label` attribute defaults to `dataset.axis_labels`, no matter what the iteration scheme is. This means that a data stream initialized with an examplewise iteration scheme will potentially have axis labels containing `'batch'`, even though there's no batch axis in the data.

I don't know exactly how to handle that, though.

vdumoulin: Would it be okay to assume that we can simply remove `'batch'` from the axis labels when a dataset is asked individual examples?

lamblin: The discussion about axis labels started [here](https://github.com/mila-udem/fuel/pull/78#discussion_r28370727).
`'features'` does not really add much over `''`, but I don't feel strongly one way or another.
Regarding `'batch'`, I think we can simply remove whatever axis was named `'batch'` when returning examples. Conversely, a Transformer grouping together examples could automatically add a `'batch'` axis.

dwf: I think so. But I'd maybe be conservative about it. Whatever that means...

fuel/transformers/__init__.py
```diff
@@ -344,12 +473,26 @@ class Batch(Transformer):
...
+            raise ValueError('the wrapped data stream must produce examples, '
+                             'not batches of examples.')
+        # The value for `produces_examples` is inferred from the iteration
+        # scheme's `requests_examples` attribute. We expect the scheme to
+        # request batches.
+        if iteration_scheme.requests_examples:
+            raise ValueError('the iteration scheme must request batches, '
+                             'not individual examples.')
+        if data_stream.axis_labels:
+            kwargs.setdefault(
```
dwf: Same question about whether this should be overrideable. I guess it's a question of flexibility versus sanity checking. Screw it, let's leave it as is.

fuel/transformers/__init__.py
dwf: SGTM to leave as is.

fuel/transformers/__init__.py
dwf: K nevermind.

fuel/transformers/text.py
```diff
@@ -24,16 +24,23 @@ class NGrams(Transformer):
         this source is 'targets'.
 
     """
-    def __init__(self, ngram_order, data_stream, target_source='targets'):
+    def __init__(self, ngram_order, data_stream, target_source='targets',
```
dwf: Axis labels here?

fuel/transformers/__init__.py
vdumoulin: It's requiring an instance of `BatchSizeScheme` as its iteration scheme, so it means `produces_examples=False` is implicitly forced through that restriction. I'll add a comment making this explicit.

fuel/transformers/__init__.py
vdumoulin: I don't know, to be honest. I can try to verify that.

fuel/transformers/text.py
vdumoulin: I'm not familiar with this transformer. I'd be more comfortable leaving this as it is and open a ticket.

fuel/transformers/__init__.py
vdumoulin: You're right about axis labels, we can safely assume they won't change, as the effect of `Cache` is to change the batch size of a data stream that produces batches.

fuel/transformers/text.py
dwf: k

vdumoulin: Ticket opened as #193 .

fuel/transformers/__init__.py
vdumoulin: I made the following test:

``` python
>>> import numpy
>>> batch = [numpy.ones(16, dtype='float32') for i in range(32)]
>>> print(numpy.asarray(batch).dtype)
float32
```

vdumoulin:
@dwf I went ahead and implemented the logic for default axis labels everywhere I could (which covers all classes you pointed out). Since at the moment there isn't anything enforced with axis labels and they're mostly there as documentation tools, I think we can let users override these default values without too much risk.

vdumoulin:
Tests are failing in Blocks because `DataStream` doesn't really allow passing no iteration scheme while `produces_examples` isn't set. It usually happens with `IndexableDataset` instances, and the workaround is to call `dataset.get_example_stream()` instead. I'll make a PR on the Blocks side of things to change that.

dwf:
What if we defaulted DataStream (but not AbstractDataStream) to be an
example stream, like I proposed? I don't know if you replied to that.

vdumoulin:
Right, I completely forgot about that. It must have gotten buried somewhere in the comments on outdated commits.

I think it's a good idea.

vdumoulin:
@dwf The tests pass, woohoo!

fuel/streams.py
```diff
@@ -101,8 +125,18 @@ class DataStream(AbstractDataStream):
 
     """
     def __init__(self, dataset, **kwargs):
-        kwargs.setdefault('axis_labels', dataset.axis_labels)
+        if dataset.axis_labels:
+            kwargs.setdefault('axis_labels', dataset.axis_labels.copy())
         super(DataStream, self).__init__(**kwargs)
+        # A DataStream with no iteration scheme is considered an example stream
+        # by default
+        if not self.iteration_scheme:
```
dwf: Is there ever a case where it would make sense for a `DataStream` to have `produces_examples` set to `False` as a kwarg?

fuel/transformers/__init__.py
dwf: Just note to self - already fixed.

fuel/datasets/text.py
```diff
@@ -49,8 +49,7 @@ class TextFile(Dataset):
     >>> text_data = TextFile(files=['sentences.txt'],
     ...                      dictionary=dictionary, bos_token=None,
     ...                      preprocess=lower)
-    >>> from fuel.streams import DataStream
-    >>> for data in DataStream(text_data).get_epoch_iterator():
+    >>> for data in text_data.get_example_stream().get_epoch_iterator():
```
dwf: Remove this `get_example_stream` I think. Can revert to the old thing.

tests/test_datasets.py
```diff
@@ -14,7 +14,7 @@
 class TestDataset(object):
     def setUp(self):
         self.data = [1, 2, 3]
-        self.stream = DataStream(IterableDataset(self.data))
+        self.stream = IterableDataset(self.data).get_example_stream()
```
dwf: Do we need get_example_stream at all anymore?

vdumoulin: Good question. The only place I can think of where we can pass no iteration scheme to `DataStream` is for `IterableDataset` datasets. This doesn't make a very predictable interface.

fuel/streams.py
vdumoulin: I can't think of any case.

dwf: Me neither. Ok.

dwf: (Also noticed there is no **kwargs so yeah)

tests/test_datasets.py
vdumoulin: I'd be inclined to have a convention along the lines of "if you instantiate `DataStream` without giving it an iteration scheme, it'll take the dataset's `example_iteration_scheme` by default"

vdumoulin: That would remove the need for `get_example_stream`, I think.

dwf: SGTM

fuel/streams.py
```diff
@@ -37,12 +37,36 @@ class AbstractDataStream(object):
...
+    @produces_examples.setter
+    def produces_examples(self, value):
+        if self.iteration_scheme:
+            raise ValueError("cannot set produces_examples on {} instance; "
+                             "determined by iteration scheme {}".format(
+                                 self.__class__.__name__,
+                                 self.iteration_scheme))
+        self._produces_examples = value
+
     def get_data(self, request=None):
```
dwf: So I just realized that this method is empty but non-abstract. Just checked with @bartvm, he can't think of any reason for that to be the case either... A subclass inheriting an empty `get_data` would be absolutely useless. Can you add `@abstractmethod`?

vdumoulin: There's the `Merge` transformer which doesn't use `get_data` at all.

dwf: Hmm. That's interesting.

vdumoulin: I've just pushed a commit for that. Short story, we can use `AbstractDataStream`'s default `get_epoch_iterator` and simplify the `Merge` implementation so that it uses `get_data`.

fuel/schemes.py
```diff
@@ -18,6 +18,12 @@ class IterationScheme(object):
     sequential batches, shuffled batches, etc. for datasets that choose to
     support them.
 
+    Attributes
+    ----------
+    requests_examples : bool
```
rizar: Hmm... it reads like a verb 'requests' and a noun 'examples' and is slightly confusing. 'produces_examples', 'requests_are_examples'?

fuel/transformers/__init__.py
```diff
@@ -2,18 +2,27 @@
...
 class Transformer(AbstractDataStream):
     """A data stream that wraps another data stream.
 
+    Subclasses must define a `transform_batch` method (to act on batches),
+    a `transform_example` method (to act on individual examples), or
+    both methods.
+
+    In all cases, the transformer is expected to have the same output
+    type (example or batch) as its input type.  If the transformer
+    subclass is going from batches to examples or vice versa, it
```
rizar: I would also add that overriding `get_data` has to be done when access to `request` is necessary, like in the `Cache` transformer.

fuel/schemes.py
vdumoulin: 'produces_{example,batch}_requests'?

dwf: Simply `example_requests`? To be honest, I find the current name perfectly clear, in the sense that the iteration scheme is in a sense *requesting* examples or batches from the stream.

rizar: With this explanation from David I am fine :)

fuel/transformers/__init__.py
dwf: I think I would also remove "In all cases," and say "Typically," but I feel this has lingered long enough. So I'm going to merge, and then send a separate pull request addressing your comment & mine.

dwf: Oh wait -- yours is already addressed! Okay, merging.

dwf:
Thanks for the hard work @vdumoulin. I think this is a big improvement.

rizar:
+1, this PR is amazing!

vdumoulin:
Thank you for your feedback!

PR #192, opened 2015-07-09
Turn chain into a list to avoid indices depletion #191

### LINKED ISSUE #191: cross_validation doesn't work with SequentialScheme
```chain``` gets depleted after one pass. The scheme returns nothing after first epoch.

```python 
>>> from fuel.schemes import SequentialScheme, cross_validation

>>> cross = cross_validation(SequentialScheme, 10, 3, False, batch_size=5)
>>> train, valid, valid_size = next(cross)

>>> print list(train.get_request_iterator())
[[3, 4, 5, 6, 7], [8, 9]]
>>> print list(train.get_request_iterator())
[]
```

bouthilx:
SequentialScheme gets depleted after one epoch as the chain iterator
gets empty.

Fix #191

### DIFF
--- a/fuel/schemes.py
+++ b/fuel/schemes.py
@@ -271,8 +271,8 @@ def cross_validation(scheme_class, num_examples, num_folds, strict=True,
     for i in xrange(num_folds):
         begin = num_examples * i // num_folds
         end = num_examples * (i+1) // num_folds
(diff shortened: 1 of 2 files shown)


vdumoulin:
LGTM, thanks!

Issue #104, opened 2015-05-11
Write installation docs


julianser:
Add to Readme.rst that libraries certifi and urllib3 are prerequisites, and add instructions on how to install these @vdumoulin.

vdumoulin:
Normally `setup.py` should handle dependencies for you. What was your installation procedure for Fuel?

dwf:
We should probably have installation docs similar to Blocks so I've changed the title of this ticket.

dwf:
Note: as of the merge of #196, Fuel will include some compiled code (Cython). We should carefully document this. `pip install` should work, but editable mode requires some care if you edit the Cython files (i.e. you need to rebuild in-place). See [this comment](https://github.com/mila-udem/fuel/pull/196#discussion_r34636213).

PR #196, opened 2015-07-14
Image-related transformers

dwf:
Includes three transformers for

* Reading streams of images represented as JPEG etc. bytes with PIL
* Upscaling individual images to a minimum width and height if necessary
* Doing random fixed-size crops

Tests forthcoming.


fuel/exceptions.py
```diff
@@ -0,0 +1,25 @@
+class AxisLabelsMismatchError(ValueError):
+    """Raised when a pair of axis labels tuples do not match."""
+    pass
```
vdumoulin: I don't think you need to put a `pass` statement if you have a docstring.

fuel/transformers/__init__.py
```diff
@@ -8,6 +9,39 @@
...
+
+class ExpectsAxisLabels(object):
+    """Mixin used to verify axis labels.
+
+    Notes
+    -----
+    Provides a method :meth:`verify_axis_labels` that should be called
+    with the expected and actual values for an axis labels tuple. If
+    `actual` is `None`, a warning is logged; if it is non-`None` and does
+    not match `expected`, a :class:`AxisLabelsMismatchError` is raised.
```
vdumoulin: Do you think we should eventually make this behaviour configurable? For instance, the user might want to be strict and raise an error if `axis_labels` is `None`, or to log a warning in case of a mismatch instead of an error.

fuel/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+
+class ConfigurationError(Exception):
+    """Error raised when a configuration value is requested but not set."""
+
+
+class MissingInputFiles(Exception):
+    """Exception raised by a converter when input files are not found.
+
+    Parameters
+    ----------
```
vdumoulin: I know you just moved this from one file to another, but I just noticed that we're missing documentation for the `message` argument.

fuel/transformers/image.py
```diff
@@ -0,0 +1,227 @@
...
+    `(channel, height, width)` layout that is pervasive in the world of
+    convolutional networks.
+
+    This SourcewiseTransformer supports streams returning single examples
+    as `bytes` objects (`str` on Python 2.x) as well as streams that
+    return iterables containing such objects. In the case of an
+    iterable, a list of loaded images is returned.
+
+    """
+    def __init__(self, data_stream, which_sources, color_mode='RGB'):
```
vdumoulin: There are some kwargs that would be worth passing up the inheritance chain (namely `iteration_scheme`, which should always be `None`, and `axis_labels`).

fuel/transformers/image.py
```diff
@@ -0,0 +1,227 @@
...
+
+import numpy
+from six import PY3, add_metaclass
+
+from ._image import window_batch_bchw
+from . import ExpectsAxisLabels, SourcewiseTransformer
+from .. import config
+
+
+@add_metaclass(ABCMeta)
```
vdumoulin: For my personal knowledge, why is this a metaclass?

setup.py
```diff
@@ -29,11 +31,15 @@
...
-    install_requires=['six', 'picklable_itertools', 'pyyaml', 'h5py',
+    install_requires=['six', 'picklable_itertools', 'pyyaml', 'h5py', 'cython',
                       'tables', 'progressbar2', 'pyzmq', 'scipy', 'pillow',
                       'requests'],
     extras_require={
         'test': ['nose', 'nose2', 'mock']
     },
-    scripts=['bin/fuel-convert', 'bin/fuel-download', 'bin/fuel-info']
+    scripts=['bin/fuel-convert', 'bin/fuel-download', 'bin/fuel-info'],
+    ext_modules=cythonize(Extension("fuel.transformers._image",
```
vdumoulin: I'm making a note here that I'd like us to discuss that extra compilation step. @bartvm is it something that would bother you?

dwf: Pinged @bartvm a little earlier, he didn't have a problem with it. It'll automatically get done on `python setup.py install`.

dwf: (or `pip install`)

fuel/transformers/image.py
```diff
@@ -0,0 +1,227 @@
...
+            # match the old aspect ratio.
+            aspect = float(width) / height
+            if width < min_width:
+                width = min_width
+                height = int(width / aspect)
+            # N.B.: not elif. Aspect-ratio corrected height may be less
+            # than minimum requested height.
+            if height < min_height:
+                height = min_height
+                width = int(height * aspect)
```
vdumoulin: Note to self: read that logic carefuly and determine if it's right.

fuel/transformers/image.py
```diff
@@ -0,0 +1,227 @@
...
+    def transform_source_example(self, example):
+        self.verify_axis_labels(('channel', 'height', 'width'),
+                                self.data_stream.axis_labels)
+        windowed_height, windowed_width = self.window_shape
+        if not isinstance(example, numpy.ndarray) or example.ndim != 3:
+            raise ValueError("uninterpretable example format; expected "
+                             "ndarray with ndim = 3")
+        image_height, image_width = example.shape[1:]
+        assert image_height >= windowed_height
+        assert image_width >= windowed_width
```
vdumoulin: Should we raise errors instead of making assertions?

fuel/transformers/_image.pyx
```diff
@@ -0,0 +1,60 @@
...
+    Parameters
+    ----------
+    batch : memoryview, 4-dimensional
+        A 4-d tensor containing a batch of images in the expected
+        format above.
+    height_offsets : memoryview, integer, 1-dimensional
+        An array of offsets for the height dimension of each image.
+        Assumed that batch.shape[0] <= height_offsets.shape[0].
+    width_offsets : memoryview, integer, 1-dimensional
+        An array of offsets for the height dimension of each image.
```
vdumoulin: Typo: `height -> width`

setup.py
vdumoulin: Does it work with `python setup.py develop` as well?

vdumoulin:
Other than what I've noted so far, LGTM.

Once my comments are adressed and the tests are implemented, I'll give it another brief read and I think we'll be ready to merge.

setup.py
dwf: Should still work with the caveat that you will need to `python setup.py build_ext --inplace` if you update the `.pyx` file.

fuel/transformers/__init__.py
dwf: +1.

fuel/transformers/image.py
dwf: Nevermind, I don't know why that was there. cdd7290

fuel/transformers/image.py
dwf: 642ca23

fuel/exceptions.py
dwf: 854664b

fuel/exceptions.py
dwf: 93ecaf22

fuel/transformers/image.py
dwf: 8bc215c

fuel/transformers/_image.pyx
dwf: c2a136b

setup.py
vdumoulin: Alright, then we should add a note in #104 so we mention that in the installation docs, but other than that it's fine by me as well.

fuel/transformers/image.py
vdumoulin: There's a more straightforward way to implement that logic: multiply `height` and `width` by `max(1, float(min_width) / width, float(min_height) / height)`. Here's the rationalization:

_Let **w** and **h** be the image width and height, and let **a** and **b** be the minimum width and height. We want to multiply **w** and **h** by the smallest number **r >= 1** such that the constraints **w >= a** and **h >= b** are satisfied._

_If both constraints are satisfied, then **r = 1** since nothing needs to be done. In addition, **w >= a** implies **a / w <= 1** and **h >= b** implies **b / h <= 1**, which means **r = max(1, a / w, b / h)**._

_If **w < a** but **h >= b**, then we need to scale up **w**. The minimum we can do is to bring it up to **a**, which means that **r = a / w**. In addition, **w < a** implies **a / w > 1** and **h >= b** implies **b / h <= 1**, which means **r = max(1, a / w, b / h)**. An analogous reasoning can be used to show that if only the **h** constraint is violated, we still have **r = max(1, a / w, b / h)**._

_Finally, if **w < a** and **h < b**, then we need to scale up both **w** and **h**. The minimum amount of scaling required for the two constraints is **r = a / w** and **r = b / h**, respectively. To satisfy both constraints, we need to choose the biggest scaling factor of those two. In other words, **r = max(a / w, b / h)**. In addition, **w < a** implies **a / w > 1** and **h < b** implies **b / h > 1**, which means **r = max(1, a / w, b / h)**._

_We conclude that **r = max(1, a / w, b / h)**._

dwf:
@vdumoulin So I realized today that there's a flaw in my plan to check `axis_labels`: `data_stream.axis_labels` is a dict. A sourcewise transformer has no idea what the name of the source it's operating on is. I've extended the API to include this information, which can be ignored if the Transformer doesn't want it.

vdumoulin:
The code looks good to me. Once you add tests for the third transformer and the CI tests pass, merge at will.

dwf:
@vdumoulin `ImagesFromBytes` is kinda tricky to set the axis labels, since enumerating every possible mode that PIL supports and figuring out the axes will be a pain. I'm thinking of just implementing it for 'L' ("luminance", i.e. grayscale) and and 'RGB' output modes, and not supporting it for now if you choose `mode=None` (I don't even know how we'd handle this, come to think of it, since each image we decode could have different axes).

vdumoulin:
That's fine by me. If some mode gets requested afterwards, we'll open a new PR to add support for it.

dwf:
@lamblin @dmitriy-serdyuk If either of you guys have time to give this a review (mostly from e18c832 to 2968364) that would be greatly appreciated (@bartvm might also want a look but he's probably busy).

dwf:
To be clear, most of the review has already been done by @vdumoulin. I just want a second look at `ImagesFromBytes` and the tests for that.

fuel/transformers/image.py
```diff
@@ -0,0 +1,263 @@
...
+from PIL import Image
+from six import PY3
+
+from ._image import window_batch_bchw
+from . import ExpectsAxisLabels, SourcewiseTransformer
+from .. import config
+
+
+class ImagesFromBytes(SourcewiseTransformer):
+    """Load from a stream of image-formatted data represented as bytes.
```
dmitriy-serdyuk: It's not very clear what is 
> represented as bytes
Do you mean that it can read any kind of image file?

fuel/transformers/image.py
```diff
@@ -0,0 +1,263 @@
...
+                else:
+                    labels[source] = ('batch', 'channel', 'height', 'width')
+            else:
+                labels[source] = (data_stream.axis_labels[source]
+                                  if source in data_stream.axis_labels
+                                  else None)
+        return labels
+
+
+class MinimumImageDimensions(SourcewiseTransformer, ExpectsAxisLabels):
```
dmitriy-serdyuk: Why do you have `MinimumImageDimensions` and don't have `MaximumImageDimensions`?

I thought usually people resize the smallest axis to some fixed size and then apply random crop.

fuel/transformers/image.py
dwf: Any encoded image supported by PIL, yes. I'll make that clearer.

fuel/transformers/image.py
dwf: You run into this with big heterogeneous collections of images like ImageNet. Some are, unbelievably, shorter than 256 pixels on the short side. However, you're right that we do need a way to downsample as well.

dwf: For now, this transformer exists mainly to make sure that your images are big enough to sample a certain given window size. Other transformers implementing the "scale the shorter side to a particular pixel length" will be forthcoming over the next little while, but for now this is sufficient to guarantee the random crop transformer doesn't choke.

fuel/transformers/image.py
dwf: @dmitriy-serdyuk 8def8d04

fuel/transformers/image.py
dmitriy-serdyuk: I see. We defenitely need an example of usage, but it's for future work.

dmitriy-serdyuk:
It's some problem with scrutinizer, everything else LGTM.

PR #206, opened 2015-07-31
Fix image transpose in MinimumImageDimensions

dwf:
Apparently the test written was not thorough enough.


dwf:
@lamblin @dmitriy-serdyuk @rizar If anyone could give this a look, it should be very quick to review.

dmitriy-serdyuk:
Oh, transpose-related bugs.. Looks good now.

PR #210, opened 2015-08-05
Work around non-GCC compliant compiler nonsense.

dwf:
Apparently Visual C++ has to be different for the sake of being different and chokes on the `-Wno-unused-function`. So, don't bother with it if `sys.platform` is `win32`.


vdumoulin:
Thanks a lot!

PR #211, opened 2015-08-06
Make fuel-* scripts work on Windows

lamblin:
This uses the entry_points/console_scripts feature of setup.py,
so the executables will be created when running pip install,
for instance.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+[report]
+omit =
+    fuel/bin/*
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ __pycache__/
 # C extensions
 *.so
 
+# Cython extensions
+fuel/transformers/_image.c
+
 # Distribution / packaging
 .Python
 env/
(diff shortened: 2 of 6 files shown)


lamblin:
I'm still working on fixing the problem on unix

lamblin:
Tested on Windows and Linux, it seems to work.

lamblin:
I'm not sure what to do about the coverage report.
There may be a way of calling these functions from python, instead of the script itself, so that this counts in the coverage.

vdumoulin:
@lamblin and I agreed offline to exclude the contents of `fuel.bin` from coverall tests for the moment and create a ticket to see what we can test in that module at a later time.

lamblin:
If anyone has an idea of why the build is failing when all individual components seem to be working, that would be appreciated. @dwf? @bartvm?

lamblin:
OK, everything is passing now.

vdumoulin:
I had a quick run with your branch on OSX and everything appears to be working fine on my end.

Issue #88, opened 2015-04-28
Add 'Adult' dataset to built-in datasets


vdumoulin:
You can find info about the Adult dataset [here](http://archive.ics.uci.edu/ml/datasets/Adult).

See these two tutorials (in order) for a crash-course on how to implement a new built-in dataset in Fuel:

* [How to use H5PYDataset to load data in Fuel](http://fuel.readthedocs.org/en/latest/h5py_dataset.html)
* [How to implement a built-in dataset in Fuel](http://fuel.readthedocs.org/en/latest/new_dataset.html)

hantek:
Hi 
I created a PR for this issue, though it hasn't pass Travis right now. I'll go back and check.

vdumoulin:
Closed via #220.

PR #220, opened 2015-08-20
add adult dataset

hantek:
A fix for #88 , not tested yet.


fuel/converters/adult.py
```diff
@@ -0,0 +1,107 @@
...
+
+    Parameters
+    ----------
+    directory : str
+        Directory in which input files reside.
+    output_directory : str
+        Directory in which to save the converted dataset.
+    output_filename : str, optional
+        Name of the saved dataset. Defaults to `None`, in which case a name
+        based on `dtype` will be used.
```
vdumoulin: This description needs adapting for Adult.

fuel/converters/adult.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def convert_adult(directory, output_directory,
+                  output_filename='adult.hdf5'):
+    """Convert the Adult dataset to HDF5.
+
+    Converts the Adult dataset to an HDF5 dataset compatible with
+    :class:`fuel.datasets.Adult`. The converted dataset is saved as
+    'adult.hdf5'.
+    This method assumes the existence of the file `adult.data` and
```
vdumoulin: This paragraph needs a new line before it.

fuel/downloaders/adult.py
```diff
@@ -0,0 +1,24 @@
+from fuel.downloaders.base import default_downloader
+
+
+def fill_subparser(subparser):
+    """Set up a subparser to download the adult dataset file.
+
+    The Iris dataset file `adult.data` and `adult.test` is downloaded from
```
vdumoulin: Typo: Iris.

fuel/converters/adult.py
```diff
@@ -0,0 +1,107 @@
...
+        Directory in which to save the converted dataset.
+    output_filename : str, optional
+        Name of the saved dataset. Defaults to `None`, in which case a name
+        based on `dtype` will be used.
+
+    Returns
+    -------
+    output_paths : tuple of str
+        Single-element tuple containing the path to the converted dataset.
+    """
```
vdumoulin: PEP257 asks that you put a blank line right before the end of a multi-line docstring, like so:

``` python
"""My docstring

Parameters
----------
param : int
    Some description

"""
```

This should solve one of your two Scrutinizer errors.

fuel/converters/adult.py
```diff
@@ -0,0 +1,107 @@
...
+        for i, var in enumerate(variables):
+            data = map(lambda l: l[i], features)
+            if var in continuous:
+                data = map(lambda l: float(l), data)
+                data = numpy.array(data)
+                data = data.reshape(data.shape[0], 1)
+            else:
+                unique_values = list(set(data))
+                data = map(lambda l: unique_values.index(l), data)
+                data = convert_to_one_hot(data)
```
vdumoulin: `convert_to_one_hot` isn't defined anywhere. Solving this will solve the last of the two Scrutinizer errors.

vdumoulin:
Good job, thanks!

There are two Scrutinizer errors to fix. I've commented inline where they appear for convenience.

Like you said, the last thing remaining to do afterwards is to add unit tests for the downloader and the converter. This should increase test coverage and make everything pass.

You can have a look at the tests for Binarized MNIST in `tests/test_downloaders.py` and `tests/test_converters.py` for an overview of what's expected to be tested in a new built-in dataset.

hantek:
@vdumoulin 

The Travis is giving this error:

ERROR: tests.test_converters.TestAdult.test_convert
  File "/home/travis/build/mila-udem/fuel/tests/test_converters.py", line 455, in test_convert
    output_directory=tempdir)
  File "/home/travis/build/mila-udem/fuel/fuel/converters/adult.py", line 63, in convert_adult
    train_content = open(train_path, 'r').readlines()
IOError: [Errno 2] No such file or directory: '/tmp/tmp5vRw54/adult.data'


However, I have already have the downloader and test code for that, Shouldn't it be there?

vdumoulin:
I'll have a look at that. By the way, I think you accidentally committed `fuel/transformers/_image.c`, could you remove it?

hantek:
Yes, sure.

hantek:
Test for python 3.4 doesn't pass, but python 2.7 is OK now. 

All "os.getcwd()" in test_converter.py is getting an error "no such file or directory." Even at positions where I haven't modified. Is there something special with that command in python 3.4?

ERROR: tests.test_converters.TestSVHN.test_format_1_converter
  File "/home/travis/build/mila-udem/fuel/tests/test_converters.py", line 533, in setUp
    cwd = os.getcwd()
FileNotFoundError: [Errno 2] No such file or directory

vdumoulin:
I'll have a look at that.

fuel/converters/adult.py
```diff
@@ -0,0 +1,140 @@
...
+    h5file['targets'].dims[1].label = 'index'
+
+    h5file.flush()
+    h5file.close()
+
+    return (output_path,)
+
+
+def fill_subparser(subparser):
+    subparser.set_defaults(func=convert_adult)
```
vdumoulin: The way `fill_subparser` is implemented has changed slightly. You now need to return the conversion function instead of setting it as `func`'s default value, i.e.

``` python
def fill_subparser(subparser):
    return convert_adult
```

I forgot to mention it, sorry about that!

fuel/downloaders/adult.py
```diff
@@ -0,0 +1,24 @@
...
+        Subparser handling the adult command.
+
+    """
+    subparser.set_defaults(
+        func=default_downloader,
+        urls=['https://archive.ics.uci.edu/ml/machine-learning-databases/'
+              'adult/adult.data',
+              'https://archive.ics.uci.edu/ml/machine-learning-databases/'
+              'adult/adult.test'],
+        filenames=['adult.data', 'adult.test'])
```
vdumoulin: The way `fill_subparser` is implemented has changed slightly. You now need to return the download function instead of setting it as `func`'s default value, i.e.

``` python
def fill_subparser(subparser):
    """[...]"""
    subparser.set_defaults(
        urls=['https://archive.ics.uci.edu/ml/machine-learning-databases/'
              'adult/adult.data',
              'https://archive.ics.uci.edu/ml/machine-learning-databases/'
              'adult/adult.test'],
        filenames=['adult.data', 'adult.test'])
    return default_downloader
```

I forgot to mention it, sorry about that!

tests/test_adult.py
```diff
@@ -0,0 +1,52 @@
+import numpy
+
+from numpy.testing import assert_raises, assert_equal, assert_allclose
+
+from fuel.datasets import Adult
+from tests import skip_if_not_available
+
+
+def test_adult_test():
+    skip_if_not_available(datasets=['adult.hdf5'])
```
vdumoulin: In order for those tests not to be skipped, you'll need to add `adult` [here](https://github.com/mila-udem/fuel/blob/master/.travis.yml#L43), like so:

```
 - ./.travis-data.sh mnist binarized_mnist "caltech101_silhouettes 16" cifar10 cifar100 iris adult
```

tests/test_converters.py
```diff
@@ -209,6 +209,52 @@ def test_read_image_value_error(self):
...
+
+        cwd = os.getcwd()
+        os.chdir(tempdir)
+
+        assert_raises(IOError,
+                      adult.convert_adult,
+                      directory=tempdir,
+                      output_directory=tempdir)
+
+        default_downloader(
```
vdumoulin: I'd prefer that we mock the data instead of downloading it. Even though it's relatively small, things add up when we have lots of datasets, which might slow tests down considerably in the long run.

fuel/converters/adult.py
```diff
@@ -0,0 +1,140 @@
...
+    output_filename : str, optional
+        Name of the saved dataset. Defaults to `adult.hdf5`.
+
+    Returns
+    -------
+    output_paths : tuple of str
+        Single-element tuple containing the path to the converted dataset.
+
+    """
+    # classes = {b'>50K': 0, b'<50K': 1}
```
vdumoulin: It this still useful?

vdumoulin:
I think I found what the issue was. In `fuel/converters/adult.py`, you're using `map`, which behaves differently in Python 2 and Python 3: in Python 2, `map` returns a list, whereas in Python 3, `map` returns an iterator, like `imap` in Python 2. The solution is to wrap `map` calls with `list(map(...))`, so that you get a list in Python 3.

tests/test_converters.py
hantek: The adult dataset has different types of features (not like N x N pixels), and for different type, the way to transform it into neural network input is quite different. So generating a mock dataset can be very complicated...

fuel/converters/adult.py
hantek: It's providing basic information about the input and output of dataset, though not very informative. What do you suggest to add there?

vdumoulin: I'd suggest just removing it.

vdumoulin:
@hantek Thank you for your investment in this PR. The only thing remaining for it to be merged would be to remove the comment I suggested to remove.

tests/test_converters.py
vdumoulin: Alright, let's leave it at that for now, and we'll do something later if it becomes problematic.

hantek:
Hi Vincent, Thanks for your help and comments. I have removed that comment now.

PR #221, opened 2015-08-26
Ensure fuel.server works with scalar features by working around a numpy bug

mjwillson:
When using ServerDatastream, scalar features get turned into length-1 vectors.

I tracked it down to a bug with numpy.ascontiguousarray, which this commit works around: ascontiguousarray makes scalars into arrays (https://github.com/numpy/numpy/issues/5300)

### DIFF
--- a/fuel/server.py
+++ b/fuel/server.py
@@ -10,6 +10,15 @@ logger = logging.getLogger(__name__)
 logging.basicConfig(level='INFO')
 
 
+def ascontiguousarray(array):
+    """Works around https://github.com/numpy/numpy/issues/5300."""
+    array = numpy.array(array)
+    if array.flags['C_CONTIGUOUS']:
+        return array
+    else:
+        return numpy.ascontiguousarray(array)
+
+
 def send_arrays(socket, arrays, stop=False):
     """Send NumPy arrays using the buffer interface and some metadata.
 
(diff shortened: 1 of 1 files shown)


rizar:
It looks good to me, but I do not know much about this part of Fuel. @dwf, @vdumoulin ?

vdumoulin:
Likewise. @mjwillson could you also add a unit test in `fuel/tests/test_server.py` to show exactly how Fuel is currently failing and how your fix solves the issue?

dwf:
It looks like from that numpy ticket that the simpler (and future proof)
thing to do is to call ascontiguousarray with ndmin=0. What do you think?

mjwillson:
so ascontiguousarray doesn't have that argument, it has ndmin=1 hardcoded internally as an argument to array.

It looks like you could call np.array yourself without that argument though, as `np.array(feature, copy=False, order='C')`, so maybe that's a better way to go

vdumoulin:
@mjwillson I'm doing a cleanup of the open PRs. Would you still like to go ahead with this one?

mjwillson:
I'm not using fuel right now so probably won't get around to doing any further work on it soon. Seems like a worthwhile (and easy) fix to make though. Sorry about lack of test, I guess consider it a bug report with a "works for me" fix attached.

vdumoulin:
No problem, thanks for letting me know! I'll close the PR for the meantime.

PR #223, opened 2015-09-02
Subset refactor

### LINKED ISSUE #164: Benchmark the usefulness of maintaining support for slices
We'd like to know what is it exactly we're gaining by using slices as opposed to lists of indices for dataset requests.

We should benchmark it and decide whether or not it's worth supporting it.

See [this discussion](https://github.com/mila-udem/fuel/pull/144#discussion-diff-32262341) for more details.

### LINKED ISSUE #244: H5PYDataset incorrect subset handling leads to dataset split contamination
```
>>> from fuel.datasets import CIFAR10
>>> # One of these should be an error, or at least the get_data call should return 0 examples
>>> valid = CIFAR10(which_sets=('train',), subset=range(50000, 60000))
>>> valid_data = valid.get_data(request=range(0, 10000))
>>> test = CIFAR10(which_sets=('test',))
>>> test_data = test.get_data(request=range(0, 10000))
>>> (test_data[0] == valid_data[0]).all()
True
>>> (test_data[1] == valid_data[1]).all()
True
```
This is a rather serious bug. I spent this weekend using the test set for model selection when I didn't mean to.

vdumoulin:
Fixes #164. Fixes #244. Like suggested, I implemented a `Subset` class which factors out the logic of indexing/slicing through multiple data types, in the context of a subset of the data.

I refactored `H5PYDataset` and `IndexableDataset` to make use of `Subset`.

One of the main highlights of this implementation is that `Subset` does sanity checks on requests (e.g. making sure that a slice won't produce an empty or truncated output), which means that it would be safe to bring slices back into Fuel.

It also has the ability to merge two subsets in the most efficient way possible (e.g. two overlapping slices with step size of 1 would be merged into a single slice, as opposed to converted to a list, which is currently the case in `H5PYDataset`), and to detect when a list is a contiguous sequence and convert it into a slice. This behaviour could be expanded to accomodate more complex cases, if it's ever needed.

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -571,13 +571,13 @@ class H5PYDataset(Dataset):
             data.append(
                 subset.index_within_subset(
                     handle[source_name], request,
-                    safe_hdf5_indexing=self.sort_indices))
+                    sort_indices=self.sort_indices))
             # If this source has variable length, get the shapes as well
             if source_name in self.vlen_sources:
                 shapes.append(
                     subset.index_within_subset(
                         handle[source_name].dims[0]['shapes'], request,
-                        safe_hdf5_indexing=self.sort_indices))
+                        sort_indices=self.sort_indices))
             else:
                 shapes.append(None)
         return data, shapes
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -8,7 +8,8 @@ import tables
 from six.moves import zip, range
 
 from fuel.datasets import Dataset
-from fuel.utils import do_not_pickle_attributes, iterable_fancy_indexing
+from fuel.utils import (do_not_pickle_attributes, iterable_fancy_indexing,
+                        Subset)
 
 
 @do_not_pickle_attributes('nodes', 'h5file')
@@ -178,7 +179,7 @@ class H5PYDataset(Dataset):
         if which_sets_invalid_value:
             raise ValueError('`which_sets` should be an iterable of strings')
         self.which_sets = which_sets
-        self._subset_template = subset if subset else slice(None)
+        self._subset_within_sets = subset if subset else slice(None)
         self.load_in_memory = load_in_memory
         self.driver = driver
         self.sort_indices = sort_indices
@@ -455,34 +456,6 @@ class H5PYDataset(Dataset):
                     indices[source] = h5file[row['indices']]
         return indices
 
-    @staticmethod
-    def unsorted_fancy_index(request, indexable):
-        """Safe unsorted list indexing.
-
-        Some objects, such as h5py datasets, only support list indexing
-        if the list is sorted.
-
-        This static method adds support for unsorted list indexing by
-        sorting the requested indices, accessing the corresponding
-        elements and re-shuffling the result.
-
-        Parameters
-        ----------
-        request : list of int
-            Unsorted list of example indices.
-        indexable : any fancy-indexable object
-            Indexable we'd like to do unsorted fancy indexing on.
-
-        """
-        if len(request) > 1:
-            indices = numpy.argsort(request)
-            data = numpy.empty(shape=(len(request),) + indexable.shape[1:],
-                               dtype=indexable.dtype)
-            data[indices] = indexable[numpy.array(request)[indices], ...]
-        else:
-            data = indexable[request]
-        return data
-
     def load(self):
         # If the dataset is unpickled, it makes no sense to have an external
         # file handle. However, since `load` is also called during the lifetime
@@ -495,70 +468,45 @@ class H5PYDataset(Dataset):
         self._out_of_memory_open()
         handle = self._file_handle
 
-        # Load subset slices / indices
-        subsets = []
-        if len(self.which_sets) > 1:
-            indices = defaultdict(list)
-            for split in self.which_sets:
-                split_start_stop = self.get_start_stop(handle, split)
-                split_indices = self.get_indices(handle, split)
-                for source_name in self.sources:
-                    if source_name in split_indices:
-                        ind = split_indices[source_name]
-                    else:
-                        ind = list(range(*split_start_stop[source_name]))
-                    indices[source_name] = sorted(
-                        set(indices[source_name] + ind))
-        else:
-            start_stop = self.get_start_stop(handle, self.which_sets[0])
-            indices = self.get_indices(handle, self.which_sets[0])
-        num_examples = None
-        for source_name in self.sources:
-            subset = self._subset_template
-            # If subset has a step greater than 1, we convert it to a list,
-            # otherwise we won't be able to take that subset within the context
-            # of a split defined by a slice.
-            if hasattr(subset, 'step') and subset.step not in (1, None):
-                subset = list(range(len(handle[source_name])))[subset]
-            self._subset_template = subset
-            if source_name in indices:
-                source_subset = indices[source_name]
-            else:
-                source_subset = slice(*start_stop[source_name])
-            if hasattr(subset, 'step') and hasattr(source_subset, 'step'):
-                subset = slice(
-                    source_subset.start
-                    if subset.start is None else subset.start,
-                    source_subset.stop
-                    if subset.stop is None else subset.stop,
-                    subset.step)
-                subsets.append(subset)
-                subset_num_examples = subset.stop - subset.start
-            else:
-                if hasattr(source_subset, 'step'):
-                    source_subset = numpy.arange(
-                        source_subset.start, source_subset.stop)
-                subset = source_subset[subset]
-                subsets.append(subset)
-                subset_num_examples = len(subset)
-            if num_examples is None:
-                num_examples = subset_num_examples
-            if num_examples != subset_num_examples:
-                raise ValueError("sources have different lengths")
-        self.subsets = subsets
+        # Infer subsets based on `which_sets`
+        subsets = [None for source_name in self.sources]
+        for split in self.which_sets:
+            start_stop = self.get_start_stop(handle, split)
+            indices = self.get_indices(handle, split)
+            for i, source_name in enumerate(self.sources):
+                if source_name in indices:
+                    source_split_subset = Subset(
+                        indices[source_name], len(handle[source_name]))
+                else:
+                    source_split_subset = Subset(
+                        slice(*start_stop[source_name]),
+                        len(handle[source_name]))
+                if subsets[i] is None:
+                    subsets[i] = source_split_subset
+                else:
+                    subsets[i] += source_split_subset
+        # Sanity check to make sure that all sources have equal length
+        if any(subset.num_examples != subsets[0].num_examples for subset in
+                subsets):
+            raise ValueError("sources have different lengths")
+        # Produce the final subsets by taking the `subset` constructor argument
+        # into account.
+        self.subsets = [Subset.subset_of(subset, self._subset_within_sets)
+                        for subset in subsets]
 
-        # Load data sources and source shapes
+        # Load data sources and source shapes (if requested)
         if self.load_in_memory:
             data_sources = []
             source_shapes = []
             for source_name, subset in zip(self.sources, self.subsets):
-                if hasattr(subset, 'step'):
-                    data_source = handle[source_name][subset]
-                else:
-                    data_source = handle[source_name][list(subset)]
-                data_sources.append(data_source)
+                list_or_slice = subset.list_or_slice
+                data_sources.append(
+                    subset.index_within_subset(
+                        handle[source_name], slice(None)))
                 if source_name in self.vlen_sources:
-                    shapes = handle[source_name].dims[0]['shapes'][subset]
+                    shapes = subset.index_within_subset(
+                        handle[source_name].dims[0]['shapes'],
+                        slice(None))
                 else:
                     shapes = None
                 source_shapes.append(shapes)
@@ -572,10 +520,7 @@ class H5PYDataset(Dataset):
 
     @property
     def num_examples(self):
-        if hasattr(self.subsets[0], 'step'):
-            return self.subsets[0].stop - self.subsets[0].start
-        else:
-            return len(self.subsets[0])
+        return self.subsets[0].num_examples
 
     def open(self):
         return None if self.load_in_memory else self._out_of_memory_open()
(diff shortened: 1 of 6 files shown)


vdumoulin:
There is one tests failing which I'd like to have an opinion on. Right now the assumption for `H5PYDataset` is that if we pass it an unsorted list as `subset` in the constructor, this subset will remain unsorted. `Subset` breaks that, because it sorts list-based subsets and makes sure indices are unique. I was wondering if it was okay to remove this assumption altogether (and notify users via the mailing list), or if there's a use case for that which I haven't thought of.

lamblin:
I think if a user explicitly requests unsorted or duplicated indices, the result should either be an error, or the examples in the requested order. Returning the examples in a different order, or a different number of examples (in the case of duplication), could lead to all sorts of bugs that are hard to detect.
This could happen for instance if the user is iterating over several indexable objects at the same time, not all of which are HDF5Datasets.

dwf:
It seems like there might be legitimate uses for returning the same example
twice...

fuel/datasets/hdf5.py
```diff
@@ -495,87 +467,62 @@ def load(self):
...
+            indices = self.get_indices(handle, split)
+            for i, source_name in enumerate(self.sources):
+                if source_name in indices:
+                    source_split_subset = Subset(
+                        indices[source_name], len(handle[source_name]))
+                else:
+                    source_split_subset = Subset(
+                        slice(*start_stop[source_name]),
+                        len(handle[source_name]))
+                if subsets[i] is None:
```
rizar: I think it should be possible to instantiate an empty subset. This would allow us not to have this branching.

vdumoulin: Could you elaborate on that? In which use case would you do that?

vdumoulin:
Regarding unsorted / duplicate indices:

* I feel like iteration schemes are better suited for iterating over a dataset subset in shuffled order. If the user wants to re-shuffle at every batch, `ShuffledScheme` is the way to go. I the user wants to shuffle only once before training, `SequentialScheme`'s constructor accepts a list as its `examples` argument which can be used for that.
* Allowing duplicates may make things very hard to debug. For instance, if the user merges two subsets which overlap, it may make the resulting subset unbalanced without the user noticing.

lamblin:
On the other hand, a user may want to sample examples with replacement, or explicitly include a subset more than once to compensate for an existing inbalance in the dataset.

vdumoulin:
@lamblin These issues can also be solved with iteration schemes.

fuel/datasets/hdf5.py
rizar: You start by creating a list of `None`'s. A few lines below you select between assignment and `+=` operator based on whether or not an element is `None`. Alternatively, you could prepare a list of `Subset()` objects which would denote the empty set and only use `+=` later.

fuel/utils.py
```diff
@@ -14,6 +16,347 @@
...
+        if self.original_num_examples != other.original_num_examples:
+            raise ValueError("trying to add two Subset instances with "
+                             "different numbers of original examples, they "
+                             "can't possibly belong to the same dataset")
+        if self.is_list:
+            if other.is_list:
+                return self.__class__(self.list_or_slice + other.list_or_slice,
+                                      self.original_num_examples)
+            else:
+                return self.__class__(self.list_or_slice +
```
rizar: To reduce copy-paste this branch could make a similar call with swapped arguments.

fuel/utils.py
```diff
@@ -14,6 +16,347 @@
...
+                    # In case of overlap, the solution is to choose the
+                    # smallest start value and largest stop value.
+                    return self.__class__(slice(min(self_start, other_start),
+                                                max(self_stop, other_stop),
+                                                self_step),
+                                          self.original_num_examples)
+                # Everything else is transformed into lists before merging.
+                else:
+                    return self.__class__(
+                        self[list(range(self.num_examples))] +
```
rizar: The idiom `self[list(range(self.num_examples))]` could be factored out as `Subset.as_list_subset`.

fuel/utils.py
```diff
@@ -14,6 +16,347 @@
...
+            return (stop - 1) // step + 1
+
+    def _subset_sanity_check(self, list_or_slice, num_examples):
+        if self._is_list(list_or_slice):
+            self._list_subset_sanity_check(list_or_slice, num_examples)
+        else:
+            self._slice_subset_sanity_check(list_or_slice, num_examples)
+
+    def _list_subset_sanity_check(self, indices, num_examples):
+        if len(indices) == 0:
```
rizar: Why not allow empty subset?

fuel/utils.py
```diff
@@ -14,6 +16,347 @@
...
+            raise ValueError('Subset instances cannot be defined by an empty '
+                             'list (it would be an empty subset)')
+        if any(index < 0 for index in indices):
```
rizar: `if min(indices) < 0`?

rizar:
@vdumoulin , I agree with you that the order of iteration should be defined by the iteration scheme, but it seems to me that the code from this PR does not preserve the order defined by iteration scheme requests. I think that it is okay to sort the indices of the subset, but requests might need more delicate handling.

fuel/utils.py
```diff
@@ -14,6 +16,347 @@
...
+                                                self_step),
+                                          self.original_num_examples)
+                # Everything else is transformed into lists before merging.
+                else:
+                    return self.__class__(
+                        self[list(range(self.num_examples))] +
+                        other[list(range(other.num_examples))],
+                        self.original_num_examples)
+
+    def __getitem__(self, key):
```
rizar: This one needs docs.

fuel/utils.py
```diff
@@ -14,6 +16,347 @@
...
+        if isinstance(subset_request, int):
+            request, = self[[subset_request]]
+        else:
+            request = self[subset_request]
+        # Integer or slice requests can be processed directly.
+        if isinstance(request, int) or hasattr(request, 'step'):
+            return indexable[request]
+        # List requests are handled differently whether the indexable is
+        # an HDF5 dataset, a numpy array or another type of indexable.
+        else:
```
rizar: `else:` is not required here.

vdumoulin:
@rizar and I talked offline. Bottom line: the code in the PR *does* preserve the order of requests.

fuel/utils.py
vdumoulin: All requests on that subset would return empty arrays, I'm not sure there's a use case for that. Plus, bugs caused by accidentally creating an empty subset may be hard to debug.

rizar: But on the other hand people might create an empty subset intentionally. It
might be useful as a neutral element in the   subset algebra.

fuel/datasets/hdf5.py
vdumoulin: That's a good point.

As a middle ground between what you suggest and my objections to allowing empty subsets, what do you think of allowing empty subsets but raising an error if someone tries to index within an empty subset?

rizar: I would like to ask for a terminology clarification.  Do you mean using
__getitem__ with an empty subset as self, or do you mean using
index_with_subset with an empty subset?

vdumoulin: Actually, there are already safegards in place that check whether we're trying to request something within the context of a subset. They would raise an error because the request would always be greater than subset's number of examples (0).

I revise my position and accept your suggestion.

vdumoulin:
@rizar Sorry about the delay. I think I addressed all of your comments. Would you mind having another look?

fuel/utils.py
```diff
@@ -14,6 +16,390 @@
...
+                             'value)')
+        if (slice_.start is not None and slice_.stop is not None and
+                slice_.start >= slice_.stop):
+            raise ValueError('slice-based requests cannot have a start value '
+                             'greater than or equal to its stop value (this '
+                             'would produce an empty return value)')
+
+    def _beautify_list(self, indices):
+        # List elements should be unique and sorted
+        indices = list(sorted(set(indices)))
```
rizar: Just for the record: this might take a while if the list is long. But we can have it as is for a while, for premature optimization is typically a bad thing.

fuel/utils.py
```diff
@@ -14,6 +16,390 @@
...
+                self.get_list_representation() +
+                other.get_list_representation(),
+                self.original_num_examples)
+        # List-based subsets are merged by concatenating their indices.
+        if self.is_list:
+            return self.__class__(self.list_or_slice + other.list_or_slice,
+                                  self.original_num_examples)
+        # Slice-based subsets are merged into a slice-based subset if they
+        # overlap, otherwise they're converted to a list-based subset.
+        else:
```
rizar: This `else:` is not required, because you have `return` in the `if` section.

fuel/datasets/hdf5.py
```diff
@@ -495,87 +467,60 @@ def load(self):
...
-            if num_examples != subset_num_examples:
-                raise ValueError("sources have different lengths")
-        self.subsets = subsets
-
-        # Load data sources and source shapes
+        # Infer subsets based on `which_sets`
+        subsets = [Subset.empty_subset(len(handle[source_name]))
+                   for source_name in self.sources]
+        for split in self.which_sets:
+            start_stop = self.get_start_stop(handle, split)
```
rizar: This can be made in a further ticker, but I guess we could have `get_subsets` instead of `get_start_top` and `get_indices`.

fuel/utils.py
```diff
@@ -14,6 +16,390 @@
...
+            request = self[subset_request]
+        # Integer or slice requests can be processed directly.
+        if isinstance(request, int) or hasattr(request, 'step'):
+            return indexable[request]
+        # If the indexable is an HDF5 dataset, it only supports fancy
+        # indexing with sorted lists. As a workaround, if
+        # `safe_hdf5_indexing` is set to `True`, Subset will do the
+        # indexing in sorted order, and reshuffle the result in the
+        # original order.
+        if isinstance(indexable, h5py.Dataset) and safe_hdf5_indexing:
```
rizar: I am not sure that we should put that much awareness of `H5Py` related issues into `Subset`. What if instead of `safe_hdf5_indexing` we just call the option `safe_indexing` and sort the indices of the subset regardless of what the indexable is? This would make things significantly simpler.

rizar: Or maybe we should even call it `sort_indices,` exactly like in `H5PyDataset`. It is perhaps less ambiguous.

fuel/datasets/hdf5.py
```diff
@@ -178,7 +178,7 @@ def __init__(self, file_or_path, which_sets, subset=None,
         if which_sets_invalid_value:
             raise ValueError('`which_sets` should be an iterable of strings')
         self.which_sets = which_sets
-        self._subset_template = subset if subset else slice(None)
+        self._subset_within_sets = subset if subset else slice(None)
```
rizar: This variable name is rather confusing. What about `user_given_subset`?

fuel/datasets/hdf5.py
rizar: Given that this is the only place in `hdf5.py` where `get_start_stop` and `get_indices` are used, I would even consider replacing them with `get_subset` in this PR.

fuel/utils.py
```diff
@@ -14,6 +16,390 @@
...
+            other_start, other_stop, other_step = other_sss
+            # In case of overlap, the solution is to choose the smallest start
+            # value and largest stop value.
+            if not (self_stop < other_start or self_start > other_stop):
+                return self.__class__(slice(min(self_start, other_start),
+                                            max(self_stop, other_stop),
+                                            self_step),
+                                      self.original_num_examples)
+            # Everything else is transformed into lists before merging.
+            else:
```
rizar: The same applies for this `else:`

fuel/utils.py
```diff
@@ -14,6 +16,390 @@
...
+        :class:`list` or :class:`slice`
+            The translated request to be used on the dataset.
+
+        """
+        self._request_sanity_check(key, self.num_examples)
+        # slice(None, None, None) selects the whole subset, no need to index
+        # anything
+        if key == slice(None, None, None):
+            return self.list_or_slice
+        elif self._is_list(key):
```
rizar: Maybe it is a matter of state, but I think that all this extra indentation makes understanding harder compared to a list of `if condition: return smth` statements.

rizar:
It is getting better and better! Can you please rebase it so that the test for Python 2.7 were run?

dwf:
Anyone know if #244 is resolved by this PR?

fuel/datasets/hdf5.py
vdumoulin: Will do.

fuel/utils.py
vdumoulin: Will do.

fuel/utils.py
vdumoulin: I personally find the alternative harder to read.

vdumoulin:
@dwf #244 is resolved by this PR: the statement instantiating `valid` produces a dataset with 0 examples, and `get_data` raises

    ValueError: slice-based requests cannot have a stop value greater than the number of
    examples the subset spans (this would produce a return value with smaller length
    than expected)

fuel/utils.py
rizar: I understand that it is a matter of habit, but I think that the version below is really simpler. And every time I review `H5PYDataset` PRs I am confused by redundant `else` statements. When I see one, I think that it should be there for a reason and I try to find this reason and I fail to find it. @dwf, I am curious to hear which style you prefer.

```python
if key == slice(None, None, None):
    return self.list_or_slice
if  self._is_list(key):
     if self.is_list:
          return [self.list_or_slice[index] for index in key]
     start, stop, step = self.slice_to_numerical_args(
         self.list_or_slice, self.original_num_examples)
     return [start + (index * step) for index in key]
if self.is_list:
    return self.list_or_slice[key]
start, stop, step = self.slice_to_numerical_args(
    self.list_or_slice, self.original_num_examples)
key_start, key_stop, key_step = self.slice_to_numerical_args(
    key, self.num_examples)
return slice(start + step * key_start,
    start + step * key_stop,
    step * key_step)
```

fuel/datasets/hdf5.py
```diff
@@ -495,87 +452,59 @@ def load(self):
...
-            if num_examples != subset_num_examples:
-                raise ValueError("sources have different lengths")
-        self.subsets = subsets
-
-        # Load data sources and source shapes
+        # Infer subsets based on `which_sets`
+        subsets = [Subset.empty_subset(len(handle[source_name]))
+                   for source_name in self.sources]
+        for split in self.which_sets:
+            start_stop, indices = self.get_subset(handle, split)
```
rizar: Sorry, I was not clear enough, I thought `get_subset` should return a `Subset` object. Is it  possible, or am I missing something?

vdumoulin:
@rizar I caved in and simplified the nested `if`s structure. I also factored out creating `Subset` objects.

I think I covered all of your comments, please let me know if there's something I forgot.

dwf:
Okay, thanks! I updated the ticket description so that #244 auto-closes.

dwf:
By the way, would it make sense to add the code in #244 as a test for this PR?

vdumoulin:
@dwf sure, I included a regression for that.

rizar:
LGTM, great job!

Issue #195, opened 2015-07-13
Write an interactive tutorial


lamblin:
- [x] Make an ipython notebook (maybe some slides as well) for the Deep Learning Summer School 2015 (https://github.com/mila-udem/summerschool2015). Done in https://github.com/mila-udem/summerschool2015/pull/2
- [x] Adapt it and integrate into Fuel's documentation

bartvm:
On this topic, I dug up an old e-mail I wrote a long time ago about IPython notebooks for documentation:

> Right. There's 3 options for examples: Python files, IPython notebooks, or in the Sphinx-generated rST documentation. I think it's important to be able to mix code examples with explanations, links to documentation, etc. so I prefer the Sphinx/rSt option right now, because it automatically gets unit-tested. In Pylearn2 half the IPython notebooks are broken because they're hard to cover by unit tests. They are also annoying to use with Git, it's hard to link to other parts of the documentation from within them. Lastly, hosting them is more of a pain.

> However, I think interactivity is important too, and it's too bad that are only static... IPython does have a relatively unknown feature that we could advertise: Using "%doctest_mode" you can copy-paste code from the docs straight into IPython, and it will automatically filter out the unnecessary whitspace and >>> signs.

> Ideally (but this is quite long-term thinking), I think it would be amazing to have a Sphinx extension that automatically converts certain docs into IPython notebooks (see e.g. https://github.com/shashi/rst2ipynb) so that any tutorial would just have a link "Download as IPython Notebook" at the top... One can dream! But at least the possibility of doing this makes me feel better about sticking with rST for now.

vdumoulin:
I wouldn't mind porting the notebook to Sphinx/rSt after the summer school.

As a presentation tool I prefer the notebook to static documentation, as it allows to do live coding, but I agree that Sphinx/rSt is more suited to the online docs.

nouiz:
Interesting to know all this. Thanks for sharing.
Le 13 juil. 2015 17:00, "Bart van Merriënboer" <notifications@github.com> a
écrit :

dwf:
Incidentally... https://pypi.python.org/pypi/pytest-ipynb

I'm hearing more and more good things about pytest.

fvisin:
Copied and pasting for loops from rst files result in an error in iPython, e.g.:
```python
In [87]: %doctest_mode
Exception reporting mode: Plain
Doctest mode is: ON
>>> >>> for scheme in schemes:
[...]

PR #225, opened 2015-09-03
Overhaul documentation

### LINKED ISSUE #219: Clarify the distinction between IterableDataset and IndexableDataset
Example:

```python
import numpy as np
import theano
from blocks.bricks import Rectifier, Softmax
from blocks.bricks import MLP
from blocks.initialization import IsotropicGaussian, Constant
from blocks.model import Model, ComputationGraph
from blocks.bricks.cost import CategoricalCrossEntropy
from fuel.datasets import IterableDataset
from fuel.schemes import ShuffledScheme
from blocks.main_loop import MainLoop
from blocks.extensions import FinishAfter, ProgressBar, Printing
from blocks.extensions.monitoring import TrainingDataMonitoring
from blocks.algorithms import GradientDescent, Adam


seed = 1234
rng = np.random.RandomState(seed)
features = rng.uniform(size=(1280, 8))
targets = rng.uniform(low=0, high=1, size=(1280, 1)).astype(np.int32)

dataset = IterableDataset(
    iterables={'features': features, 'targets': targets})

data_stream = DataStream(dataset=dataset, iteration_scheme=ShuffledScheme(batch_size=128, examples=dataset.num_examples))



x = theano.tensor.matrix('features', dtype='float64')
y = theano.tensor.matrix('targets', dtype='int32')

mlp = MLP(activations=[Rectifier(), Softmax()], dims=[8, 4, 2], name='mlp',
               weights_init=IsotropicGaussian(0.1),
               biases_init=Constant(0.))

mlp.initialize()
y_hat = mlp.apply(x)
cg = ComputationGraph(y_hat)
cost = CategoricalCrossEntropy().apply(y_hat, y)
model = Model([cost])



loop = MainLoop(data_stream=data_stream,
                algorithm=GradientDescent(step_rule=Adam(),
                                          cost=cost,
                                          parameters=cg.parameters),
                model=model,
                extensions=[
                            TrainingDataMonitoring([cost],
                                                   prefix="train",
                                                   after_epoch=True),
                            FinishAfter(after_n_epochs=5),
                            ProgressBar(),
                            Printing()]
                )
loop.run()
```

Yields:

C:\Python27\lib\site-packages\fuel\iterator.pyc in __next__(self)
     28     def __next__(self):
     29         if self.request_iterator is not None:
---> 30             data = self.data_stream.get_data(next(self.request_iterator))
     31         else:
     32             data = self.data_stream.get_data()

C:\Python27\lib\site-packages\fuel\streams.pyc in get_data(self, request)
    165     def get_data(self, request=None):
    166         """Get data from the dataset."""
--> 167         return self.dataset.get_data(self.data_state, request)
    168 
    169     def get_epoch_iterator(self, **kwargs):

C:\Python27\lib\site-packages\fuel\datasets\base.pyc in get_data(self, state, request)
    302     def get_data(self, state=None, request=None):
    303         if state is None or request is not None:
--> 304             raise ValueError
    305         return next(state)
    306 

ValueError: 

Original exception:
	ValueError

vdumoulin:
Fixes #219, #195.

@lamblin would you mind having a look and telling me what you think of it?

### DIFF
--- a/docs/extending_fuel.rst
+++ b/docs/extending_fuel.rst
@@ -13,7 +13,7 @@ New dataset classes
 New dataset classes are implemented by subclassing :class:`Dataset` and
 implementing its :meth:`get_data` method. If your dataset interacts with
 stateful objects (e.g. files on disk), then you should also override the
-:class:`open` and :class:`close` methods.
+:meth:`open` and :meth:`close` methods.
 
 If your data fits in memory, you can save yourself some time by inheriting from
 :class:`IndexableDataset`. In that case, all you need to do is load the data as
@@ -113,7 +113,7 @@ If you were to build a transformer that only works on batches, you would pass
 anyone tried to use your transformer on an example data stream, an error would
 automatically be raised.
 
-Let's test our doubler on some dummy dataset. **Note that the this implementation
+Let's test our doubler on some dummy dataset. **Note that this implementation
 is for illustration purposes only: it is brittle and only works on numpy
 arrays.**
 
--- a/docs/extending_fuel.rst
+++ b/docs/extending_fuel.rst
@@ -1,0 +1,268 @@
+Extending Fuel
+==============
+
+In this section, we'll cover how to extend three important components of Fuel:
+
+* Dataset classes
+* Transformers
+* Iteration schemes
+
+New dataset classes
+-------------------
+
+New dataset classes are implemented by subclassing :class:`Dataset` and
+implementing its :meth:`get_data` method. If your dataset interacts with
+stateful objects (e.g. files on disk), then you should also override the
+:class:`open` and :class:`close` methods.
+
+If your data fits in memory, you can save yourself some time by inheriting from
+:class:`IndexableDataset`. In that case, all you need to do is load the data as
+a :class:`dict` mapping source names to their corresponding data and pass it to
+the superclass as the ``indexables`` argument.
+
+For instance, here's how you would implement a specialized class to interface
+with ``.npy`` files.
+
+>>> from collections import OrderedDict
+>>> import numpy
+>>> from six import iteritems
+>>> from fuel.datasets import IndexableDataset
+>>>
+>>> class NPYDataset(IndexableDataset):
+...     def __init__(self, source_paths, **kwargs):
+...         indexables = OrderedDict(
+...             [(source, numpy.load(path)) for
+...              source, path in iteritems(source_paths)])
+...         super(NPYDataset, self).__init__(indexables, **kwargs)
+
+Here's this class in action:
+
+>>> numpy.save('npy_dataset_features.npy',
+...            numpy.arange(40).reshape((10, 4)))
+>>> numpy.save('npy_dataset_targets.npy',
+...            numpy.arange(10).reshape((10, 1)))
+>>> dataset = NPYDataset(OrderedDict([('features', 'npy_dataset_features.npy'),
+...                                   ('targets', 'npy_dataset_targets.npy')]))
+>>> state = dataset.open()
+>>> print(dataset.get_data(state=state, request=[0, 1, 2, 3]))
+(array([[ 0,  1,  2,  3],
+       [ 4,  5,  6,  7],
+       [ 8,  9, 10, 11],
+       [12, 13, 14, 15]]), array([[0],
+       [1],
+       [2],
+       [3]]))
+>>> dataset.close(state)
+
+.. doctest::
+   :hide:
+
+   >>> import os
+   >>> os.remove('npy_dataset_features.npy')
+   >>> os.remove('npy_dataset_targets.npy')
+
+New transformers
+----------------
+
+An important thing to know about data streams is that they distinguish
+between two types of outputs: single examples, and batches of examples.
+Depending on your choice of iteration scheme, a data stream's
+``produces_examples`` property will either be ``True`` (it produces examples)
+or ``False`` (it produces batches).
+
+Transformers are aware of this, and as such implement two distinct methods:
+:meth:`transform_example` and :meth:`transform_batch`. A new transformer is
+typically implemented by subclassing Transformer and implementing one or both
+of these methods.
+
+As an example, here's how you would double the value of some ``'features'``
+data source.
+
+>>> from fuel.transformers import Transformer
+>>>
+>>> class FeaturesDoubler(Transformer):
+...     def __init__(self, data_stream, **kwargs):
+...         super(FeaturesDoubler, self).__init__(
+...             data_stream=data_stream,
+...             produces_examples=data_stream.produces_examples,
+...             **kwargs)
+...         
+...     def transform_example(self, example):
+...         if 'features' in self.sources:
+...             example = list(example)
+...             index = self.sources.index('features')
+...             example[index] *= 2
+...             example = tuple(example)
+...         return example
+...     
+...     def transform_batch(self, batch):
+...         if 'features' in self.sources:
+...             batch = list(batch)
+...             index = self.sources.index('features')
+...             batch[index] *= 2
+...             batch = tuple(batch)
+...         return batch
+
+Most transformers you'll implement will call their superclass constructor by
+passing the data stream and declaring whether they produce examples or batches.
+Since we wish to support both batches and examples, we'll declare our output
+type to be the same as our data stream's output type.
+
+If you were to build a transformer that only works on batches, you would pass
+``produces_examples=False`` and implement only :meth:`transform_batch`. If
+anyone tried to use your transformer on an example data stream, an error would
+automatically be raised.
+
+Let's test our doubler on some dummy dataset. **Note that the this implementation
+is for illustration purposes only: it is brittle and only works on numpy
+arrays.**
+
+>>> from fuel.schemes import SequentialExampleScheme, SequentialScheme
+>>> from fuel.streams import DataStream
+>>>
+>>> dataset = IndexableDataset(
+...     indexables=OrderedDict([
+...         ('features', numpy.array([1, 2, 3, 4])),
+...         ('targets', numpy.array([-1, 1, -1, 1]))]))
+>>> example_scheme = SequentialExampleScheme(examples=dataset.num_examples)
+>>> example_stream = FeaturesDoubler(
+...     data_stream=DataStream(
+...         dataset=dataset, iteration_scheme=example_scheme))
+>>> batch_scheme = SequentialScheme(
+...     examples=dataset.num_examples, batch_size=2)
+>>> batch_stream = FeaturesDoubler(
+...     data_stream=DataStream(
+...         dataset=dataset, iteration_scheme=batch_scheme))
+>>> print([example for example in example_stream.get_epoch_iterator()])
+[(2, -1), (4, 1), (6, -1), (8, 1)]
+>>> print([batch for batch in batch_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-1,  1])), (array([6, 8]), array([-1,  1]))]
+
+If you think the :meth:`transform_example` and :meth:`transform_batch`
+implementations are repetitive, you're right! In cases where the example and
+batch implementations of a transformer are the same, you can subclass from
+:class:`AgnosticTransformer` instead. It requires that you implement a
+:meth:`transform_any` method, which will be called by both
+:meth:`transform_example` and :meth:`transform_batch`.
+
+>>> from fuel.transformers import AgnosticTransformer
+>>> 
+>>> class FeaturesDoubler(AgnosticTransformer):
+...     def __init__(self, data_stream, **kwargs):
+...         super(FeaturesDoubler, self).__init__(
+...             data_stream=data_stream,
+...             produces_examples=data_stream.produces_examples,
+...             **kwargs)
+... 
+...     def transform_any(self, data):
+...         if 'features' in self.sources:
+...             data = list(data)
+...             index = self.sources.index('features')
+...             data[index] *= 2
+...             data = tuple(data)
+...         return data
+
+So far so good, but our transformer could be more general: what if we want to
+double ``'features'`` and ``'targets'``, or only ``'targets'``?
+
+Transformers which are applied sourcewise like our doubler should usually
+subclass from :class:`SourcewiseTransformer`. Their constructor takes an
+additional ``which_sources`` keyword argument specifying which sources to apply
+the transformer to. It's expected to be a tuple of source names. If
+``which_sources`` is ``None``, then the transformer is applied to all sources.
+Subclasses of :class:`SourcewiseTransformer` should implement a
+:meth:`transform_source_example` method and/or a :meth:`transform_source_batch`
+method, which apply on an individual source.
+
+There also exists an :class:`AgnosticSourcewiseTransformer` class for cases
+where the example and batch implementations of a sourcewise transformer are the
+same. This class requires a :meth:`transform_any_source` method to be
+implemented.
+
+>>> from fuel.transformers import AgnosticSourcewiseTransformer
+>>> 
+>>> class Doubler(AgnosticSourcewiseTransformer):
+...     def __init__(self, data_stream, **kwargs):
+...         super(Doubler, self).__init__(
+...             data_stream=data_stream,
+...             produces_examples=data_stream.produces_examples,
+...             **kwargs)
+... 
+...     def transform_any_source(self, source, _):
+...         return 2 * source
+
+Let's try this implementation on our dummy dataset.
+
+>>> target_stream = Doubler(
+...     data_stream=DataStream(
+...         dataset=dataset,
+...         iteration_scheme=batch_scheme),
+...     which_sources=('targets',))
+>>> all_stream = Doubler(
+...     data_stream=DataStream(
+...         dataset=dataset,
+...         iteration_scheme=batch_scheme),
+...     which_sources=None)
+>>> print([batch for batch in target_stream.get_epoch_iterator()])
+[(array([1, 2]), array([-2,  2])), (array([3, 4]), array([-2,  2]))]
+>>> print([batch for batch in all_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-2,  2])), (array([6, 8]), array([-2,  2]))]
+
+Finally, there exists a :class:`Mapping` transformer which acts as a swiss-army
+knife transformer. In addition to a data stream, its constructor accepts a
+function which will be applied to data coming from the stream.
+
+Here's how you would implement the feature doubler using :class:`Mapping`.
+
+>>> from fuel.transformers import Mapping
+>>> 
+>>> features_index = dataset.sources.index('features')
+>>> def double(data):
+...     data = list(data)
+...     data[features_index] *= 2
+...     return tuple(data)
+>>> mapping_stream = Mapping(
+...     data_stream=DataStream(
+...         dataset=dataset, iteration_scheme=batch_scheme),
+...     mapping=double)
+>>> print([batch for batch in mapping_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-1,  1])), (array([6, 8]), array([-1,  1]))]
+
+
+New iteration schemes
+---------------------
+
+New iteration schemes are implemented by subclassing :class:`IterationScheme`
+and implementing a :meth:`get_request_iterator` method, which should return an
+iterator that returns lists of indices.
+
+Two subclasses of :class:`IterationScheme` typically serve as a basis for other
+iteration schemes: :class:`IndexScheme` (for schemes requesting examples) and
+:class:`BatchScheme` (for schemes requesting batches). Both subclasses are
+instantiated by providing a list of indices or a number of examples, and
+:class:`BatchScheme` accepts an additional ``batch_size`` argument.
+
+Here's how you would implement an iteration scheme that iterates over even
+examples:
+
+>>> from fuel.schemes import IndexScheme, BatchScheme
+>>> # `iter_` : A picklable version of `iter`
+>>> from picklable_itertools import iter_, imap
+>>> # Partition all elements of a sequence into tuples of length at most n
+>>> from picklable_itertools.extras import partition_all
+
+>>> class ExampleEvenScheme(IndexScheme):
+...     def get_request_iterator(self):
+...         indices = list(self.indices)[::2]
+...         return iter_(indices)
+>>> class BatchEvenScheme(BatchScheme):
+...     def get_request_iterator(self):
+...         indices = list(self.indices)[::2]
+...         return imap(list, partition_all(self.batch_size, indices))
+
+Here are the two iteration scheme classes in action:
+
+>>> print(list(ExampleEvenScheme(10).get_request_iterator()))
+[0, 2, 4, 6, 8]
+>>> print(list(BatchEvenScheme(10, 2).get_request_iterator()))
+[[0, 2], [4, 6], [8]]
--- a/docs/extending_fuel.rst
+++ b/docs/extending_fuel.rst
@@ -4,11 +4,30 @@ Extending Fuel
 In this section, we'll cover how to extend three important components of Fuel:
 
 * Dataset classes
-* Transformers
 * Iteration schemes
+* Transformers
+
+.. warning::
+    The code presented in this section is for illustration purposes only and
+    is not intended to be used outside of this tutorial.
 
 New dataset classes
 -------------------
+
+.. admonition:: In summary
+    :class: tip
+
+    * Subclass from :class:`Dataset`.
+
+      - Implement the :meth:`get_data` method.
+      - If your dataset interacts with stateful objects (e.g. files on disk),
+        override the :meth:`open` and :meth:`close` methods.
+
+    * Subclass from :class:`IndexableDataset` if your data fits in memory.
+
+      - :class:`IndexableDataset` constructor accepts an ``indexables``
+        argument, which is expected to be a :class:`dict` mapping from source
+        names to their corresponding data.
 
 New dataset classes are implemented by subclassing :class:`Dataset` and
 implementing its :meth:`get_data` method. If your dataset interacts with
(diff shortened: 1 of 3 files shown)


docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+* Transformers
+* Iteration schemes
+
+New dataset classes
+-------------------
+
+New dataset classes are implemented by subclassing :class:`Dataset` and
+implementing its :meth:`get_data` method. If your dataset interacts with
+stateful objects (e.g. files on disk), then you should also override the
+:class:`open` and :class:`close` methods.
```
lamblin: Maybe those should be ```:method:`open```` or ```:meth:`open```` (and same thing for `close`).

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+passing the data stream and declaring whether they produce examples or batches.
+Since we wish to support both batches and examples, we'll declare our output
+type to be the same as our data stream's output type.
+
+If you were to build a transformer that only works on batches, you would pass
+``produces_examples=False`` and implement only :meth:`transform_batch`. If
+anyone tried to use your transformer on an example data stream, an error would
+automatically be raised.
+
+Let's test our doubler on some dummy dataset. **Note that the this implementation
```
lamblin: "the this" -> "this"

docs/extending_fuel.rst
vdumoulin: You're right, that was a typo.

vdumoulin:
@fvisin if you have time, would you mind having a look? I'm looking for an outsider's perspective, and your critical eye for what's unclear or lacking would be very helpful for me.

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+...     
+...     def transform_batch(self, batch):
+...         if 'features' in self.sources:
+...             batch = list(batch)
+...             index = self.sources.index('features')
+...             batch[index] *= 2
+...             batch = tuple(batch)
+...         return batch
+
+Most transformers you'll implement will call their superclass constructor by
```
fvisin: This paragraph is a non-mandatory best practice. It could be separated from the rest of the text and added as a "Hint box" at the end of this section.

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+Since we wish to support both batches and examples, we'll declare our output
+type to be the same as our data stream's output type.
+
+If you were to build a transformer that only works on batches, you would pass
+``produces_examples=False`` and implement only :meth:`transform_batch`. If
+anyone tried to use your transformer on an example data stream, an error would
+automatically be raised.
+
+Let's test our doubler on some dummy dataset. **Note that this implementation
+is for illustration purposes only: it is brittle and only works on numpy
```
fvisin: This note could be moved into a "Warning box" at the beginning of the page.

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+>>> batch_stream = FeaturesDoubler(
+...     data_stream=DataStream(
+...         dataset=dataset, iteration_scheme=batch_scheme))
+>>> print([example for example in example_stream.get_epoch_iterator()])
+[(2, -1), (4, 1), (6, -1), (8, 1)]
+>>> print([batch for batch in batch_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-1,  1])), (array([6, 8]), array([-1,  1]))]
+
+If you think the :meth:`transform_example` and :meth:`transform_batch`
+implementations are repetitive, you're right! In cases where the example and
```
fvisin: I would refer more explicitly to the example, so that it's clear that they are not repetitive in general. Something like "You might have noticed that in this example the :meth:`transform_example` and :meth:`transform_batch` implementations are identical."

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+...     def transform_any(self, data):
+...         if 'features' in self.sources:
+...             data = list(data)
+...             index = self.sources.index('features')
+...             data[index] *= 2
+...             data = tuple(data)
+...         return data
+
+So far so good, but our transformer could be more general: what if we want to
+double ``'features'`` and ``'targets'``, or only ``'targets'``?
```
fvisin: A general remark: I always find ``'features'`` and ``'targets'`` a bit irritating as they are "hard-coded" names (even if their behavior is not hard-coded). I would avoid using them as much as possible. This is admittedly very subjective though!

More up to the point, here I would say "So far so good, but our transformer applies the same transformation to every source in the dataset. What if we want to be more general and be able to select which sources we want to process with our transformer?"

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+...     data_stream=DataStream(
+...         dataset=dataset,
+...         iteration_scheme=batch_scheme),
+...     which_sources=None)
+>>> print([batch for batch in target_stream.get_epoch_iterator()])
+[(array([1, 2]), array([-2,  2])), (array([3, 4]), array([-2,  2]))]
+>>> print([batch for batch in all_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-2,  2])), (array([6, 8]), array([-2,  2]))]
+
+Finally, there exists a :class:`Mapping` transformer which acts as a swiss-army
```
fvisin: A minor note: here you could follow the same reasoning of the rest of the tutorial, saying "Finally, what if we not only wanted to select to which source to apply the transformation, but also the transformation?"

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+...     data = list(data)
+...     data[features_index] *= 2
+...     return tuple(data)
+>>> mapping_stream = Mapping(
+...     data_stream=DataStream(
+...         dataset=dataset, iteration_scheme=batch_scheme),
+...     mapping=double)
+>>> print([batch for batch in mapping_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-1,  1])), (array([6, 8]), array([-1,  1]))]
+
```
fvisin: [Finally a comment on the content and not on the style!!] 
How should I handle the case "I want to select the source *and* the function"?

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+---------------------
+
+New iteration schemes are implemented by subclassing :class:`IterationScheme`
+and implementing a :meth:`get_request_iterator` method, which should return an
+iterator that returns lists of indices.
+
+Two subclasses of :class:`IterationScheme` typically serve as a basis for other
+iteration schemes: :class:`IndexScheme` (for schemes requesting examples) and
+:class:`BatchScheme` (for schemes requesting batches). Both subclasses are
+instantiated by providing a list of indices or a number of examples, and
```
fvisin: "a number of examples" --> "the number of examples to be retrieved"

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+:class:`BatchScheme` accepts an additional ``batch_size`` argument.
+
+Here's how you would implement an iteration scheme that iterates over even
+examples:
```
fvisin: The `imap` and `partition_all` function might be non-trivial for an entry-level python user. Maybe you could elaborate on what is returned (without getting into the details of `imap` and `partition_all` as they are not the core of this tutorial).

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+...     return tuple(data)
+>>> mapping_stream = Mapping(
+...     data_stream=DataStream(
+...         dataset=dataset, iteration_scheme=batch_scheme),
+...     mapping=double)
+>>> print([batch for batch in mapping_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-1,  1])), (array([6, 8]), array([-1,  1]))]
+
+
+New iteration schemes
```
fvisin: Logically, I would put the `Iteration schemes` section before the `Transformers` one.

docs/overview.rst
```diff
@@ -1,11 +1,324 @@
...
+Each of those four tasks is delegated to a particular class of objects, which
+we'll be introducing in order.
+
+Datasets: interfacing with data
+-------------------------------
+
+The :class:`Dataset` class is responsible for interfacing with the data and
+handling data access requests. Subclasses of :class:`Dataset` specialize in
+certain types of data.
+
```
fvisin: Datasets contain one or more **sources** of data such as, e.g., an array of images, a list of labels, a dictionary specifying an ontology, etc, ... Each source in a Dataset is identified by a unique name.

docs/extending_fuel.rst
fvisin: Nevermind, you will subclass from both.

docs/extending_fuel.rst
```diff
@@ -0,0 +1,268 @@
...
+...         iteration_scheme=batch_scheme),
+...     which_sources=None)
+>>> print([batch for batch in target_stream.get_epoch_iterator()])
+[(array([1, 2]), array([-2,  2])), (array([3, 4]), array([-2,  2]))]
+>>> print([batch for batch in all_stream.get_epoch_iterator()])
+[(array([2, 4]), array([-2,  2])), (array([6, 8]), array([-2,  2]))]
+
+Finally, there exists a :class:`Mapping` transformer which acts as a swiss-army
+knife transformer. In addition to a data stream, its constructor accepts a
+function which will be applied to data coming from the stream.
```
fvisin: [...] accepts a function *through it's constructor's parameter `mapping`* which will be applied [...]

fvisin:
Great work @vdumoulin !!!

I will go through the rest in the next days. 

A general comment on what I have read so far: the tutorials are clear and very well written and they provide an excellent introduction to Fuel for new users. 
For experienced Fuel users who are looking for a quick information, the tutorial approach might be difficult to consult though. I suggest to add a wrap-up section at the end of each tutorial (or section in a tutorial) with the main concepts expressed as a list, possibly in a well separated Tip or Info box.

E.g., for the `Transformers`:
{info:title=Transformers}
* Implement :meth:`transform_example` if they provide single examples and/or :meth:`transform_batch` if they provide batches
* Subclass from :class:`AgnosticTransformer` if the examples and batches implementation is the same
    * Implement :meth:`transform_any` 
* Subclass from :class:`SourcewiseTransformer` if are applied sourcewise. In this case their constructor takes an additional `which_sources` keyword argument
    * If which_sources is None, then the transformer is applied to all sources
    * Implement :meth:`transform_source_example` method and/or a :meth:`transform_source_batch` 
* Subclass from :class:`AgnosticSourcewiseTransformer` if are applied sourcewise and the examples and batches implementation is the same
    * Implement :meth:`transform_any_source`
* Subclass :class:`Mapping` if the function applied on the data is not fixed
    * Its constructor's parameter `mapping` specifies the function that will be applied to the data.
{info}

docs/extending_fuel.rst
vdumoulin: I've added a `notes` section explaining that. Could you tell me if you think it's clear enough?

vdumoulin:
@fvisin thanks a lot for your very valuable feedback! I think I addressed everything. Please tell me if there's something I've missed.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+>>> from fuel.datasets import IterableDataset
+>>> dataset = IterableDataset(
+...     iterables=OrderedDict([('features', features), ('targets', targets)]),
+...     axis_labels=OrderedDict([('features', ('batch', 'height', 'width')),
+...                              ('targets', ('batch', 'index'))]))
+
+We can ask the dataset what sources of data it provides by accessing its
+``sources`` attribute. We can also know which axes correspond to what by
+accessing its ``axis_labels`` attribute. It also has a ``num_examples`` property
+telling us the number of examples it contains.
```
fvisin: It is my understanding that you can get the `sources`, etc, .. with any :class:`Dataset`, not only with :class:`IterableDataset`. In this case, it would be better to write every common part to the :class:`Dataset` section and keep here only those that are specific to :class:`IterableDataset`, e.g., `iterables`

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+      - As such, it allows random access.
+
+The :class:`Dataset` class is responsible for interfacing with the data and
+handling data access requests. Subclasses of :class:`Dataset` specialize in
+certain types of data.
+
+Datasets contain one or more **sources** of data, such as an array of images,
+a list of labels, a dictionary specifying an ontology, etc. Each source in a
+dataset is identified by a unique name.
+
```
fvisin: If I am not wrong :class:`Dataset` is not an abstract class. If so, I would specify the parameters required to instantiate a :class:`Dataset` object. Especially if you specify in the :class:`IterableDataset` section later (see comment below)

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+IterableDataset
+^^^^^^^^^^^^^^^
+
+The simplest :class:`Dataset` subclass is :class:`IterableDataset`, which
+interfaces with iterable objects.
+
+It is created by passing a :class:`dict` mapping source names to their
+associated data and, optionally, a :class:`dict` mapping source names to tuples
+of axis labels.
+
```
fvisin: I suggest to add the names of the dicts, i.e.: It is created by passing a :class:`dict` `iterables` mapping source names to their associated data and, optionally, a :class:`dict` `axis_labels` mapping source names to tuples of axis labels.

fvisin: Also, this might be clearer: "mapping each source name to a tuple of axis labels."

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+>>> from collections import OrderedDict
+>>> from fuel.datasets import IterableDataset
+>>> dataset = IterableDataset(
+...     iterables=OrderedDict([('features', features), ('targets', targets)]),
+...     axis_labels=OrderedDict([('features', ('batch', 'height', 'width')),
+...                              ('targets', ('batch', 'index'))]))
+
+We can ask the dataset what sources of data it provides by accessing its
```
fvisin: It's trivial, but I would add "Once the dataset has been created, ..."

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+Datasets themselves are stateless objects (as opposed to, say, an open file
+handle, or an iterator object). In order to request data from the dataset, we
+need to ask it to instantiate some stateful object with which it will interact.
+This is done through the :meth:`Dataset.open` method:
+
+>>> state = dataset.open()
+>>> print(state.__class__.__name__)
+imap
+
+We see that in :class:`IterableDataset`'s case the state is an iterator object.
```
fvisin: Not really sure about my English here, but "We can see" sounds better to me.

Again, if `Dataset` is not abstract/an interface, this part seems generic enough to go in that section.

fvisin: You could also add a reference to [imap](https://docs.python.org/2/library/itertools.html#itertools.imap)

docs/overview.rst
vdumoulin: It is an abstract class (its `get_data` method is abstract).

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+>>> print(dataset.get_data(state=state))
+(array([[ 47, 211],
+       [ 38,  53]]), array([0]))
+>>> while True:
+...     try:
+...         __ = dataset.get_data(state=state)
+...     except StopIteration:
+...         print('Iteration over')
+...         break
+Iteration over
```
fvisin: I would merge the two examples into one:
``` python
>>> while True:
...     try:
...         print(dataset.get_data(state=state))
...     except StopIteration:
...         print('Iteration over')
...         break
...     
(array([[ 47, 211],
       [ 38,  53]]), array([0]))
(array([[204, 116],
       [152, 249]]), array([3]))
(array([[143, 177],
       [ 23, 233]]), array([0]))
(array([[154,  30],
       [171, 158]]), array([1]))
(array([[236, 124],
       [ 26, 118]]), array([2]))
(array([[186, 120],
       [112, 220]]), array([2]))
(array([[ 69,  80],
       [201, 127]]), array([2]))
(array([[246, 254],
       [175,  50]]), array([3]))
Iteration over
```

docs/overview.rst
vdumoulin: The problem is `Dataset` is an abstract class, so I won't be able to accompany these explanations with example code.

I could however move the explanation above the `IterableDataset` subsection and provide an example in that subsection saying "like all `Dataset` subclasses, we can access `IterableDataset`'s `sources`, `axis_labels` and `num_examples` attributes.

fvisin: I like this idea, as it clearly separates what is shared among all datasets and what is `IterableDataset` specific.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+>>> from fuel.datasets import IndexableDataset
+>>> dataset = IndexableDataset(
+...     indexables=OrderedDict([('features', features), ('targets', targets)]),
+...     axis_labels=OrderedDict([('features', ('batch', 'height', 'width')),
+...                              ('targets', ('batch', 'index'))]))
+
+The main advantage of :class:`IndexableDataset` over :class:`IterableDataset`
+is that it allows random access of the data it contains. In order to do so, we
+need to pass an additional ``request`` argument to :meth:`get_data` in the form
+of a list of indices.
```
fvisin: Is it possible to iterate sequentially without specifying the `request` parameter? E.g., say that I usually want to access data sequentially but in some less ordinary cases I also want to get a specific item. Should I instantiate both an `IterableDataset` object and an `IndexableDataset` object in this case?

fvisin: Sorry, I found the answer in the following example. Since it has no state I take it's not possible to iterate sequentially.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+       [[204, 116],
+        [152, 249]]]), array([[0],
+       [3]]))
+>>> dataset.close(state=state)
+
+See how :class:`IndexableDataset` returns a ``None`` state: this is because
+there's no actual state to maintain in this case.
+
+Restricting sources
+^^^^^^^^^^^^^^^^^^^
```
fvisin: This should be put in the generic `Dataset` explanation. You might want to create a `Dataset` section (now there is no `Dataset` section, as it is at the beginning of the tutorial), and put `Restricting sources` as a subsection of that section.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+
+Datasets: interfacing with data
+-------------------------------
+
+.. admonition:: In summary
+    :class: tip
+
+    * :class:`Dataset` subclasses are responsible for interfacing with your
+      data.
+    * A :class:`Dataset` instance has the following useful attributes:
```
fvisin: [Very picky comment] I would write "Any" instead of "A", to empathize that every subclass instance share these attributes

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+Each of those four tasks is delegated to a particular class of objects, which
+we'll be introducing in order.
+
+Datasets: interfacing with data
+-------------------------------
+
+.. admonition:: In summary
+    :class: tip
+
+    * :class:`Dataset` subclasses are responsible for interfacing with your
```
fvisin: :class:`Dataset` is an abstract class. It's subclasses are responsible ...

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+* At each iteration step, apply some transformation to the selected data points.
+
+Each of those four tasks is delegated to a particular class of objects, which
+we'll be introducing in order.
+
+Datasets: interfacing with data
+-------------------------------
+
+.. admonition:: In summary
+    :class: tip
```
fvisin: I would add another level to the list, moving all these items in one of this top level items:
* :class:`Dataset`
* :class:`IterableDataset`
* :class:`IndexableDataset`

fvisin: I suggest a summary like this:

* **:class:`Dataset`**
    * It is an abstract class. Its subclasses are responsible for interfacing with your data.
    * **Constructor's arguments:**
        * ``sources``: optional, to be set to use only a subset of a dataset's sources.
    * **Instances' attributes:**
        - ``sources``: tuple of strings indicating which sources are provided by the dataset, and their ordering (which determines the return order of :meth:`get_data`).
        - ``axis_labels``: :class:`dict` mapping from source names to tuples of strings or ``None``. Used to document the axis semantics of the dataset's sources. Can be set via the ``axis_labels`` constructor argument.
        - ``num_examples``: when available, represents the number of examples the dataset holds.
    * **Instances' methods:**
        - :meth:`open`: returns a ``state`` object the dataset will interact with, e.g. a file handle, or ``None`` if stateless.
        - :meth:`get_data`: given a ``state`` object, returns data.
        - :meth:`close`: given a ``state`` object, properly closes it. 
        - :meth:`reset`: given a ``state`` object, properly closes it and returns a fresh one.

* **:class:`IterableDataset`**
    * Allows to interface with iterable objects in general.
    * The state returned by :meth:`IterableDataset.open` is an iterator object.
    * Can only iterate examplewise and sequentially.
    * **Constructor's arguments:**
        - ``iterables``: a :class:`dict` mapping from source names to their corresponding iterable objects.

* **:class:`IndexableDataset`**
    * Allows to interface with indexable objects in general.
    * Allows random data access.
    * Stateless, i.e., the state returned by :meth:`IndexableDataset.open` is ``None``.
    * **Constructor's arguments:**
        - ``indexables``: is expected to be a :class:`dict` mapping from source names to their corresponding indexable objects.
    * **Instances' methods:**
        - :meth:`get_data`: expects a list of indices as its ``request`` keyword argument.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+
+    * Use :class:`IterableDataset` to interface with iterable objects in
+      general.
+
+      - :class:`IterableDataset` accepts an ``iterables`` argument, which is
+        expected to be a :class:`dict` mapping from source names to their
+        corresponding iterable objects.
+      - The state :meth:`IterableDataset.open` returns is an iterator object.
+      - Its :meth:`get_data` method doesn't accept requests.
+      - As such, it can only iterate exaplewise and sequentially.
```
fvisin: exaplewise --> examplewise

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+
+>>> state = dataset.open()
+>>> scheme = ShuffledScheme(examples=dataset.num_examples, batch_size=4)
+>>> for request in scheme.get_request_iterator():
+...     data = dataset.get_data(state=state, request=request)
+...     print(data[0].shape, data[1].shape)
+(4, 2, 2) (4, 1)
+(4, 2, 2) (4, 1)
+>>> dataset.close(state)
+
```
fvisin: I guess not all the `iterationScheme`s work with every `Dataset`, e.g. I would expect `ShuffledScheme` not to work with a sequential `IterableDataset`. You could mention it here (no need to add it also to the summary IMHO).

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+.. admonition:: In summary
+    :class: tip
+
+    * :class:`Transformer` subclasses are responsible for taking data stream(s)
+      as input and producing a data stream as output, which applies some
+      transformation to the input stream(s).
+    * Transformers can be chained together to form complex data processing
+      pipelines.
+
+Some data streams take data streams as input. We call them *transformers*, and
```
fvisin: This is confusing: is :class:`Transformer` a subclass of :class:`DataStream`? If not, I would make sure not to suggest that `Transformer`s are some kind of `DataStream`s.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+they enable us to build complex data preprocessing pipelines.
+
+Transformers are :class:`Transformer` subclasses. Most of the the transformers
+you'll encounter are located in the ``fuel.transformers`` module. Here are some
```
fvisin: You didn't mention in which module to find `DataStream`s, `Dataset`s, etc, .. before, I wouldn't do it here either. Also, mentioning that most of them are there makes me curious on where to find the rest of them. The link to the class documentation and the example should be enough.

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+    * :class:`Transformer` subclasses are responsible for taking data stream(s)
+      as input and producing a data stream as output, which applies some
+      transformation to the input stream(s).
+    * Transformers can be chained together to form complex data processing
+      pipelines.
+
+Some data streams take data streams as input. We call them *transformers*, and
+they enable us to build complex data preprocessing pipelines.
+
+Transformers are :class:`Transformer` subclasses. Most of the the transformers
```
fvisin: "the the"

docs/overview.rst
```diff
@@ -1,11 +1,410 @@
...
+        [ 0.15740086,  1.51078666]],
+<BLANKLINE>
+       [[-1.00065091, -0.84717417],
+        [ 0.84106998, -0.19140991]]]), array([[2],
+       [0],
+       [3],
+       [2]]))
+
+Now, let's imagine that for some reason (e.g. running Theano code on GPU) we
+need features to have a data type of ``float32``.
```
fvisin: [...] need features to have a data type of ``float32``. We can cast them on-the-fly with a :class:`ScaleAndShift` `Transformer`

docs/overview.rst
```diff
@@ -15,7 +414,7 @@ sight, but can be very powerful.
    IterationScheme -> RequestIterator [label=" Returns"];
    RequestIterator -> DataIterator [style=dashed, label=" Argument to"];
    DataIterator -> DataStream [label=" Gets data from"];
-   DataStream -> DataStream [style=dashed, label=" Gets data from (wrapper)"];
+   DataStream -> DataStream [style=dashed, label=" Gets data from (transformer)"];
    { rank=same; RequestIterator DataIterator }
 
 Datasets
```
fvisin: I would move the visual diagram at the end of the `Division of labour` section.

The textual summary instead doesn't seem to be useful anymore and, on the contrary, seems to be either wrong or at least badly phrased in some parts, e.g.: 
> Data streams can produce data set iterators (epoch iterators) which will use the data stream's state to return actual data.

My understanding is that 
* `Iterators` produce list of indexes 
* `DataStream`s do not produce `Iterators`, they are *given* a `Dataset` and an `Iterator` and produce actual data.

As such, I suggest to completely remove it.

fvisin:
@vdumoulin excellent work! :+1: 
I am sorry it took me so long to go though it, but I am finally done. I think this refactoring will help a lot users approaching Fuel for the first time as well as experienced users with bad memory!

dwf:
Haven't had a chance to review this in detail but skimming it it looks really nice, +1 on Francesco's comment about amnesiac experienced users. :+1:

fvisin:
@vdumoulin here is the remainder of what we discussed today:
* It would be great to refactor the rest of the documentation, at least adding the summaries.
* It would be even greater to have a *cheat sheet* with all the summaries together. The [include directive](http://docutils.sourceforge.net/docs/ref/rst/directives.html#include) could be used to avoid code duplication.
* Questions about the `Import your own data` link at the end of the `overview` tutorial:
    * If you modify `Dataset`s so that they are function instead of classes, could the `import your own data` documentation be merged with the `contribute to a dataset` one?
    * If not, it should be made clear the distinction between the two of them (changing their names to something more informative might be enough)

docs/overview.rst
vdumoulin: Note for self: fixed.

docs/overview.rst
vdumoulin: The problem is in the generic section we haven't covered `get_data` yet, which makes it kind of awkward. I'll try to cook up something.

fvisin: I agree it is not simple to find a fully satisfying solution here.

I suggest to introduce the generic dataset and all its methods in the "generic section", specifying that, given it is an abstract class, all the code will be shown using one the classes that implement it, but that the same code would work for any other class that implement the Dataset class. 
This would have these advantages:
* to show working code 
* to have in the generic section what applies to all the datasets
* to keep the other sections (e.g. `IterableDataset`) "straight to the point", showing only what is really specific to them

vdumoulin:
@fvisin I think your suggestions regarding refactoring the rest of the documentation as well as the cheat sheet are great ideas, but I'd leave that for another PR.

Regarding your question: possibly. The `contribute a dataset` section deals with `fuel-download` and `fuel-convert`, which you don't need for your own data. However, they could be presented as optional convenience steps and the emphasis could be on the similarities between built-in datasets and local datasets.

vdumoulin:
I'll merge the PR in its current state, and I'll create a ticket for @fvisin's additional suggestions.

fvisin:
> Regarding your question: possibly. The contribute a dataset section deals with fuel-download and fuel-convert, which you don't need for your own data. However, they could be presented as optional convenience steps and the emphasis could be on the similarities between built-in datasets and local datasets.

I guess that making it more clear that fuel can be used with your own data might already be enough. Anyway, I am fine with whatever you decide is best.

PR #227, opened 2015-09-03
make fuel accept multiple data paths

### LINKED ISSUE #218: Make Fuel accept a list of paths as its data path in fuelrc
See [this comment](https://github.com/mila-udem/fuel/pull/214#discussion_r37345322).

hantek:
This PR tries to fix #218

### DIFF
--- a/fuel/config_parser.py
+++ b/fuel/config_parser.py
@@ -96,6 +96,25 @@ def extra_downloader_converter(value):
     return value
 
 
+def multiple_paths_parser(value):
+    """
+    Parses data_path argument.
+
+    Parameters
+    ----------
+    value : str
+        a string of data paths separated by  ":".
+
+    Returns
+    -------
+    value : list
+        a list of strings indicating each data paths.
+
+    """
+    value = value.split(os.path.pathsep)
+    return value
+
+
 class Configuration(object):
     def __init__(self):
         self.config = {}
@@ -170,7 +189,8 @@ class Configuration(object):
 config = Configuration()
 
 # Define configuration options
-config.add_config('data_path', type_=str, env_var='FUEL_DATA_PATH')
+config.add_config('data_path', type_=multiple_paths_parser,
+                  env_var='FUEL_DATA_PATH')
 config.add_config('default_seed', type_=int, default=1)
 config.add_config('extra_downloaders', type_=extra_downloader_converter,
                   default=[], env_var='FUEL_EXTRA_DOWNLOADERS')
--- a/fuel/config_parser.py
+++ b/fuel/config_parser.py
@@ -97,8 +97,7 @@ def extra_downloader_converter(value):
 
 
 def multiple_paths_parser(value):
-    """
-    Parses data_path argument.
+    """Parses data_path argument.
 
     Parameters
     ----------
--- a/fuel/config_parser.py
+++ b/fuel/config_parser.py
@@ -110,7 +110,8 @@ def multiple_paths_parser(value):
         a list of strings indicating each data paths.
 
     """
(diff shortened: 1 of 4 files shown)


vdumoulin:
Thanks! I started making some comments on the previous PR, just make sure they're addressed.

hantek:
Thanks. I saw your comments, and all of them are fixed on this PR.

hantek:
Hi Vincent,

Travis is complaining about this error:

ERROR: tests.test_utils.TestFindInDataPath.test_raises_error_on_file_not_found

  File "/home/travis/build/mila-udem/fuel/tests/test_utils.py", line 47, in setUp
    self.original_data_path = config.data_path
  File "/home/travis/build/mila-udem/fuel/fuel/config_parser.py", line 150, in __getattr__
    return config_setting['type'](value)
  File "/home/travis/build/mila-udem/fuel/fuel/config_parser.py", line 113, in multiple_paths_parser
    value = value.split(os.path.pathsep)
AttributeError: 'list' object has no attribute 'split'

But when I ran the test locally on my machine, it is OK:
"""
[linzhou@bart7 tests]$ python -i test_utils.py 
>>> aaa = TestFindInDataPath()
>>> aaa.setUp()
>>> aaa.test_raises_error_on_file_not_found()
>>> aaa.test_returns_file_path()
>>> aaa.test_returns_first_file_found()
>>> 
"""

Am I running the test in a wrong way?

fuel/datasets/caltech101_silhouettes.py
```diff
@@ -39,4 +39,6 @@ def __init__(self, which_sets, size=28, load_in_memory=True, **kwargs):
 
     @property
     def data_path(self):
-        return os.path.join(config.data_path, self.filename)
+        for data_path in config.data_path:
+            if os.path.exists(os.path.join(data_path, self.filename)):
+                return os.path.join(data_path, self.filename)
```
vdumoulin: You should use `fuel.utils.find_in_data_path` instead of explicitly making the check here.

tests/__init__.py
```diff
@@ -35,7 +35,11 @@ def skip_if_not_available(modules=None, datasets=None, configurations=None):
     if datasets and not hasattr(config, 'data_path'):
         raise SkipTest
     for dataset in datasets:
-        if not os.path.exists(os.path.join(config.data_path, dataset)):
```
vdumoulin: Ditto.

fuel/config_parser.py
```diff
@@ -96,6 +96,24 @@ def extra_downloader_converter(value):
...
+    value : str
+        a string of data paths separated by  ":".
+
+    Returns
+    -------
+    value : list
+        a list of strings indicating each data paths.
+
+    """
+    value = value.split(os.path.pathsep)
```
vdumoulin: This will fail if `value` is already a list. To work around this, use

``` python
if isinstance(value, six.string_types):
    value = value.split(os.path.pathsep)
```

Incidentally, it's what makes the tests fail, I think.

vdumoulin:
@hantek if you want to run the tests like they'll be run on Travis, you can enter

``` bash
$ python -c nose2.__main__ -B *tests
```

in the repo root directory.

fuel/config_parser.py
hantek: Yes that solves the problem. 

So the "value" passed to multiple_paths_parser() could be a list, which can only happen when it is already an output of multiple_paths_parser(). Does that mean, config.add_config('data_path', ...) is called even if after parsing command line parameters?

hantek:
Travis has passed.

Issue #202, opened 2015-07-24
Merge streams never terminate?


dwf:
Discovered while trying to pin down #201:

Tested on a branch that is rebased on top of 74634c910... Probably still the case in master:

```python
from fuel.datasets import IterableDataset
from fuel.streams import DataStream
from fuel.transformers import Merge
from fuel.schemes import SequentialScheme
english = IterableDataset(['Hello world!'])
french = IterableDataset(['Bonjour le monde!'])
next(merged_stream.get_epoch_iterator())
streams = (DataStream(english, iteration_scheme=SequentialScheme(batch_size=1, examples=1)),
           DataStream(french, iteration_scheme=SequentialScheme(batch_size=1, examples=1)))
merged_stream = Merge(streams, ('english', 'french'))
iterator = merged_stream.get_epoch_iterator()
next(iterator)
next(iterator)
```
This should raise a `StopIteration` but doesn't. In fact, you get an empty tuple, and keep getting empty tuples seemingly forever.

Returning the same number of items in a tuple than the number of sources for a stream is something that should be `assert`ed.

vdumoulin:
Fixed via #229.

Issue #201, opened 2015-07-24
Merge does not work with Batch transformer


ejls:
We get a backtrace when running the example code of Merge with an added Batch :

```python
from fuel.datasets import IterableDataset
english = IterableDataset(['Hello world!'])
french = IterableDataset(['Bonjour le monde!'])
from fuel.streams import DataStream
streams = (DataStream(english), DataStream(french))
from fuel.transformers import Batch, Merge
from fuel.schemes import ConstantScheme
streams = tuple(Batch(stream, ConstantScheme(1)) for stream in streams) # <-- XXX
merged_stream = Merge(streams, ('english', 'french'))
next(merged_stream.get_epoch_iterator())
```

dwf:
The backtrace would be helpful...

ejls:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-10-361370a0eaf9> in <module>()
----> 1 next(merged_stream.get_epoch_iterator())

/home/ejls/.local/lib/python2.7/site-packages/six.pyc in next(self)
    533 
    534         def next(self):
--> 535             return type(self).__next__(self)
    536 
    537     callable = callable

/home/ejls/work/lisa/fuel/fuel/iterator.pyc in __next__(self)
     30             data = self.data_stream.get_data(next(self.request_iterator))
     31         else:
---> 32             data = self.data_stream.get_data()
     33         if self.as_dict:
     34             return dict(zip(self.data_stream.sources, data))

/home/ejls/work/lisa/fuel/fuel/transformers/__init__.py in get_data(self, request)
    781         return sum(
    782             (data_stream.get_data() for data_stream in self.data_streams),
--> 783             tuple())
    784 
    785 

/home/ejls/work/lisa/fuel/fuel/transformers/__init__.py in <genexpr>((data_stream,))
    780             raise ValueError
    781         return sum(
--> 782             (data_stream.get_data() for data_stream in self.data_streams),
    783             tuple())
    784 

/home/ejls/work/lisa/fuel/fuel/transformers/__init__.py in get_data(self, request)
    584         """Get data from the dataset."""
    585         if request is None:
--> 586             raise ValueError
    587         data = [[] for _ in self.sources]
    588         for i in range(request):

ValueError: 
```

Not sure if helpful, but I also have the following backtrace using Batch followed by Padding (adding the line `streams = tuple(Padding(stream, ()) for stream in streams`) :
```
---------------------------------------------------------------------------
[...]

PR #229, opened 2015-09-15
Fix two very nasty bugs in Merge

rizar:
The two bugs:

- when `data_stream.get_data()` raises `StopIteration` within is a generator expression, the raised exception is caught and generation stops. As a result instead of finishing the iteration the epoch iterator of `Merge` returned empty tuples. This caused the machine translation bug discussed [here](git@github.com:bartvm/fuel.git).

- `data.get_epoch_iterator()` did not reset the merged streams. Even after fixing the previous bug machine translation was not functional (crashed in the very beginning of the second epoch).

Both I and @orhanf get grade F for bug search, because the bug was perfectly reproducible with the latest Fuel. It was introduced by great Fuel interface overhaul #188, that is it was around since the beginning of July.

@heeyoulchoi, with the version of Fuel from this pull request everything seems to work well. This is however orthogonal to the data processing PR https://github.com/mila-udem/blocks-examples/pull/39, in which we still have to understand what @orhanf had in mind.

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -773,15 +773,16 @@ class Merge(AbstractDataStream):
             data_stream.next_epoch()
 
     def get_epoch_iterator(self, **kwargs):
-        self.next_epoch()
+        self.child_epoch_iterators = [data_stream.get_epoch_iterator()
+                                      for data_stream in self.data_streams]
         return super(Merge, self).get_epoch_iterator(**kwargs)
 
     def get_data(self, request=None):
         if request is not None:
             raise ValueError
         result = []
-        for data_stream in self.data_streams:
-            result.extend(data_stream.get_data())
+        for child_epoch_iterator in self.child_epoch_iterators:
+            result.extend(next(child_epoch_iterator))
         return tuple(result)
 
 
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -773,14 +773,16 @@ class Merge(AbstractDataStream):
             data_stream.next_epoch()
 
     def get_epoch_iterator(self, **kwargs):
+        self.next_epoch()
         return super(Merge, self).get_epoch_iterator(**kwargs)
 
     def get_data(self, request=None):
         if request is not None:
             raise ValueError
-        return sum(
-            (data_stream.get_data() for data_stream in self.data_streams),
-            tuple())
+        result = []
+        for data_stream in self.data_streams:
+            result.extend(data_stream.get_data())
+        return tuple(result)
 
 
 class BackgroundProcess(object):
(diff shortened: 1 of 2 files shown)


rizar:
@vdumoulin , @dwf , I would appreciate a quick review because these bugs render an important code example non-functional.

dwf:
Dima, check the issues page. At least one of these has an associated issue
(created by me).

rizar:
Also, a quick take-away message from this story is that generator expressions should be used with great care in Fuel, if not be deprecated in favour of list comprehensions:

```
In [1]: def f(): raise StopIteration

In [2]: [f() for i in range(3)]
---------------------------------------------------------------------------
StopIteration                             Traceback (most recent call last)
<ipython-input-2-1a557417277f> in <module>()
----> 1 [f() for i in range(3)]

<ipython-input-1-14502eea55aa> in f()
----> 1 def f(): raise StopIteration

StopIteration: 

In [3]: list(f() for i in range(3))
Out[3]: []
```

rizar:
Right @dwf, this PR fixes #202. It is also related to #201, which I haven't yet fully read, it is quite lengthy.

orhanf:
@rizar, [mila-udem/blocks-examples#39](https://github.com/mila-udem/blocks-examples/pull/39) has nothing to do with this bug, it is a fix just for smaller shuffled corpora

dwf:
Worth reading #201 as the last commit is result of a bisect I ran :)

rizar:
And sorry for spaming, but I think it is also a good time to start keeping track of the latest Blocks and Fuel versions with which a code example works.

rizar:
@orhanf, I know that #mila-udem/blocks-examples#39  did not cause the crash, but I just mentioned it to make it clear that after fixing these bugs the machine translation example will still need some work.

orhanf:
i see, okay then i'll rerun the example with the latest versions and make a pr for the fix. Also there was [this](https://github.com/mila-udem/blocks-examples/issues/32) issue  that has to be resolved as well. May be we can enhance/refactor that issue and continue accordingly. 
btw do not hesitate to assign ccw tickets to me related to these issues :)

rizar:
@dwf, indeed, you were almost there in #229! I am sorry for ignoring that ticket, it was created when I was in a paper rush, then there was a vacation, etc...

fuel/transformers/__init__.py
```diff
@@ -773,14 +773,16 @@ def next_epoch(self):
...
 
     def get_data(self, request=None):
         if request is not None:
             raise ValueError
-        return sum(
-            (data_stream.get_data() for data_stream in self.data_streams),
-            tuple())
+        result = []
+        for data_stream in self.data_streams:
+            result.extend(data_stream.get_data())
```
vdumoulin: There is a hard-coded assumption here that the data stream doesn't accept a `request` argument. This could fail e.g. if one of the merged data streams is a `Batch` instance.

rizar:
@vdumoulin , your comment in the code is correct. I have changed the code to support iteration schemes in the merged streams, can you please take a look?

vdumoulin:
Looks good to me.

PR #236, opened 2015-09-21
Add weizmann horse dataset

fvisin:
Wrapper for the Weizmann Horse Dataset


fvisin:
The class name should be fine now. 
I will rebase at the end of the review to squash the PEP8 commits in the initial commit.

PR #237, opened 2015-09-22
Update requirements, install fortran as temporary workaround

bartvm:


rizar:
The build still fails, I cleared the cache and restarted.

bartvm:
There seems to be a separate issue now, looks like this one: http://software-engineer.gatsbylee.com/python-pip-error-invalid-command-bdist_wheel/

rizar:
Restarted the build to see if updating Blocks requirements helps.

rizar:
Thank you very much @bartvm for fixing this!

Issue #222, opened 2015-08-28
Target indices in SVHN format 2 are 1-based instead of 0-based.


vdumoulin:
We should map `10` to `0`, as it's done for SVHN format 1.

vdumoulin:
@hantek This will be your ticket this week.

hantek:
I created a PR in #239

vdumoulin:
Closed via #239.

PR #239, opened 2015-09-25
Change target indices in SVHN format 2 to 0-based

hantek:
This is a fix for #222 .

### DIFF
--- a/fuel/converters/svhn.py
+++ b/fuel/converters/svhn.py
@@ -291,14 +291,17 @@ def convert_svhn_format_2(directory, output_directory,
     train_set = loadmat(os.path.join(directory, FORMAT_2_TRAIN_FILE))
     train_features = train_set['X'].transpose(3, 2, 0, 1)
     train_targets = train_set['y']
+    train_targets[train_targets == 10] = 0
 
     test_set = loadmat(os.path.join(directory, FORMAT_2_TEST_FILE))
     test_features = test_set['X'].transpose(3, 2, 0, 1)
     test_targets = test_set['y']
+    test_targets[test_targets == 10] = 0
 
     extra_set = loadmat(os.path.join(directory, FORMAT_2_EXTRA_FILE))
     extra_features = extra_set['X'].transpose(3, 2, 0, 1)
     extra_targets = extra_set['y']
+    extra_targets[extra_targets == 10] = 0
 
     data = (('train', 'features', train_features),
             ('test', 'features', test_features),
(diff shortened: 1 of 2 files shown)


fuel/converters/svhn.py
```diff
@@ -290,11 +290,11 @@ def convert_svhn_format_2(directory, output_directory,
 
     train_set = loadmat(os.path.join(directory, FORMAT_2_TRAIN_FILE))
     train_features = train_set['X'].transpose(3, 2, 0, 1)
-    train_targets = train_set['y']
+    train_targets = train_set['y'] - 1
```
vdumoulin: What we need is for `10` to be mapped to `0` while other labels are untouched. In this case, you map `1` to `0`, `2` to `1`, etc.

vdumoulin:
@hantek Thanks! We also need to change the corresponding test in `tests/test_converters.py`.

vdumoulin:
At lines 581, 585 and 589 in `tests/test_converters.py`, the call to `numpy.random.randint` uses `0` and `10` as its `low` and `high` arguments. This is an error on my part introduced when I mocked out the files: we should replace `0` with `1` and `10` with `11` so it corresponds to the range of labels in the actual files.

Could you make that change (and modify its corresponding assertion at line 647)? Afterwards we'll be ready to merge.

hantek:
Yes that is my bad... I didn't check the range of the mock dataset...

tests/test_converters.py
```diff
@@ -649,6 +649,9 @@ def test_format_2_converter(self):
             numpy.vstack([self.f2_train_targets_mock,
                           self.f2_test_targets_mock,
                           self.f2_extra_targets_mock]))
+        assert h5file['targets'][...].max() < 10
```
vdumoulin: The tests are currently failing because `self.f2_{train,test,extra}_targets_mock` contains labels ranging from 1 to 10.

The following should solve it:

```python
f2_train_targets_mock = self.f2_train_targets_mock
f2_train_targets_mock[f2_train_targets_mock == 10] = 0
f2_test_targets_mock = self.f2_test_targets_mock
f2_test_targets_mock[f2_test_targets_mock == 10] = 0
f2_extra_targets_mock = self.f2_extra_targets_mock
f2_extra_targets_mock[f2_extra_targets_mock == 10] = 0
assert_equal(
    h5file['targets'][...],
    numpy.vstack([f2_train_targets_mock,
                  f2_test_targets_mock,
                  f2_extra_targets_mock]))
```

vdumoulin:
@hantek Thanks! I'm merging.

PR #243, opened 2015-10-01
Added one hot encoding to transformers

markusnagel:
We added a OneHotEncoding class that transforms a given source (in int) to one-hot-encoding (also known as 1-of-K encoding) as well as a corresponding test.
One-hot-encoding is useful if (1) a problem has many classes and thus by storing the class number space can be saved and (2) if a problem can be treated as different types of problems (e.g. ordinal regression, classification).


fuel/transformers/__init__.py
```diff
@@ -928,3 +928,50 @@ def __init__(self, data_stream, sources, **kwargs):
...
+    ----------
+    data_stream : :class:`DataStream` or :class:`Transformer`.
+        The data stream.
+    num_classes : int
+        The number of classes.
+    source_name : str, default 'targets'
+        The name of the source that will be transformed.
+
+    """
+    def __init__(self, data_stream, num_classes, source_name='targets',
```
vdumoulin: You don't need an additional `source_name` kwarg, as `which_sources` will already filter out the sources to be transformed.

In fact, I think having a `targets` default value might not be worth it as well.

fuel/transformers/__init__.py
```diff
@@ -928,3 +928,50 @@ def __init__(self, data_stream, sources, **kwargs):
...
+                 **kwargs):
+        self.num_classes = num_classes
+        self.source_name = source_name
+
+        super(OneHotEncoding, self).__init__(
+            data_stream, data_stream.produces_examples, **kwargs)
+
+    def transform_source_example(self, source_example, source_name):
+        if source_name == self.source_name:
+            assert source_example <= self.num_classes
```
vdumoulin: Instead of an assertion, it'd be preferable to raise an informative `ValueError`.

fuel/transformers/__init__.py
```diff
@@ -928,3 +928,50 @@ def __init__(self, data_stream, sources, **kwargs):
...
+            output = numpy.zeros((1, self.num_classes))
+            output[0, source_example] = 1
+            return output
+        else:
+            return source_example
+
+    def transform_source_batch(self, source_batch, source_name):
+        if source_name == self.source_name:
+            assert numpy.max(source_batch) < self.num_classes
```
vdumoulin: Likewise.

vdumoulin:
@markusnagel sorry about the delay, and thanks a lot for your contribution!

markusnagel:
Thanks for your feedback. Indeed the source_name was completely unnecessary. I implemented your suggestions, what do you think of it?

aukejw:
@vdumoulin , I squashed the corrections into one commit. Any idea what's going on with travis?

vdumoulin:
No idea. I restarted the build yesterday and had the same error, I'm trying again today.

PR #245, opened 2015-10-06
Fix missing packages preventing RTD to build

vdumoulin:


vdumoulin:
@rizar do you mind having a quick look?

bartvm:
Note that RTD doesn't allow the compilation of any C code on their servers, so I'm not sure it will be happy with you trying to install h5py and pyzmq. You might need to mock them instead: https://read-the-docs.readthedocs.org/en/latest/faq.html#i-get-import-errors-on-libraries-that-depend-on-c-modules

vdumoulin:
I'll have a look, thanks!

dwf:
Do RTD support conda yet?

vdumoulin:
It's been requested, but not yet: https://github.com/rtfd/readthedocs.org/issues/857.

vdumoulin:
I'll merge to test @bartvm's suggested solution.

PR #247, opened 2015-10-07
RTD/installation fixes

bartvm:
Went down the rabbit hole of RTD, ended up cleaning the installation process a bit. Basically, when we allow the new Sphinx to run, it can manage to build docs without all the mocking stuff (great!)

Otherwise, brought a whole bunch of dependencies up to date. Note that because of a bug in RTD https://github.com/rtfd/readthedocs.org/issues/1621 it will try to install from `requirements.txt` even when told not to... This breaks stuff, so I renamed `requirements.txt` to `req.txt` for now.

Lastly, currently any attempt at installing gives an error if Cython isn't present. Following the official guidelines (http://docs.cython.org/src/reference/compilation.html#distributing-cython-modules) I think we should distribute the `*.c` file instead; just need to make sure that if someone edits the Cython files, they run `cythonize` before committing, shouldn't be a big deal.

RTD runs, now let's see what Travis says.


bartvm:
All tests pass (https://travis-ci.org/mila-udem/fuel/builds/84013707). Reverting the last commit that was using Blocks' `rtd` branch so that this can be merged. Needs to be merged simultaneously with https://github.com/mila-udem/blocks/pull/861.

req-travis-pip.txt
```diff
@@ -1,3 +1,4 @@
 nose2[coverage-plugin]==0.5.0
 coveralls==1.0
-wheel==0.26.0
+picklable-itertools==0.1.1
+progressbar2==2.7.3
```
vdumoulin: I'm a bit confused, why is `progressbar2` here and in `req.txt` at the same time?

vdumoulin:
Great! The mocking stuff didn't work anyways: for a reason which I haven't investigated yet, the RTD build currently crashes under an infinite recursion.

req-travis-pip.txt
vdumoulin: Nevermind: `req.txt` is for users, whereas `req-travis-{pip,conda}.txt` is for Travis.

vdumoulin:
LGTM. I'll wait for @dmitriy-serdyuk to give his OK on the blocks side, and then we can merge.

bartvm:
Great!

The infinite recursion is actually because the current line reads `from mock import MagicMock` instead of `from mock import Mock as MagicMock`.

vdumoulin:
TIL I can't read **Tip** sections.

dmitriy-serdyuk:
@vdumoulin , seems everything should work now, let's try to merge.

PR #249, opened 2015-10-07
Add mocking back in

bartvm:
Analogous to https://github.com/mila-udem/blocks/pull/864

RTD won't compile C extensions, so the importing of the Cython module needs to be guarded (good practice anyway I guess, so that failing to compile the Cython extension doesn't render the other transformers that don't rely on it useless)


bartvm:
https://fuel.readthedocs.org/en/rtd2/

bartvm:
Tests pass, it's just the Blocks tests fail as long as https://github.com/mila-udem/blocks/pull/864 isn't merged.

dmitriy-serdyuk:
Can someone restart the tests?

PR #254, opened 2015-10-19
Simpler version management

rizar:
https://github.com/mila-udem/blocks/issues/863

Version is defined in a very convoluted in Fuel, and I would like to change it before the release.

### DIFF
--- a/fuel/__init__.py
+++ b/fuel/__init__.py
@@ -1,3 +1,3 @@
-from fuel.config_parser import config
+from fuel.config_parser import config # noqa
 
 __version__ = '0.0.1'
--- a/fuel/__init__.py
+++ b/fuel/__init__.py
@@ -1,15 +1,1 @@
-import os.path
-from pkg_resources import get_distribution, DistributionNotFound
-
-from fuel.config_parser import config  # noqa
-
-try:
-    DIST = get_distribution('fuel')
-    DIST_LOC = os.path.normcase(DIST.location)
-    HERE = os.path.normcase(__file__)
-    if not HERE.startswith(os.path.join(DIST_LOC, 'fuel')):
-        raise DistributionNotFound
-except DistributionNotFound:
-    __version__ = 'not installed'
-else:
-    __version__ = DIST.version
+__version__ = '0.0.1'
--- a/fuel/__init__.py
+++ b/fuel/__init__.py
@@ -1,1 +1,3 @@
+from fuel.config_parser import config
+
 __version__ = '0.0.1'
--- a/fuel/__init__.py
+++ b/fuel/__init__.py
@@ -1,3 +1,3 @@
-from fuel.config_parser import config # noqa
+from fuel.config_parser import config  # noqa
 
 __version__ = '0.0.1'
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,8 @@ from setuptools import find_packages, setup
 from Cython.Build import cythonize
 from distutils.extension import Extension
 
+import fuel
+
 HERE = path.abspath(path.dirname(__file__))
 
 with open(path.join(HERE, 'README.rst')) as f:
@@ -19,7 +21,7 @@ else:
 
 setup(
     name='fuel',
-    version='0.0.1',  # PEP 440 compliant
+    version=fuel.__version__,  # PEP 440 compliant
     description='Data pipeline framework for machine learning',
     long_description=LONG_DESCRIPTION,
     url='https://github.com/mila-udem/fuel.git',


rizar:
Tests pass, a quick review would be appreciated.

bartvm:
This makes no sense... It imports Fuel before ever installing it; you can't rely on Fuel being importable in `setup.py`. As expected, installing Fuel in a clean environment now fails:

```bash
(fuel_install)bartvm-mbp:fuel bartvm$ python setup.py install
  File "setup.py", line 7, in <module>
    import fuel
  File "/Users/bartvm/fuel/fuel/__init__.py", line 1, in <module>
    from fuel.config_parser import config  # noqa
  File "/Users/bartvm/fuel/fuel/config_parser.py", line 74, in <module>
    import six
ImportError: No module named six
```
The reason it is done "very convoluted" is because there is chicken and egg problem here: You preferably define the version of the package in only one place, this needs to be `setup.py` (because nothing else is importable at this point). Then, however, you need to a bit of trickery to recover the version that was passed to `setup.py` in order to set `fuel.__version__` in `__init__.py`. At this point you need to deal with some corner cases: Fuel being imported without ever having been installed, or the current imported installation not being the same as the installed one. Those are the checks being done.

If you don't like that approach, the alternative is just to define the version twice, once in `__init__.py` and once in `setup.py`, and just make sure they stay in sync.

rizar:
Okay, but how is it different from what the way it is done in Blocks (
https://github.com/mila-udem/blocks/blob/master/setup.py)?

bartvm:
I clearly wasn't thinking when committing https://github.com/mila-udem/blocks/commit/d0a79694d840a08b740b101f41130813b55fc3be, but the import of Blocks inside `setup.py` should be fixed there as well. It could work in cases where (a) the current directory is in the `PYTHON_PATH` and (b) `__init__.py` doesn't rely on the importing of any other modules, but both Blocks and Fuel read their configuration on import, which requires libraries like Six and PyYAML.

A list of alternative approaches can be found here: https://python-packaging-user-guide.readthedocs.org/en/latest/single_source_version/ I was using an extended version of (5), but (1) and (3) could also be reasonable options (that said, I would just stick with the current one).

rizar:
Hold on, but nobody complains on Blocks installation problems. That's probably because the current directory is a place for module search, so `import blocks` should work fine. Another thing is that importing config is actually a pretty bad idea in this case. Can you please explain how `from blocks import config` works even though `config` is not imported in `blocks/__init__.py`? I see that a `config` object is created in the global namespace in `blocks/config.py`, but I can not understand why that helps.

bartvm:
`from blocks import config` works, but doesn't do what you want it to do. The Blocks equivalent to Fuel's `from fuel import config` is `from blocks.config import config`. This was changed in https://github.com/mila-udem/blocks/commit/d0a79694d840a08b740b101f41130813b55fc3be. In Blocks, we don't rely on `__init__.py` importing anything (which means that `setup.py` can get away with importing Blocks).  The reason we're using this approach is because it enabled namespace packages (namespace packages don't like it if there is any code in `__init__.py`). The downside of this approach is that the configuration doesn't get parsed until a file does `from blocks.config import config`. This makes it possible to import Blocks even if there are errors in your configuration files or flags (which can be annoying in interactive sessions especially). Secondly, it means it's not possible to use the slightly cleaner `from blocks import config` syntax. Because Fuel didn't need namespace packages, I didn't change this for Fuel.

rizar:
Wow, I swear I merged this PR by an accident! Sorry! I have just noticed that it is in the list of closed PRs...

Thanks for your explanations! Now I understand why we have what we have in `version.py` in Blocks and Fuel. 

Out of the methods discussed at the page you referred to I would choose 4 for its simplicity. `pkg_resources` is not very well documented and and code using might be intimidating for many people, me included. I will prepare a PR now.

rizar:
Now I do not understand why setting `__version__` in `blocks/__init__.py` works, even though according to [this page](https://pythonhosted.org/setuptools/setuptools.html#namespace-packages)

```
You must NOT include any other code and data in a namespace package’s __init__.py.
 Even though it may appear to work during development, or when projects are installed as 
.egg files, it will not work when the projects are installed using “system” packaging tools –
 in such cases the __init__.py files will not be installed, let alone executed.
```

(sighs)

bartvm:
When they refer to "system packaging tools" they mostly mean things like `apt-get`, `yum`, etc. When packages are created for those tools, `__init__.py` files are sometimes stripped. Conda and `pip` tend to use `.egg` files or will copy all files.

There's another reason why setting `__version__` in `__init__.py` is dangerous in namespace packages: The `__init__.py` file can be overridden by any other `__init__.py` file from a namespace package. So when `blocks-extras` gets installed, the `__init__.py` file we have in Blocks could be overridden by the [this one](https://github.com/mila-udem/blocks-extras/blob/master/blocks/__init__.py) from `blocks-extras` instead. Whether this happens is up to the package manager.

In short: Package management in Python is an absolute mess.

rizar:
Having all that said, do you still mind renaming `blocks.extras` into `blocks-extras` and getting rid of namespace packages?

bartvm:
I don't have a big issue with getting rid of namespace packages. (`blocks_extras` by the way, since hyphens aren't allowed in package names.)

rizar:
All right, then the question is should we do it before 0.1 (which I was
going to release right after cleaning up version management) or after.

A good argument to do it after is to let people who now have projects using
Blocks and Blocks-extras base them on the stable version + the last version
of blocks_extras that supported namespace packages.

PR #255, opened 2015-10-19
Caltech101

abergeron:
This is a converter and a loader for caltech101, which could be useful for some other people.

It might lack a bit of the polish that other datasets have, but it could be useful, at least as a starting point.

There is no downloader and I couldn't figure out how to get scipy to read from an open file.

If you have ideas on how to improve that don't necessitate a long time investment, I can probably do it.  Otherwise, this is mostly to avoid duplication of effort in case someone else is working on this.

### DIFF
--- a/fuel/converters/caltech101.py
+++ b/fuel/converters/caltech101.py
@@ -1,8 +1,8 @@
 import os
-import tarfile
 import h5py
 
-from scipy.misc import imread
+import numpy
+import scipy.misc
 
 from fuel.converters.base import fill_hdf5_file, MissingInputFiles
 
@@ -110,6 +110,7 @@ CATEGORIES = (
     'scissors',
     'ibis')
 
+
 def convert_silhouettes(directory, output_directory,
                         output_file=None):
     """ Convert the CalTech 101 Datasets.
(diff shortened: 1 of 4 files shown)


fuel/converters/caltech101.py
```diff
@@ -0,0 +1,196 @@
...
+import h5py
+
+import numpy
+import scipy.misc
+
+from six.moves import range
+
+from fuel.converters.base import fill_hdf5_file, MissingInputFiles
+
+CATEGORIES = (
```
vdumoulin: 101 lines just for the `CATEGORIES` tuple is quite long, could we condense that?

fuel/converters/caltech101.py
```diff
@@ -0,0 +1,196 @@
...
+    'stegosaurus',
+    'tick',
+    'scissors',
+    'ibis')
+
+NUM_TRAIN = 20
+NUM_TEST = 10
+
+
+def read_image(imfile):
```
vdumoulin: I'm not very fond of resizing the images to 256x256, as it hard-codes part of the pre-processing right into the converter. I think we should leave them as variable-length data and instead make the resizing part of the dataset's default transformer.

fuel/converters/caltech101.py
```diff
@@ -0,0 +1,196 @@
...
+    im = scipy.misc.imresize(scipy.misc.imread(imfile), (256, 256))
+    if im.ndim == 2:
+        return im.reshape(1, 256, 256)
+    else:
+        return numpy.rollaxis(im, 2, 0)
+
+
+def convert_silhouettes(directory, output_directory,
```
vdumoulin: `convert_silhouettes -> convert_caltech101`?

vdumoulin:
I think having a downloader would be very helpful for the dataset.

fuel/converters/caltech101.py
abergeron: There isn't any transformer that resizes images currently.  Unless I'm wrong.

vdumoulin:
@abergeron I'm doing a cleanup of the open PRs. Would you still like to go ahead with this one?

abergeron:
Is there much to do?

fuel/converters/caltech101.py
vdumoulin: There is now: [MinimumImageDimensions and RandomFixedSizeCrop](https://github.com/mila-udem/fuel/blob/master/fuel/transformers/image.py#L105-L272).

vdumoulin:
Aside from my inline comments, we would need a downloader for the dataset, and we would also need to make this a variable-length dataset.

I think you could reuse a lot of code from the [Imagenet converter](https://github.com/mila-udem/fuel/blob/master/fuel/converters/ilsvrc2010.py) if you encode images as a variable-length vector of raw bytes.

MartinThoma:
You might find this helpful: https://github.com/MartinThoma/algorithms/blob/master/ML/datasets/caltech101.py#L22-L37

PR #256, opened 2015-10-21
Single version for Fuel in an easy way

rizar:
Fixes the consequences of #254


setup.py
```diff
@@ -18,9 +16,13 @@
 else:
     extra_compile_args = []
 
+exec_results = {}
+exec(open(path.join(path.dirname(__file__), 'fuel/version.py')), exec_results)
```
bartvm: Nitpick, but I prefer using `with` instead of relying on the GC to close this file handle.

docs/conf.py
```diff
@@ -92,9 +92,10 @@ def __getattr__(cls, name):
 # built documents.
 #
 # The short X.Y version.
-version = '0.0'
+import fuel
+version = fuel.__version__[:fuel.__version__.rindex('.')]
```
bartvm: Again, tiny nitpick, but if you don't want to assume that there are no more than 3 dots:
```
v[::-1].split('.', 1)[-1][::-1]
```
or
```
re.match(r'(.*?\..*?)\.', v).expand(r'\1')
```

setup.py
rizar: Makes sense, done.

docs/conf.py
rizar: Check out yet another way to do it in the code.

bartvm: But 10 characters longer *and* O(n)! :open_mouth: 

Kidding aside, LGTM.

rizar: Sorry, I did try to find an exponential algorithm for that, but this is all I could do...

PR #266, opened 2015-10-30
WIP: Added a transformer to segment sequences. This is useful to apply tbptt.

sotelo:
I coded this transformer to segment sequences. I think it can be useful for people working with RNNs. The way I use it is to segment sequences in parts and so the gradient will not have to be propagated all the way back. You can also reuse the (final) states to initialize the initial_states to have some information about the past.

Please let me know if this is useful enough to be included in fuel. If that's the case, I'll clean up the code.

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -928,3 +928,77 @@ class FilterSources(AgnosticTransformer):
     def transform_any(self, data):
         return [d for d, s in izip(data, self.data_stream.sources)
                 if s in self.sources]
+
+class SegmentSequence(Transformer):
+    """Segments the sequences in a batch.
+    This transformer is useful to do tbptt. All the sequences to segment
+    should have the time dimension as their first dimension.
+    
+    Parameters
+    ----------
+    data_stream : instance of :class:`DataStream`
+        The wrapped data stream.
+    seq_size : int
+        maximum size of the resulting sequences.
+    which_sources : tuple of str, optional
+        sequences to segment
+    add_flag : bool, optional
+        add a flag indicating the beginning of a new sequence.
+    flag_name : str, optional
+        name of the source for the flag
+        
+    """
+    def __init__(self, data_stream,seq_size=100,which_sources=None,
+                 add_flag=False, flag_name = None, **kwargs):
+        super(SegmentSequence, self).__init__(data_stream=data_stream,
+            produces_examples=data_stream.produces_examples,**kwargs)
+
+        if which_sources is None:
+            which_sources = data_stream.sources
+        self.which_sources = which_sources
+
+        self.seq_size = seq_size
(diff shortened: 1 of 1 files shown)


rizar:
In my opinion it would be a valuable contribution to Fuel. A bit of work would be however required to make the docstring compliant with our standard. It should explain clearly what the transformer does without assuming that the reader know what TBPTT abbreviation means. That said, it would be still useful to mention it as the usecase in mind.

sotelo:
Great to know! I should have a better version by the end of this week.

vdumoulin:
@sotelo I'm doing a cleanup of the open PRs. Would you still like to go ahead with this one?

sotelo:
Yes. I forgot to finish it a while ago. I will update it today.

sotelo:
SORRY for the delay! I got distracted by other things and finally I am submitting some improvements for this. I've improved the documentation a bit and please let me know if you find it satisfactory. Also, I guess I am missing an example although I'm not really sure if I should do it with one of the existing datasets or create a synthetic one. Suggestions?

I am studying for technical interview, so I am still a bit busy but as soon as that is finished and this is also finished I will turn into the basic extension documentation.

bartvm:
I missed this earlier, but I think this overlaps a lot with #310...

sotelo:
Yes! It's very similar. There are a few differences though... The other one seems to be more for when you have only 1 large sequence. I'm using this one for instance for handwriting, where you have a lot of different length sequences and maybe you want to use tbptt.

bartvm:
The window one I wrote should do the same thing and can take multiple,
variable length sequences (it was actually meant for TBTT on audio).

sotelo:
I see now. Well, then I guess it's the same. I'll close the PR.

bartvm:
The `add_flag` and `share_value` are missing from my implementation though

bartvm:
Had another look and I think the implementations do something quite different. In my version, it takes e.g. `[1, 2, 3, 4, 5]` and turns it into e.g. `[1, 2, 3]`, `[2, 3, 4]` with targets `[4]` and `[5]`. Your version seems to take `[1, 2, 3, 4, 5]` and return `[1, 2]`, `[3, 4]`, `[5]`.

Basically, my transformer has two sliding windows (one becomes the source, the other the target) of fixed size, which slide with a step size of 1. Your version only has 1 window which has a step size equal to the window size, but the window size isn't fixed (and can be smaller for the end of the sequence).

I feel like they overlap enough though to see if we can combine all the functionality into one transformer, which basically has one or two sliding windows, a configurable step size, and configurable boundary conditions.

sotelo:
Yes! I did it like that because I mostly use it for bpttt. Maybe we can talk next week for what should be the more complete transformer and I'll be happy to make the changes.

PR #270, opened 2015-12-02
Fix Travis miniconda.

dwf:


vdumoulin:
Looks good, merge at will.

dwf:
Merging before blocks tests complete since we've already run the blocks tests against this version of Fuel many times.

PR #271, opened 2015-12-03
Added Random2DRotation transformer and corresponding test.

aukejw:
I haven't looked into Cython implementations for 2D image rotation yet, so this implementation is still (very) slow. Furthermore, I currently use `scipy.ndimage.interpolation.rotate` for rotation, which (as far as I know) can only rotate per-image. As such, `transform_source_batch` simply calls `transform_source_example` for every image in the batch.

Timeit example in which we rotate 3x256x256 images: 
```python
>>> import timeit                                                                                                                                                
>>> setup = """                                                                     
... from collections import OrderedDict                                             
... import numpy                                                                    
... from fuel.datasets.base import IndexableDataset                                 
... from fuel.schemes import SequentialScheme, SequentialExampleScheme              
... from fuel.streams import DataStream                                             
... from fuel.transformers.image import Random2DRotation                            
...                                                                              
... batch = numpy.zeros((100, 3, 256, 256), dtype=int)                              
... img = numpy.arange(3*256*256, dtype=int).reshape(3, 256, 256)                   
... batch[:] = img                                                                  
...                                                                                 
... data = OrderedDict([('image', batch)])                                          
... dataset = IndexableDataset(                                                     
...     data, axis_labels={'image': ('batch', 'channel', 'height', 'width')})       
... stream1 = DataStream(dataset, iteration_scheme=SequentialExampleScheme(1))      
... stream1 = Random2DRotation(stream1, which_sources=('image',))                   
... stream2 = DataStream(dataset, iteration_scheme=SequentialScheme(100, 100))      
... stream2 = Random2DRotation(stream2, which_sources=('image',))                   
... """                                                                             
>>> main_statement = """                                                            
... stream1.transform_source_example(img, source_name='image') """      
>>> # example by example                                                                       
>>> t = timeit.Timer(main_statement, setup)                                         
>>> print t.repeat(5, 1000)                                                         
[38.28747606277466, 38.34248900413513, 38.31103992462158, 38.42443799972534, 38.43785381317139]                                                                                 
>>> # equivalent, but per batch                                                                       
>>> main_statement = """                                                            
... stream2.transform_source_batch(batch, source_name='image') """
>>> t = timeit.Timer(main_statement, setup)                                         
>>> print t.repeat(5, 10)                  
[37.95794892311096, 37.86270499229431, 37.8510000705719, 37.841578006744385, 37.87645411491394]
```

Lastly, any pointers on output correctness checks are greatly appreciated! I haven't seen many tests for which the expected output is a hard-coded array, so if this is regarded as undesirable or just plain ugly, I'm more than willing to change it to something more readable.  

I realize that in its current state, this transformer is probably not ready to be merged just yet. Still, random rotation is (in my opinion) a useful transformer for data augmentation, so I hope this PR generates some discussion on its implementation.

### DIFF
--- a/fuel/transformers/image.py
+++ b/fuel/transformers/image.py
@@ -270,3 +270,100 @@ class RandomFixedSizeCrop(SourcewiseTransformer, ExpectsAxisLabels):
             off_w = 0
         return example[:, off_h:off_h + windowed_height,
                        off_w:off_w + windowed_width]
+
+
+class Random2DRotation(SourcewiseTransformer, ExpectsAxisLabels):
+    """Randomly rotate 2D images in the spatial plane.
+
+    Parameters
+    ----------
+    data_stream : :class:`AbstractDataStream`
+        The data stream to wrap.
+    maximum_rotation : float, default `math.pi`
+        Maximum amount of rotation in radians. The image will be rotated by
+        an angle in the range [-maximum_rotation, maximum_rotation].
+    resample : str, optional
+        Resampling filter for PIL to use to upsample any images requiring
+        it. Options include 'nearest' (default), 'bilinear', and 'bicubic'.
+        See the PIL documentation for more detailed information.
+
+    Notes
+    -----
+    This transformer expects to act on stream sources which provide one of
+
+     * Single images represented as 3-dimensional ndarrays, with layout
+       `(channel, height, width)`.
+     * Batches of images represented as lists of 3-dimensional ndarrays,
+       possibly of different shapes (i.e. images of differing
+       heights/widths).
+     * Batches of images represented as 4-dimensional ndarrays, with
+       layout `(batch, channel, height, width)`.
+
+    The format of the stream will be un-altered, i.e. if lists are
+    yielded by `data_stream` then lists will be yielded by this
+    transformer.
+
+    """
+    def __init__(self, data_stream, maximum_rotation=math.pi,
+                 resample='nearest', **kwargs):
+        if maximum_rotation <= 0 or maximum_rotation > math.pi:
+            raise ValueError('maximum_rotation ({:.5f}) must be in the range '
+                             '(0, math.pi]'.format(maximum_rotation))
+        self.maximum_rotation = numpy.rad2deg(maximum_rotation)
+        try:
+            self.resample = getattr(Image, resample.upper())
+        except AttributeError:
+            raise ValueError("unknown resampling filter '{}'".format(resample))
+
+        self.rng = kwargs.pop('rng', None)
+        self.warned_axis_labels = False
+        if self.rng is None:
+            self.rng = numpy.random.RandomState(config.default_seed)
+        kwargs.setdefault('produces_examples', data_stream.produces_examples)
+        kwargs.setdefault('axis_labels', data_stream.axis_labels)
+        super(Random2DRotation, self).__init__(data_stream, **kwargs)
+
+    def transform_source_batch(self, source, source_name):
+        self.verify_axis_labels(('batch', 'channel', 'height', 'width'),
+                                self.data_stream.axis_labels[source_name],
+                                source_name)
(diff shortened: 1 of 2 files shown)


vdumoulin:
@aukejw Very sorry for the delay! I'll try to have a look at that by the end of the week.

bartvm:
Overall the code looks good to me. I don't see an issue with the hardcoded arrays; I can't think of another good way to test things.

The current code loops over the channels in order to rotate each one independently, you don't need to do that. The `rotate` command takes an `axes` argument (which defaults to `(1, 0)`) which you can just set to `(2, 1)` instead in order to rotate the entire array in one go (it's not any faster though).

I also benchmarked Pillow's rotate method, and that looks promising (and it's already a dependency). It would be nice if the resampling method could be an argument to the transformer so that the user can choose the quality/speed trade-off.

```ipython
In [118]: import numpy

In [119]: import scipy

In [120]: from PIL import Image

In [121]: image = numpy.random.randint(0, 256, size=(3, 256, 256)).astype('uint8')

In [122]: %timeit scipy.ndimage.interpolation.rotate(image, numpy.pi / 2, reshape=False, axes=(2, 1))
10 loops, best of 3: 23.1 ms per loop

In [123]: %timeit numpy.array(Image.fromarray(image.transpose((1, 2, 0))).rotate(45, resample=Image.BICUBIC))
100 loops, best of 3: 5.88 ms per loop

In [124]: %timeit numpy.array(Image.fromarray(image.transpose((1, 2, 0))).rotate(45))
1000 loops, best of 3: 1.26 ms per loop
```

dwf:
API adaptation trick: you can use "gettattr(Image, resample.upper())" in
some places in order to support using a nicer string argument of "bicubic",
"nearest", "linear", etc. rather than have the user know the byzantine
module-level constants.

>

aukejw:
@bartvm : Thanks for the feedback. You're right, looping over channels is unnecessary. If Image.rotate turns out to be the better option, I'll replace the scipy rotate call, and add the resampling method as argument. 

@dwf Agreed.

aukejw:
I took inspiration for the `resample` argument from the `MinimumImageDimensions` transformer constructor -- although this transformer seems [not to use the argument](https://github.com/mila-udem/fuel/blob/master/fuel/transformers/image.py#L169) (?)

Repeat of the timeits shows that the speedup when using `Image.rotate` is considerable:
```
>>> # example by example                                                                       
>>> t = timeit.Timer(main_statement, setup)                                         
>>> print t.repeat(5, 1000)
[27.805537939071655, 28.955289125442505, 29.08978796005249, 29.47601008415222, 29.370649099349976]
>>> # equivalent, but per batch                                                                       
>>> main_statement = """                                                            
... stream2.transform_source_batch(batch, source_name='image') """
>>> t = timeit.Timer(main_statement, setup)                                         
>>> print t.repeat(5, 10)
[28.193249940872192, 27.960711002349854, 27.922738075256348, 28.211444854736328, 28.142256021499634]
```

I could make the transformer a bit more flexible by adding an auxiliary function (like in `MinimumImageDimensions`) that performs the rotation, rather than having `transform_source_example` do the work.

bartvm:
One of those Travis errors should be fixed by https://github.com/mila-udem/fuel/pull/286 (it's a mysterious non-deterministic error), the other one is new but my guess is that it's just a Travis hiccup. I'll try clearing the Travis cache otherwise.

Seems like you're right, the `resample` argument somehow got looked over in #196. (@dwf)

Having a separate method do the rotation would be slightly cleaner perhaps, but I think it would be fine to merge either way.

aukejw:
Actually, `tranform_source_example` triggered unnecessary `verify_axis_labels` checks, so the auxiliary. function is IMO necessary. This also allows me to generate the rotation angles for all items in the batch simultaneously. 

If you want me to squash some commits, let me know.

bartvm:
Squashing everything would be great. Otherwise LGTM.

aukejw:
Done.

PR #272, opened 2015-12-04
Fixed size crops

markusnagel:
I added a transformer that performs a fixed size crop. It is similar to the already existing RandomFixedSizeCrop transformer. One key difference is that the user can specify the crop with a relative location instead of taking a random crop. This is useful for example for predicting on imagenet where often a center crop of the image is used or a center crop plus the 4 corner crops.


vdumoulin:
@markusnagel Very sorry for the delay! I'll try to have a look at it by the end of the week.

fuel/transformers/image.py
```diff
@@ -270,3 +270,99 @@ def transform_source_example(self, example, source_name):
             off_w = 0
         return example[:, off_h:off_h + windowed_height,
                        off_w:off_w + windowed_width]
+
+
+class FixedSizeCrop(SourcewiseTransformer, ExpectsAxisLabels):
+    """Crop images to a fixed window size.
```
bartvm: Sections in the docstring should have newlines.

fuel/transformers/image.py
```diff
@@ -270,3 +270,99 @@ def transform_source_example(self, example, source_name):
...
+     * Single images represented as 3-dimensional ndarrays, with layout
+       `(channel, height, width)`.
+     * Batches of images represented as lists of 3-dimensional ndarrays,
+       possibly of different shapes (i.e. images of differing
+       heights/widths).
+     * Batches of images represented as 4-dimensional ndarrays, with
+       layout `(batch, channel, height, width)`.
+    The format of the stream will be un-altered, i.e. if lists are
+    yielded by `data_stream` then lists will be yielded by this
+    transformer.
```
vdumoulin: There should also be a blanck line before the closing set of quotes.

vdumoulin:
There seems to be quite a lot of overlap between `FixedSizeCrop` and `RandomFixedSizeCrop`. You could make `FixedSizeCrop` inherit from `RandomFixedSizeCrop` and factor cropping into `_crop_batch` and `_crop_example` methods which accept `off_h` and `off_w` arguments.

aukejw:
The FixedSizeCrop transformer will AFAIK not benefit from the Cython implementation  `window_batch_bchw`. The same offset is used for all images in the batch, so you can simply use
```
return source[:, :,
              off_h:off_h + windowed_height,
              off_w:off_w + windowed_width]
```
rather than first constructing arrays `offsets_h` and `offsets_w` of the same length as the batch, then invoking `window_batch_bchw`. 

I haven't tested whether it actually makes a difference in terms of speed, but it seems like overkill to use `window_batch_bchw` here.

PR #274, opened 2015-12-13
make all the source to be padded

lixiangnlp:
I think all the source in the data should be padded, but some source need not mask.


vdumoulin:
@bartvm I think you're more familiar than I am with this transformer. Would you mind having a look?

bartvm:
As it stands, I don't think this is a sensible change to make. Forcing each source to be padded, only giving the option to leave out the mask, would fail if any of the data sources are non-numerical for example. The current behaviour is technically as flexible as can be: Either the source is left alone, or it is padded and given a mask. If you don't need the masks, simply discard them later on.

I'll close this for now, but am of course open to hearing why @lixiangnlp thinks this change is needed.

Issue #275, opened 2015-12-14
intall via PyPI


dribnet:
Any objection to having the latest official release listed on the Python Package Index (PyPI)? I think it would help adoption to support

```bash
pip install fuel
```

rizar:
+1

dmitriy-serdyuk:
And seems, that this name is available, we should hurry before someone else takes it.

PR #279, opened 2016-01-02
repo cleanup related to releases

dribnet:
I did some housekeeping I thought it might be useful to offer back:

 1. Merged in previous releases to master so that they would be upstream of master
 2. Updated setup.py and added setup.cfg to support PyPI release (#275)
 3. Added git annotated tags for releases since the repo had only lightweight tags

Note that any of the above can be taken independently.

for 1 and 2, merge this pull request
for 1 only, merge 1f22757
for 2 only, cherry-pick af9488c
3 is independent of 1,2 but could be done by fetching and pushing my repo's annotated tags

Just wanted to offer this work back - not offended if you decide to close and not pull this into master.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,11 +23,11 @@ version = exec_results['version']
 
 setup(
     name='fuel',
-    version='0.1.1',
+    version=version,  # PEP 440 compliant
     description='Data pipeline framework for machine learning',
     long_description=LONG_DESCRIPTION,
     url='https://github.com/mila-udem/fuel.git',
-    downlaod_url='https://github.com/mila-udem/fuel/archive/v0.1.1.tar.gz',
+    download_url='https://github.com/mila-udem/fuel/tarball/v' + version,
     author='Universite de Montreal',
     license='MIT',
     # See https://pypi.python.org/pypi?%3Aaction=list_classifiers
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -23,10 +23,11 @@
 
 setup(
     name='fuel',
-    version=version,  # PEP 440 compliant
+    version='0.1.1',
     description='Data pipeline framework for machine learning',
     long_description=LONG_DESCRIPTION,
     url='https://github.com/mila-udem/fuel.git',
+    downlaod_url='https://github.com/mila-udem/fuel/archive/v0.1.1.tar.gz',
```
bartvm: Looks like a typo

setup.py
```diff
@@ -23,10 +23,11 @@
 
 setup(
     name='fuel',
-    version=version,  # PEP 440 compliant
+    version='0.1.1',
```
bartvm: Why not use the `version` variable? This is done in order to avoid having the version defined in two separate places (`setup.py` and `fuel/__init__.py`).

vdumoulin:
@bartvm are you still interested in handling the reviewing of this PR?

bartvm:
In principle I'm fine with this PR, it's just those two comments that I made that need to be addressed.

PR #280, opened 2016-01-12
Move basicConfig call to start_server

rizar:
`logging.basicConfig` call should be done explicitly by the user, because it can only be done once in the lifetime of the program. Before this PR it was implicitly invoked at `from fuel.streams import DataStream`.


vdumoulin:
LGTM, thanks!

PR #281, opened 2016-01-19
Add description for `default_transformers` attribute

johnarevalo:
taken from https://github.com/mila-udem/summerschool2015/blob/master/fuel_tutorial/fuel_tutorial.ipynb


vdumoulin:
Looks good, thanks!

PR #282, opened 2016-01-20
H5PYDataset to support get_example_stream

rizar:
Looks like a very simple PR, a quick review would be appreciated. Replaces #179

### DIFF
--- a/fuel/datasets/base.py
+++ b/fuel/datasets/base.py
@@ -45,7 +45,8 @@ class Dataset(object):
         elements:
             - the Transformer subclass to apply,
             - a list of arguments to pass to the subclass constructor, and
-            - a dict of keyword arguments to pass to the subclass constructor.
+            - a dict of keyword arguments to pass to the subclass
+              constructor.
 
 
     Notes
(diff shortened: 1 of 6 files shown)


rizar:
Also fixes a few syntax errors that somehow were present in master.

rizar:
@vdumoulin , doctests are somehow crashing for this pull request, which I find quite strange. Can you please comment on this test run: https://travis-ci.org/mila-udem/fuel/jobs/103674824 ?

vdumoulin:
I confirm that the tests are failing locally for me as well.

vdumoulin:
I found the source of the failure. If you move setting the `example_iteration_scheme` attribute after the superclass constructor call, the tests will pass.

This is a nasty issue: accessing the `num_examples` property causes the instance to access its `subsets` attribute, which is lazy-loaded. This causes `H5PYDataset.load` to be called *before* the `sources` attribute has been set, and then all data sources are loaded into memory as opposed to the ones specified by `sources`.

vdumoulin:
I think we should document this quirk *really* well, as it's very easy to overlook.

vdumoulin:
There's only one test failing ATM, and it's caused by the fact that `Subset.sorted_fancy_indexing` doesn't support integer requests.

rizar:
Done, tests pass! @vdumoulin , do you have time to take a look?

vdumoulin:
Looks good, thanks!

PR #283, opened 2016-01-20
Make toy datasets deterministic.

dwf:
Adds a configurable seed that's popped off of kwargs if it's there, otherwise uses config.default_seed.

CC @jbornschein

### DIFF
--- a/fuel/datasets/toy.py
+++ b/fuel/datasets/toy.py
@@ -4,6 +4,7 @@ import numpy
 
 from collections import OrderedDict
 
+from fuel import config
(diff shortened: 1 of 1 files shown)


vdumoulin:
Good idea! LGTM.

jbornschein:
Looks good to me. 

It obviously should have been done like this from the beginning.

vdumoulin:
Merge at will when tests pass.

dwf:
Scrutinizer errors do not relate to this PR.

PR #285, opened 2016-01-26
Dogs vs cats

bartvm:
I added the Dogs vs. Cats dataset (for IFT6266). In general, it'd be nice if e.g. @vdumoulin could have a quick look to see if I'm doing it right. Also, few things I ran into:

* Following RFC5987 the `Content-Disposition` headers are formed differently from what the code assumes i.e. instead of ending with `filename=file` they look something like `filename=file; filename=*UTF8'file'` or something weird like that.
* Fuel prints "Downloading [target filename]" but the progress bar then displays the source filename, seems a bit odd.
* If the filesize can't be determined, a progressbar without `maxval` is created, resulting in a default progress bar that goes to 100. When `bar.update` is called this gives an error and crashes.
* H5PyDataset seems to have a hardcoded assumption that it is reshaping a batch, when you create an example stream instead it crashes when trying to reshape.

### DIFF
--- a/fuel/converters/dogs_vs_cats.py
+++ b/fuel/converters/dogs_vs_cats.py
@@ -21,7 +21,7 @@ def convert_dogs_vs_cats(directory, output_directory,
     :class:`fuel.datasets.dogs_vs_cats`. The converted dataset is saved as
     'dogs_vs_cats.hdf5'.
 
-    It assumes the existence of the following file:
+    It assumes the existence of the following files:
 
     * `dogs_vs_cats.train.zip`
     * `dogs_vs_cats.test1.zip`
--- a/fuel/converters/dogs_vs_cats.py
+++ b/fuel/converters/dogs_vs_cats.py
@@ -1,0 +1,122 @@
+import os
+import zipfile
+
+import h5py
+import numpy
+from PIL import Image
+
+from fuel.converters.base import check_exists, progress_bar
+from fuel.datasets.hdf5 import H5PYDataset
+
+TRAIN = 'dogs_vs_cats.train.zip'
+TEST = 'dogs_vs_cats.test1.zip'
+
+
+@check_exists(required_files=[TRAIN, TEST])
+def convert_dogs_vs_cats(directory, output_directory,
+                         output_filename='dogs_vs_cats.hdf5'):
+    """Converts the Dogs vs. Cats dataset to HDF5.
+
+    Converts the Dogs vs. Cats dataset to an HDF5 dataset compatible with
+    :class:`fuel.datasets.dogs_vs_cats`. The converted dataset is saved as
+    'dogs_vs_cats.hdf5'.
+
+    It assumes the existence of the following file:
+
+    * `dogs_vs_cats.train.zip`
+    * `dogs_vs_cats.test1.zip`
+
+    Parameters
+    ----------
+    directory : str
+        Directory in which input files reside.
+    output_directory : str
+        Directory in which to save the converted dataset.
+    output_filename : str, optional
+        Name of the saved dataset. Defaults to 'dogs_vs_cats.hdf5'.
+
+    Returns
+    -------
+    output_paths : tuple of str
+        Single-element tuple containing the path to the converted dataset.
+
+    """
+    # Prepare output file
+    output_path = os.path.join(output_directory, output_filename)
+    h5file = h5py.File(output_path, mode='w')
+    dtype = h5py.special_dtype(vlen=numpy.dtype('uint8'))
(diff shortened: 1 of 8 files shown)


bartvm:
Failing because of unrelated test, fixed in #286

fuel/converters/dogs_vs_cats.py
```diff
@@ -0,0 +1,121 @@
...
+TRAIN = 'dogs_vs_cats.train.zip'
+TEST = 'dogs_vs_cats.test1.zip'
+
+
+@check_exists(required_files=[TRAIN, TEST])
+def convert_dogs_vs_cats(directory, output_directory,
+                         output_filename='dogs_vs_cats.hdf5'):
+    """Converts the Dogs vs. Cats dataset to HDF5.
+
+    Converts the CIFAR-10 dataset to an HDF5 dataset compatible with
```
vdumoulin: Typo: `CIFAR-10`

fuel/converters/dogs_vs_cats.py
```diff
@@ -0,0 +1,121 @@
...
+    :class:`fuel.datasets.dogs_vs_cats`. The converted dataset is saved as
+    'dogs_vs_cats.hdf5'.
+
+    It assumes the existence of the following file:
+
+    * `dogs_vs_cats.zip`
```
vdumoulin: Don't you have `dogs_vs_cats.train.zip` and `dogs_vs_cats.test1.zip`?

fuel/converters/dogs_vs_cats.py
```diff
@@ -0,0 +1,121 @@
...
+                hdf_labels[i] = 0 if 'cat' in image_name else 1
+
+                # Update progress
+                i += 1
+                bar.update(i if split == TRAIN else i - 25000)
+
+    # Add the labels
+    split_dict = {}
+    sources = ['image_features', 'targets']
+    for name, slice_ in zip(['train', 'valid', 'test'],
```
vdumoulin: Is the validation set an official split? If not, I think we should stick to `train` and `test` and let users define their own validation split.

vdumoulin:
@bartvm thanks a lot!

Aside from my inline comments, I'm wondering if we should encode JPEG images as in the [ILSVRC2010 converter](https://github.com/mila-udem/fuel/blob/master/fuel/converters/ilsvrc2010.py#L283-L316). One one hand, this might mean slower processing at training time, but on the other hand we would stick closer to the original data, which I think is one of Fuel's goals regarding reproducibility. What are your thoughts on this?

bartvm:
I'm not too sure about storing things in JPEG. I thought that for ILSVRC2010 this was decided mostly for memory considerations (JPEG is just so much more space efficient)? If not for that, isn't it faster (as you said) and more user-friendly to store them as arrays. It just seems more intuitive if a user can just open the file in HDF5 without needing to read JPEG blobs.

Reproducibility is important, but having the conversion code being public and part of Fuel arguably meets that principle.

dwf:
The ILSVRC call was for disk space reasons.

Iteration speed is definitely a concern considering how CPU-hungry JPEG
decoding is. It's not really practical without a transcode (which I'm
working on now) or a fuel-server process. It can take upwards of 1.7
seconds to read, decode, resize, and crop a batch of 100 images from
ILSVRC2010 -- the overwhelming bulk of that is decoding. Someone should
eventually look into parallelizing across cores with jpeg4py (a
threadsafe/GIL-aware wrapper around the non-legacy API of libjpeg-turbo),
I'm apparently doing CCW tickets again so maybe I'll look into it in March.

Sticking close to the original data isn't a problem, storing the results of
JPEG decoding are as good as storing the JPEG, just not nearly as space
efficient.

>

vdumoulin:
@dwf @bartvm sounds good!

vdumoulin:
Would you mind adding unit tests, at least for the downloader and the dataset class?

bartvm:
0662d46ed5be1e385b4bc3724194217cd375a2c9

bartvm:
Remaining Travis error is unrelated, so let me know if this is good to merge (that way I can tell the students in IFT6266 to use it tomorrow morning) :)

dwf:
Just a note: reading HDF5 vlen data turns out to be pretty damn slow (close to a second with batch size 100 from local magnetic disk on ILSVRC2010). It's fine to keep it around, but it's worth considering maybe a more performant alternative for the case of images (honestly not sure what that might be). 

Also, once I get a jpeg4py transformer going, because it is both fast and sensible about the GIL, JPEG decode time will be a much smaller deal than once believed.

PR #286, opened 2016-01-26
Update dependencies

bartvm:


bartvm:
For some reason the `test_pickle` method of `test_server` randomly fails on Travis. I can't reproduce this behaviour at home, but it has made `master` fail a bunch of times. Simply raising `rtol` seems to fix it (although it'd be interesting to know what exactly is happening, considering the data isn't transformed in any way before being compared).

Either way, it would be nice to fix `master` and merge this.

bartvm:
@vdumoulin @dwf Will take just 1 minute to look at and will make Travis a bit more reliable (I hope)

vdumoulin:
LGTM, sorry about the delay!

Issue #173, opened 2015-06-17
Local cache for datasets


lamblin:
In order to avoid hammering a network drive when several processes need access to a dataset at the same time, which can be a problem at MILA, it would be really useful to have an automatic way of caching datasets on a local drive.
Pylearn2 has had that functionality for a while [[1](https://github.com/lisa-lab/pylearn2/blob/master/pylearn2/datasets/cache.py)], and it proved useful to reduce the load on file servers.
One way this could work would be:
- Use a new environment variable, for instance `FUEL_DATA_CACHE` to specify the path of that cache. If that variable is not set, no cache will be used.
- In `fuel.utils.find_in_data_path`, when the original data file is found, try to:
  - see if it is already in the cache, and check that the version is up to date,
  - if not, try to copy if into the cache,
  - if it works, return the path of the cached version, otherwise, return the original path.
- Maybe add an option to `find_in_data_path` to disable caching, for instance `ignore_data_cache`. This can be used by a particular Dataset class to disable caching for its data.

CC @nouiz who worked on the Pylearn2 version.

dwf:
I would caution against doing this automatically if the file exceeds about
a gigabyte (though this could be configurable). I also think it's important
to double check byte size and file metadata written inside to make sure
that a cached version is up to date.

nouiz:
It would be easy to port this. I suggested this by email today to Vincent:)

The code for the cache is located in only one file with no dependencies on
pylearn2. So just copy this file and use it.

As you recently added find_in_data_path, the change will be trivial to use
it.

Copy the file is : pylearn2/datasets/cache.py

the to use it:

from pylearn2.datasets.cache import datasetCache
label_path = datasetCache.cache_file(label_path)

If we can't copy it locally (like no space) we will return the original
path. If we cached it, we will return the local path.

We should take care as if multiple path are in FUEL_DATA_PATH and they have
name that colide, we  could colide them in the local cache. I'm not sure of
that, maybe not, so just double check that behavior.

nouiz:
It was done for the size and metadata check from memory.

We check the size availability and make sure only 1 process do the copy via
a lock (from memory same lock as Theano compiledir).

lamblin:
The code in Pylearn2 currently has a dependency on Theano, though (it re-uses theano's compile lock), but Fuel does not depend on Theano.

nouiz:
I think that part of Theano is standalone, you could copy it too...

dwf:
[...]

PR #287, opened 2016-01-26
Pull Request for Issue "Local cache for datasets #173"

alexmlamb:
Pull request to address this issue: 

https://github.com/mila-udem/fuel/issues/173

### DIFF
--- a/fuel/datasets/cache.py
+++ b/fuel/datasets/cache.py
@@ -18,6 +18,7 @@ import logging
 import os
 import stat
 import time
+import shutil
 
 import theano.gof.compilelock as compilelock
 
@@ -25,7 +26,7 @@ from fuel.utils import string_utils
 
 log = logging.getLogger(__name__)
 
-class LocalDatasetCache:
+class LocalDatasetCache(object):
 
     """
     A local cache for remote files for faster access and reducing
@@ -210,8 +211,9 @@ class LocalDatasetCache:
         if not os.path.exists(head):
             os.makedirs(os.path.dirname(head))
 
-        command = 'cp ' + remote_fname + ' ' + local_fname
-        os.system(command)
+        shutil.copyfile(remote_fname, local_fname)
+
+
         # Copy the original group id and file permission
         st = os.stat(remote_fname)
         os.chmod(local_fname, st.st_mode)
--- a/fuel/datasets/cache.py
+++ b/fuel/datasets/cache.py
@@ -1,0 +1,383 @@
+"""
+Dataset preloading tool
+This file provides the ability to make a local cache of a dataset or
+part of it. It is meant to help in the case where multiple jobs are
+reading the same dataset from ${PYLEARN2_DATA_PATH}, which may cause a
+great burden on the network.
+With this file, it is possible to make a local copy
+(in ${PYLEARN2_LOCAL_DATA_PATH}) of any required file and have multiple
+processes use it simultaneously instead of each acquiring its own copy
+over the network.
+Whenever a folder or a dataset copy is created locally, it is granted
+the same access as it has under ${PYLEARN2_LOCAL_DATA_PATH}. This is
+gauranteed by default copy.
+"""
+
+import atexit
+import logging
+import os
+import stat
+import time
+
+import theano.gof.compilelock as compilelock
+
+from fuel.utils import string_utils
+
+log = logging.getLogger(__name__)
+
+class LocalDatasetCache:
+
+    """
+    A local cache for remote files for faster access and reducing
+    network stress.
+    """
+
+    def __init__(self):
+        default_path = '${PYLEARN2_DATA_PATH}'
+        local_path = '${PYLEARN2_LOCAL_DATA_PATH}'
+        self.pid = os.getpid()
+
+        try:
+            self.dataset_remote_dir = string_utils.preprocess(default_path)
+            self.dataset_local_dir = string_utils.preprocess(local_path)
+        except (ValueError, string_utils.NoDataPathError,
+                string_utils.EnvironmentVariableError):
+            # Local cache seems to be deactivated
+            self.dataset_remote_dir = ""
+            self.dataset_local_dir = ""
+
+        if self.dataset_remote_dir == "" or self.dataset_local_dir == "":
+            log.debug("Local dataset cache is deactivated")
+
+    def cache_file(self, filename):
+        """
+        Caches a file locally if possible. If caching was succesfull, or if
+        the file was previously successfully cached, this method returns the
+        path to the local copy of the file. If not, it returns the path to
+        the original file.
+        Parameters
+        ----------
+        filename : string
+            Remote file to cache locally
+        Returns
+        -------
+        output : string
+            Updated (if needed) filename to use to access the remote
+            file.
+        """
+
+        remote_name = string_utils.preprocess(filename)
+
+        # Check if a local directory for data has been defined. Otherwise,
+        # do not locally copy the data
+        if self.dataset_local_dir == "":
+            return filename
+
+        common_msg = ("Message from Pylearn2 local cache of dataset"
+                      "(specified by the environment variable "
+                      "PYLEARN2_LOCAL_DATA_PATH): ")
+        # Make sure the file to cache exists and really is a file
+        if not os.path.exists(remote_name):
+            log.error("Error : Specified file %s does not exist" %
+                      remote_name)
+            return filename
+
+        if not os.path.isfile(remote_name):
+            log.error("Error : Specified name %s is not a file" %
+                      remote_name)
+            return filename
+
+        if not remote_name.startswith(self.dataset_remote_dir):
+            log.warning(
+                common_msg +
+                "We cache in the local directory only what is"
+                " under $PYLEARN2_DATA_PATH: %s" %
+                remote_name)
+            return filename
+
+        # Create the $PYLEARN2_LOCAL_DATA_PATH folder if needed
+        self.safe_mkdir(self.dataset_local_dir,
+                        (stat.S_IRUSR | stat.S_IWUSR | stat.S_IXUSR |
+                         stat.S_IRGRP | stat.S_IWGRP | stat.S_IXGRP |
+                         stat.S_IROTH | stat.S_IWOTH | stat.S_IXOTH))
+
+        # Determine local path to which the file is to be cached
+        local_name = os.path.join(self.dataset_local_dir,
+                                  os.path.relpath(remote_name,
+                                                  self.dataset_remote_dir))
+
+        # Create the folder structure to receive the remote file
+        local_folder = os.path.split(local_name)[0]
+        try:
+            self.safe_mkdir(local_folder)
+        except Exception as e:
+            log.warning(
+                (common_msg +
+                 "While creating the directory %s, we got an error."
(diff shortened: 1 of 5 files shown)


dmitriy-serdyuk:
Good job. What still needs to be done:

- Configuration for the local path. We use module-level configurations like [here](https://github.com/mila-udem/fuel/blob/master/fuel/config_parser.py#L193). 
- Remove Theano dependency. We don't want to depend on Theano. As I can see, this piece of code uses a compile lock from theano and six library. It's easy to fix later one. 
- Use caching by default. I propose to add a `do_not_cache` flag to the [HDF5 dataset class](https://github.com/mila-udem/fuel/blob/master/fuel/datasets/hdf5.py#L94) which defaults to false. We use hdf5 for almost everything, so it should be enough.
- Style and comments. Take a look at this [document](https://github.com/mila-udem/blocks/blob/master/CONTRIBUTING.rst). We use numpy-style docstrings: short description, blank line, long description, blank line, parameters, etc.
- Test. After caching for HDF5 is done, this code will be tested. But if you want, you can add an example (doctest) here, it would be useful.
- Figure out what's wrong with Travis build.

alexmlamb:
Are there any unit tests that you would recommend emulating?  I imagine
that I'd want to use the usual unit tests but with caching added.

fuel/utils/exc.py
```diff
@@ -0,0 +1,86 @@
+"""Exceptions used by basic support utilities."""
+__author__ = "Ian Goodfellow"
```
dmitriy-serdyuk: We don't usually put the author. Can you just acknowledge Ian and pylearn in the docstring?

dwf:
There is definitely some duplication in the utils file that was imported. We already use a `reraise_as` all over the place.

dwf:
Actually, it's apparently in Blocks. Since Fuel is a dependency of Blocks but not vice-versa, it makes sense to move the version from Blocks (including its better docstring and tests) to Fuel, and then send a pull request to Blocks to change imports around to use the Fuel version.

fuel/utils/string_utils.py
```diff
@@ -0,0 +1,74 @@
...
+
+from theano.compat.six import string_types
+from theano.compat.six.moves import xrange
+
+from fuel.utils.exc import EnvironmentVariableError, NoDataPathError
+from fuel.utils.exc import reraise_as
+from fuel.utils.common_strings import environment_variable_essay
+
+
+def preprocess(string, environ=None):
```
dwf: @dmitriy-serdyuk I really don't think we want to import this thing. It's pretty ugly.

dwf:
Re: tests, the point of true *unit* tests is to test some specific behaviour, in this case the copying behaviour. This implies setting up simple, controlled environments and verifying that the copy actually takes place. A more complicated integration test is good but not sufficient.

I also think some additional thought is required given that

a) Lots more people use Fuel outside of the lab infrastructure, so turning on caching by default is questionable (not everybody's using a network drive by default, and this adds complexity/magic behaviour)
b) Fuel already has a multiple-data-path system in place; fitting the caching strategy into that sanely is a concern. Seems like we can address with a much simpler mechanism that designates one directory on the path as a cache path and copies files to it if caching is enabled and they are found in any other place.

fuel/datasets/cache.py
```diff
@@ -0,0 +1,383 @@
+"""
```
bartvm: https://blocks.readthedocs.org/en/latest/development/docs.html#docstrings

fuel/datasets/cache.py
```diff
@@ -0,0 +1,383 @@
...
+import stat
+import time
+
+import theano.gof.compilelock as compilelock
+
+from fuel.utils import string_utils
+
+log = logging.getLogger(__name__)
+
+class LocalDatasetCache:
```
bartvm: Use new-style classes

fuel/datasets/cache.py
```diff
@@ -0,0 +1,383 @@
...
+        """
+
+        remote_name = string_utils.preprocess(filename)
+
+        # Check if a local directory for data has been defined. Otherwise,
+        # do not locally copy the data
+        if self.dataset_local_dir == "":
+            return filename
+
+        common_msg = ("Message from Pylearn2 local cache of dataset"
```
bartvm: ~~Pylearn2~~

fuel/utils/exc.py
```diff
@@ -0,0 +1,86 @@
...
+        Arguments passed to Exception()
+    """
+
+    def __init__(self, *args):
+        super(EnvironmentVariableError, self).__init__(*args)
+
+
+# This exception is here as string_utils need it and setting it in
+# datasets.exc would create a circular import.
+class NoDataPathError(EnvironmentVariableError):
```
bartvm: Handled by https://github.com/mila-udem/fuel/blob/db8fe3177762053638842dc6b7b5e63490a47b69/fuel/config_parser.py#L149-L150

bartvm:
Looking at the caching code, I honestly think it wouldn't hurt to clean it up a bit. It does stuff like issuing `cp` commands in a subshell instead of using `shutil`, runs code on import (by constructing the `LocalDatasetCache` object) and mixes camelCase and snake_case. I also think it could do with a bit of EAFP; why not just attempt the copy and inspect any resulting `IOError` (using the `errno` module to give sensible errors/log messages if needed) instead of trying to determine the disk usage manually.

I guess there's a design decision to make here as well. The current code does everything possible to fail gracefully if the cache is outdated, unwritable, unreadable, etc. Personally I'm more inclined to fail fast; if the user's configuration suggests that (s)he wanted to cache and we failed to do that, we should just give up. The user then has the choice to disable caching, or to fix the issue. That seems better than inadvertently bludgeoning a network drive with nothing but a few warnings hidden in the log.

dmitriy-serdyuk:
I'm pretty sure, that Fred is going to push caching by default. Or we should place flashing warnings everywhere that one needs to use caching. What do you thing about the following behaviour: if `FUEL_DATA_CACHE` is not set or equal to `FUEL_DATA_PATH` then fuel doesn't do caching, otherwise it does. So we can setup these variable in the lab and on cluster.

Concerning refactoring, I think it's better to have a working version first and then start refactoring it. 

Blocks depends on fuel, but if we move `reraise_as` to fuel it is weird a bit. It's not logical that some common exception utility is found in a data piping library and will cause frustration. I don't see a better solution, though.

bartvm:
@dmitriy-serdyuk Right, I think that's what @dwf and I had in mind, enabling caching using a configuration setting (with an environment variable) that can be set in the lab or on the cluster, but that isn't set by default so that it doesn't bother people outside of the lab.

I disagree with postponing the refactoring if someone is working on the code right now anyway. If we merge a working version first, we need to go through the entire overhead again of finding a person to do the refactoring, opening a new pull request, new discussion, new code review, etc. Easier to just do it in one go, especially since the required code is pretty easy and straightforward.

dmitriy-serdyuk:
@alexmlamb , we use Travis to run tests, you take a look at the script. Here we set the data [path](https://github.com/mila-udem/fuel/blob/master/.travis.yml#L23) and here we download the [datasets](https://github.com/mila-udem/fuel/blob/master/.travis.yml#L41). You can take a look at the downloader [tests](https://github.com/mila-udem/fuel/blob/master/tests/test_downloaders.py#L95).

fuel/utils/string_utils.py
nouiz: This was added for a reason. Before removing this we should see if that reason still apply.

nouiz:
I don't see a problem with fuel multi-path. In Pylearn2 from  memory, in the cache, we set the full path inside the local cache directory. If you still think there is a problem, can you explain more?

We should not copy if there isn't enough space. We do not want to fill by default the space to 100%. This can cause other problem outside fuel. We can't easily assume that by default that local cache is reserved for the cache.

If we can't copy into the cache due to permission, the user isn't always able to fix it. Don't forget, the cache is shared by many user, maybe one user didn't set the permission correctly. Then only an admin can fix this. So I'm not sure this is better then using a remove copy. What do you think? This happened due to problem with the first merged version inside pylearn2. As now, the bugs are fixed and we don't risk too much people still using that old bugged pylearn2 version, that risk is low.

bartvm:
I think @dwf is referring to the fact that the current code does things like `if not os.path.exists(remote_name)`, but in Fuel there might be many possible remotes so that needs to change (using `fuel.utils.find_in_data_path`).

I don't understand what you mean about permissions. If the user gets an error because of permissions which don't allow him to fill the cache or read it, (s)he can just disable the cache after getting an error (e.g. start the script with `FUEL_CACHE= python train.py` instead).

Failing silently like it is done right now is a bad idea in my mind. It makes more sense to fail fast instead of tailoring the behaviour to MILA's setup. For us, it might make sense to simply use the remote file and ignore the cache, but imagine this is being started on a cluster and the user expects caching to happen, in that case silently failing could completely kill the remote server/result in super slow training.

Does this work as an overall approach? We create two configuration settings with accompanying environment variables: `FUEL_CACHE=path` and `FUEL_MAX_DISK_USAGE=0.9`. By default they are empty and 1.0, but we can change that at MILA. If `FUEL_CACHE` is non-empty we look for the file in any of the data paths, and see if we can copy it to the cache while staying below `FUEL_MAX_DISK_USAGE`. If any of this fails, if the local file is newer, etc. we simply raise an error and ask the user to fix it or disable the cache.

Also, why are the write locks obtained the way they are? Wouldn't it be more usual to use `fcntl.lockf`?

bartvm:
Also, does it make sense to define `FUEL_MAX_DISK_USAGE` as a fraction of the disk? Maybe it makes more sense to specify an absolute number in bytes that you want to keep free, that way the setting more naturally applies to disks of different sizes (it's a bit weird to insist on keep 1GB free on a 10GB disk, but 100GB on a 1TB disk).

dmitriy-serdyuk:
@bartvm , the second option sounds better. In this case I would call this variable like `FUEL_DISK_RESIDUAL`, what are your thoughts on it?

nouiz:
The proposition make sense. I was telling why it was not always raising an
error in the past. It make sense to change that now.

It was quicker for me to use the Theano lock then the lock you tall as I
was knowing it. As the cache should be local, the lock you talk about
should work (it don't work on all remote path, like on NFS and on some
cluster like Helios).

For the max size, I whould go to is less then 1, then a %, if 1 or more
then 1, the max size the cache can take. Not the size to keep free. We
specify space more frequently as what to use, not as what to let free.

bartvm:
How exactly would the size of the cache be defined though? If someone wants
to use /Tmp as a cache path we can't just take the size of that directory,
many other things could be in there. Moreover, getting the size of a
directory is much slower than getting the free space on a disk. For those
reasons I figured that defining the minimum amount of free space would be
easier.

nouiz:
I got confused. What you describe is what is implemented. See the method
check_enough_space

bartvm:
Yes, except that it checks whether a minimum of X% is free instead of checking whether a minimum of N bytes remain free after caching, which is the change I was proposing. We could do as you suggest (interpreting it as % when less than 1, otherwise as (mega)bytes), although I find it a bit confusing to have multiple interpretations of a single configuration setting.

nouiz:
I like your definition of % free after caching.

maximecb:
@rizar we should probably close this PR

PR #289, opened 2016-01-26
Support per-example iteration for vlen datasets

rizar:


rizar:
In my previous PR #289 I forgot about testing the variable length case, and of course it did not work :)

rizar:
@dmitriy-serdyuk , maybe you can review and we can merge it without disturbing Vincent?

fuel/datasets/hdf5.py
```diff
@@ -542,8 +542,11 @@ def get_data(self, state=None, request=None):
             data, shapes = self._out_of_memory_get_data(state, request)
         for i in range(len(data)):
```
dmitriy-serdyuk: `for data_point, shape in zip(data, shapes)`?

fuel/datasets/hdf5.py
```diff
@@ -542,8 +542,11 @@ def get_data(self, state=None, request=None):
...
             if shapes[i] is not None:
-                for j in range(len(data[i])):
-                    data[i][j] = data[i][j].reshape(shapes[i][j])
+                if isinstance(request, numbers.Integral):
+                    data[i] = data[i].reshape(shapes[i])
+                else:
+                    for j in range(len(data[i])):
```
dmitriy-serdyuk: Same here. Why do you need `j`?

fuel/datasets/hdf5.py
dwf: equizip would be more defensive

tests/test_hdf5.py
```diff
@@ -232,7 +232,7 @@ def test_out_of_memory_unsorted_indices(self):
         assert_raises(TypeError, dataset.get_data, handle, [7, 4, 6, 2, 5])
         dataset.close(handle)
 
-    def test_in_of_memory_example_scheme(self):
+    def test_in_memory_example_scheme(self):
```
dmitriy-serdyuk: I think, it should be of-memory, in-memory is tested below.

dmitriy-serdyuk: Oh, it should be a typo: see line 237.

fuel/datasets/hdf5.py
rizar: Because we are doing inplace modification, which makes sense in this performance-critical place.

tests/test_hdf5.py
rizar: It is a fixed typo, isn't it?

dmitriy-serdyuk: https://github.com/mila-udem/fuel/pull/289/files#diff-7c09a0dc93ce18f66103684bcb4d2b01R244 tests in-memory dataset.

dmitriy-serdyuk: My bad, everything is fine now.

dmitriy-serdyuk:
Ok, LGTM

tests/test_hdf5.py
rizar: Thanks!

PR #297, opened 2016-01-27
Add YouTube audio dataset

bartvm:
@sotelo

Allows you to download audio from a YouTube video, convert to .wav format and save in HDF5 e.g.

```bash
fuel-download youtube_audio --youtube-id XqaJ2Ol5cC4
fuel-convert youtube_audio --youtube-id XqaJ2Ol5cC4
```
```python
from fuel.datasets.youtube_audio import YouTubeAudio
dataset = YouTubeAudio('XqaJ2Ol5cC4')
```

Requiers the Python `pafy` module for downloading, and `ffmpeg` for converting.

### DIFF
--- a/fuel/converters/youtube_audio.py
+++ b/fuel/converters/youtube_audio.py
@@ -1,0 +1,50 @@
+import os
+
+import h5py
+import scipy.io.wavfile
+
+from fuel.converters.base import fill_hdf5_file
+
+
+def convert_youtube_audio(directory, output_directory, youtube_id, channels,
+                          sample, output_filename=None):
+    input_file = os.path.join(directory, '{}.m4a'.format(youtube_id))
+    wav_filename = '{}.wav'.format(youtube_id)
+    wav_file = os.path.join(directory, wav_filename)
+    command = "ffmpeg -y -i {} -ac {} -ar {} {}".format(
+            input_file, channels, sample, wav_file)
+    os.system(command)
+
+    # Load WAV into array
+    _, data = scipy.io.wavfile.read(wav_file)
+    if data.ndim == 1:
+        data = data[:, None]
+
+    # Store in HDF5
+    if output_filename is None:
+        output_filename = '{}.hdf5'.format(youtube_id)
+    output_file = os.path.join(output_directory, output_filename)
+
+    with h5py.File(output_file, 'w') as h5file:
+        fill_hdf5_file(h5file, (('train', 'features', data),))
+        h5file['features'].dims[0].label = 'time'
+        h5file['features'].dims[1].label = 'feature'
+
+    return (output_file,)
+
+
+def fill_subparser(subparser):
+    subparser.add_argument(
+        '--youtube-id', type=str, required=True,
+        help=("The YouTube ID of the video from which to extract audio, "
+              "usually an 11-character string.")
+    )
+    subparser.add_argument(
+        '--channels', type=int, default=1,
+        help="The number of audio channels to convert to"
+    )
+    subparser.add_argument(
+        '--sample', type=int, default=16000,
+        help="The sampling rate in Hz"
+    )
+    return convert_youtube_audio
--- a/fuel/converters/youtube_audio.py
+++ b/fuel/converters/youtube_audio.py
@@ -1,4 +1,6 @@
 import os
+import subprocess
+import sys
 
 import h5py
 import scipy.io.wavfile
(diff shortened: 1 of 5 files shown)


bartvm:
@vdumoulin @rizar Are there any conventions on how to deal with time series data in Fuel?

fuel/converters/youtube_audio.py
```diff
@@ -0,0 +1,50 @@
...
+
+from fuel.converters.base import fill_hdf5_file
+
+
+def convert_youtube_audio(directory, output_directory, youtube_id, channels,
+                          sample, output_filename=None):
+    input_file = os.path.join(directory, '{}.m4a'.format(youtube_id))
+    wav_filename = '{}.wav'.format(youtube_id)
+    wav_file = os.path.join(directory, wav_filename)
+    command = "ffmpeg -y -i {} -ac {} -ar {} {}".format(
```
dwf: Kinda brittle, but I remember pyffmpeg being a nightmare, so. Any way we can tighten this up a bit?

bartvm: Yeah, I briefly looked at pyffmpeg and it came with a warning saying "will probably not compile with any recent version of ffmpeg" which didn't sound promising.

I can add a check to see if ffmpeg is available and whether the input file exists and give sensible errors if either is false, but otherwise I don't see an easy way around calling this command in a subshell.

fuel/datasets/youtube_audio.py
```diff
@@ -0,0 +1,10 @@
+from fuel.datasets.hdf5 import H5PYDataset
+from fuel.utils import find_in_data_path
+
+
+class YouTubeAudio(H5PYDataset):
+    def __init__(self, youtube_id, **kwargs):
+        super(YouTubeAudio, self).__init__(
+            file_or_path=find_in_data_path('{}.hdf5'.format(youtube_id)),
+            which_sets=('train',), **kwargs
+        )
```
vdumoulin: Is the trailing closing parenthesis PEP8-compliant?

vdumoulin:
@bartvm There is no convention specific to time series data other than the ones you'd have to respect if your data had variable length, but if I understand correctly this dataset is composed of only one example.

Out of curiosity, is there a reason why you don't put a `batch` axis of size one as the first axis? The way things are set up right now, iterating over the dataset will iterate over time steps.

If that's the desired behaviour, I think the following would be more "standard":
* Have your dataset be of shape `(1, time_steps, num_channels)` with axis labels `('batch', 'time', 'features')`.
* Write a `Transformer` that takes a single example as input and iterates over the time axis according to some iteration scheme.

This would have the advantage of being more future-proof if you ever wanted to include more than one audio track in the same dataset.

fuel/datasets/youtube_audio.py
bartvm: Yup.

> The closing brace/bracket/parenthesis on multi-line constructs may either line up under the first non-whitespace character of the last line of list, as in:

> ```
my_list = [
    1, 2, 3,
    4, 5, 6,
    ]
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
    )
```
> or it may be lined up under the first character of the line that starts the multi-line construct, as in:
```
my_list = [
    1, 2, 3,
    4, 5, 6,
]
result = some_function_that_takes_arguments(
    'a', 'b', 'c',
    'd', 'e', 'f',
)
```

bartvm:
Yeah, in this case there is only one example. In general you could have a series of audio clips though, and audio can have multiple channels, so you need at least 3 axes: `batch`, `time` and `feature`. In this case both the `batch` and `feature` dimension are of size 1, but I think you're right, it's best to create them anyway.

It would be good if the `TextFile` was made consistent though, because there too you have batches (sentences), time and feature (the feature has only 1 dimension in this case as well, but could have more for e.g. part-of-speech tagged sentences).

The `NGrams` transformer should probably be re-used for what you're proposing (combined with the `Cache` transformer) because that basically results in iterating over the time axis in the way you'd want to. `NGrams` sounds very NLP-y to me though, maybe it should be renamed, or do they use that term in e.g. speech as well when they chop up their sequences?

rizar:
I would  rather expect speech people the use the word window, but I am not
an expert.

vdumoulin:
We could rename `NGrams` to something like `Chunk`, and then alias it with both `NGrams` and `Window` (or `Frame`) to make it more discoverable by NLP and speech people.

bartvm:
@vdumoulin I didn't have time this weekend to add tests, but it'd be handy if we could merge this so that IFT6266 students don't need to figure out how to install a particular Fuel branch. Can I merge with the promise of adding tests sometime later this week? :sweat:

vdumoulin:
Sure!

PR #298, opened 2016-01-28
Support for list of same-length examples

bartvm:
Right now `Flatten` chokes when you feed it a list of identically sized arrays (as output by e.g. `RandomFixedSizeCrop` and `MinimumImageSize`. Using `len` instead of `.shape[0]` should support both cases (list and NumPy array).


bartvm:
@vdumoulin @dwf One-liner that would be nice to have merged for the students in IFT6266 that are starting their projects today, last one for now I hope!

vdumoulin:
Test failure unrelated, merging. Thanks!

PR #265, opened 2015-10-30
treating the bug while data url fails.

hantek:
Tries to fix #261

PR #300, opened 2016-01-28
CelebA dataset

vdumoulin:
For the time being, it's only the aligned and cropped version (as well as the scaled and cropped 64x64 version used in e.g. [Larsen et. al. Autoencoding beyond pixels using a learned similarity metric](http://arxiv.org/pdf/1512.09300.pdf).

Someone can later integrate the non-cropped version of the dataset should the need arise.

@bartvm @dwf ?

### DIFF
--- a/fuel/converters/celeba.py
+++ b/fuel/converters/celeba.py
@@ -1,0 +1,102 @@
+import os
+import zipfile
+
+import h5py
+import numpy
+import six
+from six.moves import range, cPickle
+from PIL import Image
+
+from fuel.converters.base import check_exists, progress_bar
+
+IMAGE_FILE = 'img_align_celeba.zip'
+ATTRIBUTES_FILE = 'list_attr_celeba.txt'
+DATASET_FILES = [IMAGE_FILE, ATTRIBUTES_FILE]
+NUM_EXAMPLES = 202599
+TRAIN_STOP = 162770
+VALID_STOP = 182637
+
+
+@check_exists(required_files=DATASET_FILES)
+def convert_celeba(directory, output_directory, output_filename='celeba.hdf5'):
+    """Converts the CelebA dataset to HDF5.
+
+    Converts the CelebA dataset to an HDF5 dataset compatible with
+    :class:`fuel.datasets.CelebA`. The converted dataset is saved as
+    'celeba.hdf5'.
+
+    It assumes the existence of the following files:
+
+    * `img_align_celeba.zip`
+    * `list_attr_celeba.txt`
+
+    Parameters
+    ----------
+    directory : str
+        Directory in which input files reside.
+    output_directory : str
+        Directory in which to save the converted dataset.
+    output_filename : str, optional
+        Name of the saved dataset. Defaults to 'celeba.hdf5'.
+
+    Returns
+    -------
+    output_paths : tuple of str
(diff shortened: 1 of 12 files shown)


fuel/converters/celeba.py
```diff
@@ -0,0 +1,212 @@
...
+TRAIN_STOP = 162770
+VALID_STOP = 182637
+
+
+def _initialize_conversion(directory, output_path, image_shape):
+    h5file = h5py.File(output_path, mode='w')
+    split_dict = {
+        'train': {
+            'features': (0, TRAIN_STOP),
+            'attributes': (0, TRAIN_STOP)},
```
bartvm: Why not just stick with `targets` for consistency with most other datasets? Otherwise it should probably be mentioned in the docstring somewhere.

fuel/converters/celeba.py
```diff
@@ -0,0 +1,212 @@
...
+    with zipfile.ZipFile(image_file_path, 'r') as image_file:
+        with progress_bar('images', NUM_EXAMPLES) as bar:
+            for i in range(NUM_EXAMPLES):
+                image_name = 'img_align_celeba/{:06d}.jpg'.format(i + 1)
+                features_dataset[i] = numpy.asarray(
+                    Image.open(
+                        image_file.open(image_name, 'r'))).transpose(2, 0, 1)
+                bar.update(i + 1)
+
+    h5file.flush()
```
bartvm: I've seen this in a bunch of places. What's the point to `flush()` right before `close()`, doesn't the latter imply the former?

tests/test_converters.py
```diff
@@ -292,6 +293,70 @@ def test_converter(self):
                      ('batch', 'channel', 'height', 'width'))
 
 
+class TestCelebA(object):
+    def setUp(self):
+        numpy.random.seed(21 + 1 + 2016)
+        self.tempdir = tempfile.mkdtemp()
+        cwd = os.getcwd()
```
bartvm: Not needed for this PR perhaps, but this reminds me that we finally need the context manager that @hantek wrote in #265.

bartvm:
LGTM

fuel/converters/celeba.py
vdumoulin: To be honest, I'm not sure. The [online documentation](http://docs.h5py.org/en/latest/high/file.html#File.close) isn't very informative, and it does not explicitly mention that a `close` call will flush to disk:

```
close()
    Close this file. All open objects will become invalid.
```

I guess it's a superstitious pattern I've seen repeated enough time that I've assumed it's necessary.

fuel/converters/celeba.py
vdumoulin: 0e1fa83

tests/test_converters.py
vdumoulin: #302 squashes and rebases #265. If I can manage to get it merged soon I'll use the context manager for this PR.

bartvm: Merged!

vdumoulin:
Tests are failing because of that elusive `test_pickling` issue, but other than that everything's passing.

vdumoulin:
I'll go ahead and merge, since I've got the "LGTM".

PR #305, opened 2016-02-02
UTF-8 and Gzip for text files

bartvm:
Adds a helper function to open text files which ensures UTF-8 encoding
is used in Python 2, and allows the user to pass files with Gzip
compression (which can be faster as well as more space efficient on
disk).

### DIFF
--- a/fuel/utils/formats.py
+++ b/fuel/utils/formats.py
@@ -1,6 +1,48 @@
 """Low-level utilities for reading a variety of source formats."""
+import codecs
+import gzip
+import io
 import tarfile
 import six
+
+
+def open_(filename, mode='r', encoding=None):
+    """Open a text file with encoding and optional gzip compression.
+
+    Note that on legacy Python any encoding other than ``None`` or opening
+    GZipped files will return an unpicklable file-like object.
+
+    Parameters
+    ----------
+    filename : str
+        The filename to read.
+    mode : str, optional
+        The mode with which to open the file. Defaults to `r`.
+    encoding : str, optional
+        The encoding to use (see the codecs documentation_ for supported
+        values). Defaults to ``None``.
+
+    .. _documentation:
+    https://docs.python.org/3/library/codecs.html#standard-encodings
+
+    """
+    if filename.endswith('.gz'):
+        if six.PY2:
+            zf = io.BufferedReader(gzip.open(filename, mode))
+            if encoding:
+                return codecs.getreader(encoding)(zf)
+            else:
+                return zf
+        else:
+            return io.BufferedReader(gzip.open(filename, mode,
+                                               encoding=encoding))
+    if six.PY2:
+        if encoding:
+            return codecs.open(filename, mode, encoding=encoding)
+        else:
+            return open(filename, mode)
+    else:
+        return open(filename, mode, encoding=encoding)
 
 
 def tar_open(f):
--- a/fuel/converters/youtube_audio.py
+++ b/fuel/converters/youtube_audio.py
@@ -28,8 +28,8 @@ def convert_youtube_audio(directory, output_directory, youtube_id, channels,
     sample : int
         The sampling rate to use in Hz, e.g. 44100 or 16000.
     output_filename : str, optional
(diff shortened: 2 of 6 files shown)


fuel/utils/formats.py
```diff
@@ -1,8 +1,35 @@
...
+    encoding : str, optional
+        The encoding to use (see the codecs documentation_ for supported
+        values). Defaults to `utf-8`.
+
+    .. _documentation:
+    https://docs.python.org/3/library/codecs.html#standard-encodings
+
+    """
+    if filename.endswith('.gz.'):
+        zf = gzip.open(filename, mode)
```
nouiz: I already tried this gzip and it was very slow to compute. It use a too small buffer to uncompress. It was way faster to uncompress on the command line wigh gunzip. Did you see speed up with this compared to regular file? If so, with which version of Python?

bartvm: Using `io.BufferedReader` (6a32dd44bd10eaf87ae7f41778f4942cd5ac339c) speeds things up a lot, it's now only 10% slower than the command line in my test (I'm comparing with `subprocess.Popen(['zcat', fn], stdout=subprocess.PIPE)`). For 4.5 million lines (670MB uncompressed, 270MB compressed) it takes `zcat` 5.5 seconds and `gzip.open` 6.4.

On my desktop the slowdown is about 2x compared to reading the raw text file (takes 2.9 seconds), but I have an SSD and lots of space; I figured that when reading from a network drive and when the text files take up a lot of space it might definitely be worth using `gzip` (keeping in mind that we have monolingual datasets that are 100GB *compressed*).

fuel/utils/formats.py
```diff
@@ -1,8 +1,36 @@
...
+        The mode with which to open the file. Defaults to `r`.
+    encoding : str, optional
+        The encoding to use (see the codecs documentation_ for supported
+        values). Defaults to `utf-8`.
+
+    .. _documentation:
+    https://docs.python.org/3/library/codecs.html#standard-encodings
+
+    """
+    if filename.endswith('.gz.'):
```
vdumoulin: Is the trailing period intentional?

vdumoulin:
Other than my very brief inline comment, LGTM.

fuel/utils/formats.py
bartvm: Nope, thanks.

bartvm:
Okay, I think this is good to go. In summary, `gzip` file handles are only picklable in Python 3 (I tried fixing it in picklable-itertools for Python 2, but it's a nightmare because the buffered stream makes it really hard to tell where you are in the underlying file). Same goes for UTF-8 encoding, which works fine in Python 3 but in Python 2 becomes unpicklable.

vdumoulin:
Wait, would that mean that for Blocks and under Python2 we wouldn't be able to pickle the main loop within an epoch?

bartvm:
Yes, but only if you use `.gz` files or if you pass `encoding='utf-8'`. Neither is serializable in Python 2 without a lot of custom pickling (that I tried to do this morning, but gave up on after 2 hours).

In general though, you can get away with never using UTF-8 in Python 2 as long as you make sure that the dictionary doesn't use UTF-8 either. I was using this within a project where the dictionary was already created elsewhere as UTF-8, so in that case it's useful to have the keyword argument.

In short, nothing changes about the current behaviour, but if you insist on using the new functionality of reading files in UTF-8 or using `gzip` in Python 2 you will lose picklability, yes.

vdumoulin:
As long as it's not breaking anything existing, I'm okay with that. You should document it, though.

dwf:
Yes, please document it, [and please use the phrase "legacy Python" and not "Python 2" to do so](https://carreau.github.io/posts/planning-an-early-death-for-python-2.html). :wink:

bartvm:
It's documented already as part of the `TextFile` docstring, or do you mean that I should put in the `open_` docstring as well?

@dwf Sure.

vdumoulin:
I'd put it in the `open_` docstring as well, just in case it's reused by future library components or user code.

bartvm:
Done and rebased.

PR #307, opened 2016-02-03
Add a flag to ignore UNK token check

dmitriy-serdyuk:
I know all the tokens in my dataset!

### DIFF
--- a/fuel/datasets/text.py
+++ b/fuel/datasets/text.py
@@ -122,7 +122,8 @@ class TextFile(Dataset):
             return value
         else:
             if self.unk_token is None:
-                raise KeyError("UNK token")
+                raise KeyError("token '{}' not found in dictionary and no "
+                               "`unk_token` given".format(symbol))
             return self.dictionary[self.unk_token]
 
     def get_data(self, state=None, request=None):
--- a/fuel/datasets/text.py
+++ b/fuel/datasets/text.py
@@ -27,7 +27,10 @@ class TextFile(Dataset):
         ``bos_taken``.
     unk_token : str, optional
         The token in the dictionary to fall back on when a token could not
-        be found in the dictionary. ``<UNK>`` by default.
+        be found in the dictionary. ``<UNK>`` by default. Pass ``None`` if
+        the dataset doesn't contain any out-of-vocabulary words/characters
+        (the data request is going to crash if meets an unknown symbol).
+
     level : 'word' or 'character', optional
         If 'word' the dictionary is expected to contain full words. The
         sentences in the text file will be split at the spaces, and each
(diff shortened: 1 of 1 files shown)


fuel/datasets/text.py
```diff
@@ -75,20 +91,27 @@ class TextFile(Dataset):
...
-            raise ValueError
+            raise ValueError(
+                "BOS token '{}' is not in the dictionary".format(bos_token))
         self.bos_token = bos_token
         if eos_token is not None and eos_token not in dictionary:
-            raise ValueError
+            raise ValueError(
+                "EOS token '{}' is not in the dictionary".format(eos_token))
         self.eos_token = eos_token
         if unk_token not in dictionary:
```
bartvm: Why not just do the same for the unknown-token as for the others i.e. allow `unk_token=None` to be passed and check `if unk_token is not None and unk_token not in dictionary`. Saves an extra keyword argument, plus it's more consistent.

fuel/datasets/text.py
```diff
@@ -98,6 +121,13 @@ def open(self):
         return chain(*[iter_(open_(f, encoding=self.encoding))
                        for f in self.files])
 
+    def _get_from_dictionary(self, symbol):
+        if symbol in self.dictionary:
```
bartvm: Nitpicking, but since this is potentially done many times, it's worth considering doing something like this, so that we only perform 1 dictionary lookup instead of 2.
```python
value = self.dictionary.get(symbol)
if value is None:
    return self.dictionary[self.unk_token]
```

fuel/datasets/text.py
dmitriy-serdyuk: Good idea, I'll do it.

dmitriy-serdyuk:
@bartvm , done.

dmitriy-serdyuk:
TestServer crashes, I have no idea why...

bartvm:
It's a non-deterministic error that I can't reproduce locally, don't worry about it, we should probably disable it.

fuel/datasets/text.py
```diff
@@ -98,6 +116,15 @@ def open(self):
         return chain(*[iter_(open_(f, encoding=self.encoding))
                        for f in self.files])
 
+    def _get_from_dictionary(self, symbol):
+        value = self.dictionary.get(symbol)
+        if value is not None:
+            return value
+        else:
+            if self.unk_token is None:
+                raise KeyError("UNK token")
```
bartvm: Perhaps less confusing: ```token '{}' not found in dictionary and no `unk_token` given```

bartvm:
LGTM besides the unclear error message

dmitriy-serdyuk:
The same test crashes, everything else is fine, so I merge.

PR #310, opened 2016-02-08
Add Window transformer

bartvm:
As discussed in #297.


dwf:
Any reason this isn't a `SourcewiseTransformer`? Seems like it would simplify it.

dwf:
Nevermind, I misunderstood.

dwf:
Oops, I gave this an initial look before but didn't merge.

PR #314, opened 2016-02-23
Merge master into stable get a new stable version

rizar:
See https://github.com/mila-udem/blocks/issues/1000


rizar:
@vdumoulin , what do you think about the error below?


ERROR: tests.test_downloaders.FunctionTestCase (test_ensure_directory_exists)

  File "/home/travis/build/mila-udem/fuel/tests/test_downloaders.py", line 90, in test_ensure_directory_exists

    assert_raises(OSError, ensure_directory_exists(filepath))

  File "/home/travis/miniconda/lib/python2.7/site-packages/numpy/testing/utils.py", line 1109, in assert_raises

    return nose.tools.assert_raises(*args,**kwargs)

  File "/home/travis/miniconda/lib/python2.7/unittest/case.py", line 475, in assertRaises

    callableObj(*args, **kwargs)

TypeError: 'NoneType' object is not callable

vdumoulin:
For starters, the call to `assert_raises` is wrong:

``` python
assert_raises(OSError, ensure_directory_exists(filepath))
```

should be

``` python
assert_raises(OSError, ensure_directory_exists, filepath)
```

and I'm not even sure why this test isn't crashing at the moment.

If we fix that, then the test fails.

rizar:
Funny, there's a bug in both `ensure_directory_exists` and test! On my machine, when I try to create a directory at the path, where currently there is a file, the exact exception raised is exactly `os.eerno.EEXIST`, which is silently intercepted in `ensure_directory_exists`. Sigh...

rizar:
Nooooooo, `test_pickling` is still episodically failing :( @bartvm  and @vdumoulin , please tell me what to do with it! I don't want to have this test in the stable branch.

bartvm:
I haven't been able to reproduce it, so we need to debug it on Travis at one point, which is a pain. In the meantime, just disable it for `stable`.

rizar:
Can I just comment `assert_close`, merge it to master and create an issue?
It is good to have a picklability test, even if we can't compare the
values.

On 24 February 2016 at 17:48, Bart van Merriënboer <notifications@github.com

dwf:
@rizar Check out my PR not long ago in blocks, just try it and if it fails,
raise a SkipTest, leaving a comment/issue that this needs investigation.

https://github.com/dwf/blocks/commit/417dbdf8d916c0620c19fb40dd4ccb7fc18e632d

PR #315, opened 2016-02-24
Simpler way of checking if a directory exists

rizar:
See #314


vdumoulin:
LGTM.

vdumoulin:
Tests are failing because the UCI Machine Learning Repository is down.

PR #316, opened 2016-02-24
ilsvrc2012 dataset

aam-at:
Implementation and documentation is based on ilsvrc2010.

##### Compare to ilsvrc2010:
- DevKit changes:
  1. No public url for download.
  2. Meta.mat no longer contains `cost_matrix`
- Data changes:
  1. Test ground truth is not provided. To get a number of test example, I use the number of files in test archive.
  2. No image patching.

Given above differences, I tried to reuse code as much as possible. But duplication wasn't avoided.

### DIFF
--- a/fuel/converters/__init__.py
+++ b/fuel/converters/__init__.py
@@ -21,6 +21,7 @@ from fuel.converters import iris
 from fuel.converters import mnist
 from fuel.converters import svhn
 from fuel.converters import ilsvrc2010
+from fuel.converters import ilsvrc2012
 from fuel.converters import youtube_audio
 
 __version__ = '0.2'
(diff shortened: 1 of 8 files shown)


dwf:
Hey, thanks for beginning this! I have some thoughts on how to reuse the metadata parsing code a bit, but I'll take a look at what you've done when I get a chance over the next few days. I think most of the tar reading code can just be moved to a common location and reused in both implementations and thereby only be tested once.

dwf:
Glad they got rid of the "patch images", those were a big pain. :)

aam-at:
@dwf Currently, I didn't do any reusing. "patch images" are also a big pain for refactoring :)
In the initial commit, I just copied your implementation for ilsvrc2010 with the idea that it can help us figure out what refactorings we can do.

The only code reusing which went smoothly was for tests.

aam-at:
I did minor refactoring. As a fast solution to avoid code duplication, create empty temporary tar file for patch images. A better way should probably provide some sort of flag. But for now, I just leave it here for feedback.

aam-at:
Issues:
- Tests fail on `test_server` for python3.4. It seems isolated from changes that I've made.
> FAIL: tests.test_server.TestServer.test_pickling

- Coverage decreases but tests are similar to `ilsvrc2010` which has also low coverage.

PR #318, opened 2016-02-25
Skip test_pickling

rizar:


rizar:
@bartvm , is this OK?

bartvm:
Sure, although ideally I would suggest doing something like

```python
def try(f):
    def try_f(*args, **kwargs):
        try:
            f(*args, **kwargs)
        except AssertionError as e:
            raise SkipTest('Skipping test that failed with: {}'.format(e))
    return try_f
```

bartvm:
And for completeness, I meant it as a decorator of course:
```python
@try
def test_something():
    pass
```

rizar:
I just put that directly in the code, let's hope that will not have more unpredictable tests and will never need a decorator for them.

tests/test_server.py
```diff
@@ -34,12 +35,15 @@ def test_server(self):
...
-                assert_allclose(*data, rtol=1e-2)
+        try:
+            self.stream = cPickle.loads(cPickle.dumps(self.stream))
+            server_data = self.stream.get_epoch_iterator()
+            expected_data = get_stream().get_epoch_iterator()
+            for _, s, e in zip(range(3), server_data, expected_data):
+                for data in zip(s, e):
+                    assert_allclose(*data, rtol=1e-3)
+        except AssertionError as e:
+            raise SkipTest("Skip test_that failed with:".format(e))
```
bartvm: You're missing the `{}`

rizar:
Done!

PR #320, opened 2016-02-29
Fix cache dropping last batch

bartvm:
In some situations the `Cache` transformer was dropping the last batch silently. This fixes that and adds a test for it (which fails on the current `master` branch).


dwf:
LGTM

PR #323, opened 2016-03-03
Test set has no labels

bartvm:
@rizar This should get rid of the labels for the test set.


rizar:
Thanks, looks plausible, I will now check that a sensible exception is raised when one tries to access the test set targets.

PR #327, opened 2016-03-06
recover numpy.ndarray alignment after transmisson

Beronx86:
After transformation by start_server and DataServerStream, the numpy.ndarray may become unaligned.

### DIFF
--- a/fuel/server.py
+++ b/fuel/server.py
@@ -70,7 +70,7 @@ def recv_arrays(socket):
         raise StopIteration
     arrays = []
     for header in headers:
-        data = socket.recv()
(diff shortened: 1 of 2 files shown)


abergeron:
That will introduce a copy which is bad for performance.

bartvm:
I guess you could do `data = [a if a.flags.aligned else numpy.require(a, requirements='A') for a in data]`?

abergeron:
Or you could use copy=False which avoids the problem entirely.

rizar:
@bartvm , I checked, it works.

Issue #334, opened 2016-03-24
Passing output_filename to fuel_convert


sygi:
When you convert a dataset, `fuel_convert` creates a hdf5 file of some default name. I saw the converters have the option `output_filename` to overwrite default behavior, but one cannot pass it to fuel_convert.
It would be nice to have, especially for youtube audio, where one currently does:
```
data = YouTubeAudio('XqaJ2Ol5cC4')
```
but it could be:
```
fuel-convert youtube_audio --youtube-id XqaJ2Ol5cC4 --output_filename music
data = YouTubeAudio('music')
```
I have written it for my usage - I'll send you a PR and you'll be able to merge it if you'd like to.

PR #335, opened 2016-03-24
passing output_filename to fuel_convert

sygi:
Fixes just created #334.

### DIFF
--- a/fuel/bin/fuel_convert.py
+++ b/fuel/bin/fuel_convert.py
@@ -58,12 +58,19 @@ def main(args=None):
         subparser.add_argument(
             "-o", "--output-directory", help="where to save the dataset",
             type=str, default=os.getcwd(), action=CheckDirectoryAction)
+        subparser.add_argument(
+            "-r", "--output_filename", help="new name of the created dataset",
+            type=str, default=None)
         # Allows the parser to know which subparser was called.
         subparser.set_defaults(which_=name)
         convert_functions[name] = fill_subparser(subparser)
 
     args = parser.parse_args(args)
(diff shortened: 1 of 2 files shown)


rizar:
This looks good to me, but I am not knowledgeable enough in this part of Fuel to merge it myself.

@vdumoulin , @bartvm , here is a very small PR, can you please take a look?

fuel/bin/fuel_convert.py
```diff
@@ -58,12 +58,17 @@ def main(args=None):
...
+        subparser.add_argument(
+            "-r", "--output_filename", help="new name of the created dataset",
+            type=str, default=None)
         # Allows the parser to know which subparser was called.
         subparser.set_defaults(which_=name)
         convert_functions[name] = fill_subparser(subparser)
 
     args = parser.parse_args(args)
     args_dict = vars(args)
+    if args_dict['output_filename'] is not None:
```
bartvm: Maybe add something like `and os.path.splitext(args_dict['output_filename'])[1] not in ('.hdf5', '.hdf', '.h5')`?

bartvm:
LGTM to me as well, maybe have a look at the one comment I made.

fuel/bin/fuel_convert.py
sygi: Good idea, thanks. Done.

rizar:
In fact, that PR has broken a lot of converters. Namely, all those in which `output_filename` was a keyword argument with a default value. Including https://github.com/mila-udem/fuel/blob/master/fuel/converters/dogs_vs_cats.py#L17

PR #337, opened 2016-04-02
[requires.io] dependency update on master branch

nouiz:


dwf:
This looks fine but why do the builds fail?

bartvm:
Pillow wasn't available yet on the Conda channels, which caused the Conda installation to fail. Pip then tried to install all the packages that were missing, but probably failed with `h5py` because it requires C libraries.

PR #340, opened 2016-04-07
Add links to the code in docs

dmitriy-serdyuk:
A twin brother of mila-udem/blocks#1058 .


dmitriy-serdyuk:
I merge it since it worked for blocks.

dmitriy-serdyuk:
It doesn't work with decorated functions =( and some converters are decorated. I'll dig into it.

dwf:
Might be rectifiable with functools.wraps?

PR #345, opened 2016-04-13
Fix line number retrieval for wrapped functions

dmitriy-serdyuk:
Fixes [problem](https://github.com/mila-udem/fuel/pull/340#issuecomment-209596028) with wrapped functions.


dmitriy-serdyuk:
It works locally.

dmitriy-serdyuk:
I'm afraid, read the docs has ancient python and it doesn't work there. `__wrapped__` attribute is not added in python2.

PR #346, opened 2016-04-13
Get `wraps` function from six

dmitriy-serdyuk:
Another PR in this series: #345 .


bartvm:
Easiest way of testing this is by merging, so let's try.

dmitriy-serdyuk:
Finally! It works on RTD.

PR #348, opened 2016-04-18
FIX: use UTF-8 encoding when reading non-ASCII README.rst

wilrich-msft:
Hi,

because of the Umlauts in README.rst, setup.py fails with
```
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-efh843rm/fuel/setup.py", line 10, in <module>
        LONG_DESCRIPTION = f.read().strip()
      File "/usr/local/anaconda3/lib/python3.5/encodings/ascii.py", line 26, in decode
        return codecs.ascii_decode(input, self.errors)[0]
    UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 1803: ordinal not in range(128)
```

Thanks,
wr

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 """Installation script."""
 from os import path
 import sys
+from io import open
 from setuptools import find_packages, setup
 from distutils.extension import Extension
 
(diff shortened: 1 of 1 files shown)


dwf:
LGTM!

PR #349, opened 2016-04-21
Auto-open out-of-memory HDF5 files.

### LINKED ISSUE #343: Resuming blocks training with manually opened H5PYDataset fails
"No open file handle".

dwf:
Fixes #343.

Realizing now that @rizar proposes a different solution. I'll look at implementing that.

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -564,6 +564,10 @@ class H5PYDataset(Dataset):
             raise ValueError()
         data = []
         shapes = []
+        # TODO: This is not an ideal solution, really unpickling should be 
+        # restoring the state of the dataset fully, i.e. load() must be
+        # modified to account for the out-of-memory case.
+        # See https://git.io/vKkSm
         try:
             handle = self._file_handle
         except IOError:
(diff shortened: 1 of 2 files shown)


fuel/datasets/hdf5.py
```diff
@@ -564,7 +564,11 @@ def _out_of_memory_get_data(self, state=None, request=None):
             raise ValueError()
         data = []
         shapes = []
-        handle = self._file_handle
+        try:
+            handle = self._file_handle
```
rizar: This would probably do the job, but it is somewhat inconsistent with the general idea of `@do_not_pickle_attributes` and `load`. The fact that `load` does not restore the status of the dataset is a bug, and this is probably what should be fixed.

dwf:
I think we should stop letting the perfect be the enemy of the good and merge this. It's technical debt but it's a very manageable kind that is worth amassing IMHO. I will add a comment to the effect that this is a giant hack that needs to be addressed in some follow-up work to do the "right" thing instead of the expedient thing.

rizar:
*sigh*... I don't really have time now to solve it properly, and I guess  neither do you. LGTM.

dmitriy-serdyuk:
Is it ready to merge then?

dwf:
Yep, seems fine to me (I just pushed a better commit message last night, so
assuming all the test still pass...).

PR #351, opened 2016-04-29
Set encoding when opening README.rst

isacarnekvist:
This is a problem when LC_ALL=C for example.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@ from distutils.extension import Extension
 
 HERE = path.abspath(path.dirname(__file__))
 
-with open(path.join(HERE, 'README.rst')) as f:
+with open(path.join(HERE, 'README.rst'), encoding='utf-8') as f:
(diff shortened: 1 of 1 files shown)


rizar:
It does not seem to work with Python 2: 

```
Processing /home/travis/build/mila-udem/fuel

    Complete output from command python setup.py egg_info:

      File "<string>", line 1, in <module>

      File "/tmp/pip-mQKjA8-build/setup.py", line 9, in <module>

        with open(path.join(HERE, 'README.rst'), encoding='utf-8') as f:

    TypeError: 'encoding' is an invalid keyword argument for this function
```

dwf:
io.open
https://stackoverflow.com/questions/10971033/backporting-python-3-openencoding-utf-8-to-python-2

On Fri, Apr 29, 2016 at 10:56 AM, Dzmitry Bahdanau <notifications@github.com

PR #355, opened 2016-05-06
Rename: one feature and three bug fixes

dwf:
New feature: allow the transformer to intentionally ignore sources in the rename dictionary that don't appear in the dataset.

Bug fixes:
* Only allow mappings that are one-to-one as many-to-one mappings make no sense (ab55194). Actually, only enforce this for sources that are actually present in the dataset (03d99b1).
* Don't let the Rename transformer rename a source to the name of another source (unless that other source is also being renamed). (c14cea6).
* Don't operate wholly in-place by looking up indices in a mutable list, as in certain cases this can lead to incorrect behaviour. (714feac).

### DIFF
--- a/tests/transformers/test_transformers.py
+++ b/tests/transformers/test_transformers.py
@@ -662,6 +662,10 @@ class TestRename(object):
         assert_raises(ValueError, Rename, self.stream,
                       {'X': 'features'}, on_non_existent='foo')
 
+    def test_raises_on_not_one_to_one(self):
+        assert_raises(KeyError, Rename, self.stream, {'X': 'features',
+                                                      'y': 'features'})
+
     def test_intentionally_ignore_missing(self):
         assert_equal(Rename(self.stream,
                             {'X': 'features', 'y': 'targets',
--- a/tests/transformers/test_transformers.py
+++ b/tests/transformers/test_transformers.py
@@ -682,6 +682,13 @@ class TestRename(object):
                             on_non_existent='ignore').sources,
                      ('features', 'targets'))
 
+    def test_not_one_to_one_ok_if_not_a_source_in_data_stream(self):
+        assert_equal(Rename(self.stream,
+                            {'X': 'features', 'y': 'targets',
+                             'Z': 'targets'},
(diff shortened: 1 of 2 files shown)


dwf:
Okay, tests pass (initial failure was some spurious Travis problem).

It's probably easiest to review this commit by commit, as each commit is pretty easy to digest on its own (by design).

tests/transformers/test_transformers.py
```diff
@@ -662,6 +662,10 @@ def test_raises_on_invalid_kwargs(self):
         assert_raises(ValueError, Rename, self.stream,
                       {'X': 'features'}, on_non_existent='foo')
 
+    def test_raises_on_not_one_to_one(self):
+        assert_raises(ValueError, Rename, self.stream, {'X': 'features'},
+                      on_non_existent='foo')
```
dmitriy-serdyuk: The test name doesn't correspond to the internals.

tests/transformers/test_transformers.py
```diff
@@ -658,6 +658,37 @@ def test_leaves_data_unchanged(self):
...
+
+    def test_name_swap(self):
+        assert_equal(Rename(self.stream,
+                            {'X': 'y', 'y': 'X'},
+                            on_non_existent='ignore').sources,
+                     ('y', 'X'))
+
+    def test_raises_on_not_one_to_one(self):
+        assert_raises(ValueError, Rename, self.stream, {'X': 'features'},
+                      on_non_existent='foo')
```
dmitriy-serdyuk: Still, this test tests some other stuff. Copy-paste?

dmitriy-serdyuk:
LGTM after the test is fixed.

tests/transformers/test_transformers.py
dwf: Yep, my bad. Will fix that shortly.

dwf:
Rebased in the correct test in 768c435. Merging as soon as Travis passes.

dwf:
... No, not quite. fcf44b0 contains the test with the right exception name.

PR #357, opened 2016-05-10
Another Rename fix

dwf:
#355 didn't quite implement one case I was interested in: the ability to do many-to-one mappings when only one of the old source names exists. This fixes it and adds a test.


dwf:
Since this is a fix of an old PR's stated aims, and a 10 line diff, I'll merge without review this time...

PR #368, opened 2016-10-11
Make get_data optional in AbstractDataStream.

dwf:
I've noticed that it's often a lot more expedient to stub out `get_data` if all you care is epoch iterators, so we might as well make it an optional implement.

### DIFF
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -67,7 +67,6 @@ class AbstractDataStream(object):
                                  self.iteration_scheme))
         self._produces_examples = value
 
-    @abstractmethod
     def get_data(self, request=None):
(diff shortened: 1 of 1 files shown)


rizar:
I understand you right, your point is that very often the only method of a data stream that is actually called is `get_epoch_iterator`. Thus, it makes sense to make all other methods optional.

If my understanding is correct, I approve this change. I guess you still have to rebase.

dwf:
@rizar I made everything but `get_epoch_iterator` optional. Waiting on a Theano fix to Theano/Theano#5088 to un-break our tests.

nouiz:
THeano PR is merged.

PR #379, opened 2017-01-26
Refactor conda download into a separate script

dmitriy-serdyuk:
I want to add this script to @mila-udem/blocks. In order not to copy paste the same code, let's create a single script.

This will be a multiple stage process:
 - Merge this script
 - Point travis to it
 - Use it in blocks

@rizar?


dmitriy-serdyuk:
Hmm, it is already in blocks. In this case I don't know why tests there are failing.

PR #381, opened 2017-02-08
Make mapping transformer accept dictionary

### LINKED ISSUE #372: Allow passing dict -> dict functions to Mapping
Currently `Mapping` works with functions which transform list of sources to another list. Although, order of sources is fixed, a small mistake inside this function may cause weird bugs.

dmitriy-serdyuk:
I frequently feel frustrated that when I change my data pipeline some mapping transformer in the middle receives a different number of arguments. Now it is possible to work with a dictionary rather than with a tuple:

```python
# Easiest way to use
def double_square(data):
    return {'features': 2 * data['features'], 'targets': (data['targets']) ** 2}
stream = Mapping(stream, double_square, mapping_accepts=dict)
```

Fixes #372.

### DIFF
--- a/docs/api/transformers.rst
+++ b/docs/api/transformers.rst
@@ -25,3 +25,10 @@ Transformers for sequences
     :undoc-members:
     :show-inheritance:
 
+Other
+-----
+
+.. automodule:: fuel.transformers
+    :members:
+    :undoc-members:
+    :show-inheritance:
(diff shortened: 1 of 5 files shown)


dmitriy-serdyuk:
Last commit is just refactoring while I'm at it and can be reviewed separately.

dmitriy-serdyuk:
@rizar , can you take a look?

dmitriy-serdyuk:
@rizar , it works with dict -> dict functions now as we discussed.

PR #383, opened 2017-02-28
Local cache [WIP]

### LINKED ISSUE #173: Local cache for datasets

dmitriy-serdyuk:
Continuation of #287

Fixes #173


fuel/utils/cache.py
```diff
@@ -0,0 +1,387 @@
...
+        # do not locally copy the data
+        if self.dataset_local_dir == "":
+            return filename
+
+        common_msg = ("Message from fuel local cache of dataset"
+                      "(specified by the environment variable "
+                      "FUEL_LOCAL_DATA_PATH): ")
+        # Make sure the file to cache exists and really is a file
+        if not os.path.exists(remote_name):
+            log.error("Error : Specified file %s does not exist" %
```
rizar: The `%`-style formatting is not available in Python 3. Please use `str.format`.

fuel/utils/common_strings.py
```diff
@@ -0,0 +1,62 @@
...
+Mac OS X
+========
+Environment variables on Mac OS X work the same as in Linux, except you should
+modify and run the "source" command on ~/.profile rather than ~/.bashrc.
+"""
+
+viewer_command_error_essay = """
+PYLEARN2_VIEWER_COMMAND not defined. PLEASE READ THE FOLLOWING MESSAGE
+CAREFULLY TO SET UP THIS ENVIRONMENT VARIABLE:
+pylearn2 uses an external program to display images. Because different
```
rizar: Mentions of pylearn2 it the docstring are confusing

fuel/datasets/hdf5.py
```diff
@@ -173,6 +174,7 @@ def __init__(self, file_or_path, which_sets, subset=None,
             self.external_file_handle = file_or_path
         else:
             self.path = file_or_path
+            self.path = dataset_cache.cache_file(self.path)
```
rizar: I assume that we should also change the way the files are opened in all other basic dataset classes.

fuel/utils/common_strings.py
```diff
@@ -0,0 +1,62 @@
+environment_variable_essay = """
```
rizar: `essay` is a weird name for a variable.

fuel/utils/write_lock.py
```diff
@@ -0,0 +1,327 @@
+import atexit
```
rizar: Do we really want to copy all this code to Fuel? Can we not use another library to lock files?

fuel/utils/common_strings.py
rizar: In general, I think that `viewer_command_error_essay` is not used in the cache code.

fuel/utils/write_lock.py
dmitriy-serdyuk: I checked it and I haven't found libraries for that. Stackoverflow mostly suggests to copy-paste a piece of code of questionable quality that doesn't cover all cases. Should we create a small library?

fuel/datasets/hdf5.py
dmitriy-serdyuk: Hdf5 dataset is the standard one, I decided to start with it.

dmitriy-serdyuk:
Users may share the cache, it depends on the `FUEL_LOCAL_DATA_PATH` variable. The code checks if there is space available. Then the dataset is copied to the local directory and the function returns a new path to the dataset.

CC @nouiz

dmitriy-serdyuk:
Perhaps, we need to add a checksum to the local filename so we can compare if the original file is the same. 

@nouiz , a user may have disc quota smaller than available space, what do we do about that?

nouiz:
On our computers, users don't have quota on the local disk. I didn't see
that. If that happen, I think there is a way to limit the size of the
cache, the user should just use that.

fuel/config_parser.py
```diff
@@ -51,6 +51,11 @@
    The default :class:`~numpy.dtype` to use for floating point numbers. The
    default value is ``float64``. A lower value can save memory.
 
+.. option:: local_data_path
+
+   The local path where the dataset is going to be copied. This is a useful
```
rizar: Could you please elaborate that the copying will only take place at the first time, and after that the cached copy will be used?

fuel/utils/cache.py
```diff
@@ -0,0 +1,264 @@
...
+
+    Parameters
+    ----------
+    remote_fname : str
+        Remote file to copy
+    local_fname : str
+        Path and name of the local copy to be made of the remote file.
+
+    """
+
```
rizar: Can we add `log.debug` here, so that impatient users could see that their files are being copied?

maximecb:
Following a discussion with Fred, I would like to take over this PR and try to get the feature completed. Will fork dmitriy's fuel repository, fix merge conflicts and open a new PR.

rizar:
Great! I will take a look at Dima's PR to refresh my memory of what was
left to be done.

rizar:
Closed as #400 has been merged.

PR #385, opened 2017-03-27
Don't import Theano

rizar:
Fuel should not import Theano. Right now if I run  Fuel data server in the same shell where I run my training job, it ends up occupying another GPU because `device` in `THEANO_FLAGS` is set to GPU. This basically makes it impossible to use Fuel server with slurm.

On the other hand, we should notify the users after this PR, because they will have to set `floatX` in their `~/.fuelrc` explicitly.

@bartvm

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ before_install:
   - export PATH=$HOME/miniconda/bin:$PATH
   - conda update -q --yes conda
   - export FUEL_DATA_PATH=$TRAVIS_BUILD_DIR/data
+  - export FUEL_FLOATX=float64
 install:
   # Install all Python dependencies
   - |
--- a/fuel/config_parser.py
+++ b/fuel/config_parser.py
@@ -196,14 +196,6 @@ config.add_config('extra_downloaders', type_=extra_downloader_converter,
                   default=[], env_var='FUEL_EXTRA_DOWNLOADERS')
 config.add_config('extra_converters', type_=extra_downloader_converter,
                   default=[], env_var='FUEL_EXTRA_CONVERTERS')
-
-# Default to Theano's floatX if possible
-try:
-    from theano import config as theano_config
-    default_floatX = theano_config.floatX
-except Exception:
-    default_floatX = 'float64'
-config.add_config('floatX', type_=str, env_var='FUEL_FLOATX',
-                  default=default_floatX)
+config.add_config('floatX', type_=str, env_var='FUEL_FLOATX')
 
 config.load_yaml()
--- a/fuel/config_parser.py
+++ b/fuel/config_parser.py
@@ -186,6 +186,7 @@ class Configuration(object):
         if default is not NOT_SET:
             self.config[key]['default'] = default
 
+
 config = Configuration()
 
 # Define configuration options


nouiz:
It is possible to know if Theano is a loaded module in Python, and if so, use it floatX. Would doing this help here?

dmitriy-serdyuk:
Good idea, @nouiz . As far as I understand, it is a simple check `'theano' in sys.modules`.  I like it, but a small downside is that a different order of imports changes configuration.

rizar:
Hmm, but then I anticipate a situation where the training job and the fuel
server have different floatX settings. Would not it be very bad?

dmitriy-serdyuk:
Yeah, this is bad. I wish it was possible to import theano config without the rest of theano.

nouiz:
I think there is a design problem. There should be only 1 floatX defined by
the user. Why this can't be pushed to the server when the client connect to
it?

Anyway, I don't know how much work that is as I don't know the current code
base. Just take my comments as comments. Your current PR will cause cases
where Theano floatX is set, but not fuel floatX, so it will happen they
aren't the same, just a different places. Maybe that place is better?

rizar:
I am not sure if it is a design problem. Fuel was intended to be a
standalone framework, and it should not look into Theano's configuration, I
think. As you said, there will  be troubles when users set different floatX
for the two, but I am afraid there's nothing we can do. If Theano were not
that aggressive in grabbing a GPU right when `import theano` is invoked, or
if it were possible to import Theano config without importing Theano, we
could go for these solutions. But for now it seems like the best we can do
is just remove these dangerous lines of code from Fuel.

dmitriy-serdyuk:
@nouiz , it is very hard to propagate a client's config to a data server because the later is created by user (as in the [tutorial](http://fuel.readthedocs.io/en/latest/server.html#putting-it-together)). This approach is very flexible, you can do crazy thing like one data path for the server and another for the client. 

@rizar , is it better to add a check if the floatX is the same and throw a loud exception if not?

rizar:
@dmitriy-serdyuk , you can't check Theano's floatX without grabbing the GPU :)

dmitriy-serdyuk:
I mean following:

1. If theano module is loaded, fuel assigns `fuel.config.floatX` to the one from theano. Of course, the fuel config file always has a priority and this issue does not exist if `floatX` is defined in `fuelrc`.
2. If the floatX is never defined we assign it to float64. The same as it done in numpy.
3. As soon as a client connects for a fist time to a server they exchange their inferred `fuel.config.floatX` and ensure it is the same.

rizar:
It sounds like a lot of stuff to implement and maintain, especially any
kind of client-server communication. And solution that depend on the order
in which models are imported... I would much rather just cut the link
between Theano and Fuel until the time when Theano supports reading the
config without grabbing a GPU.

dmitriy-serdyuk:
Ok. Fix the test then and check if documentation mentions this.

nouiz:
That is false. Just make device=cpu and Theano won't grad a GPU.

ServerDataStream() could take as a parameter a dtype that would be the
floatX to use for that server. So the user could pass
theano.config.floatX of whatever dtype he want.

dmitriy-serdyuk:
Are you suggesting to change from a python server script `device=gpu` to `device=cpu` in `THEANO_FLAGS` variable? In this case we'll have to parse the `THEANO_FLAGS` string manually, and this is even more cumbersome.

rizar:
I updated Travis, tests seems to pass. Waiting for an LGTM (@bartvm, @dmitriy-serdyuk). 

When this is merged, I will write an email to the mailing lists.

dmitriy-serdyuk:
LGTM.

PR #386, opened 2017-04-19
fix pickling ServerDataStream and make it lazy

rizar:
Two changes:
- pickling an unpickled stream used to fail
- I would really like `ServerDataStream` to postpone establishing the connection until it's really necessary for flexibility reasons. Also, it was kind of asymmetric so far that in during the construction it connects immediately but it doesn't so when unpickled.

### DIFF
--- a/fuel/streams.py
+++ b/fuel/streams.py
@@ -219,7 +219,7 @@ class ServerDataStream(AbstractDataStream):
         self.host = host
         self.port = port
         self.hwm = hwm
-        self.connect()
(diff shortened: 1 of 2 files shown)


dmitriy-serdyuk:
I don't see where you call `connect()`. Should it lazily connect when you access `context`?

rizar:
get_data() calls connect()

dmitriy-serdyuk:
Ok, LGTM then.

PR #388, opened 2017-04-19
Add test for mapping adding sources with dict

### LINKED ISSUE #387: Mapping won't work with mapping_accepts=dict and add_sources
@dmitriy-serdyuk , there was another option in `Mapping` transofrmer, and I think it's incompatible now with the new feature that you introduced.

dmitriy-serdyuk:
Fixes #387


rizar:
oh no, I just created #389 ...

dmitriy-serdyuk:
Haha, almost the same, but you didn't forget to save original data.

PR #389, opened 2017-04-19
fix the case mapping_accepts=dict and add_sources=smth

### LINKED ISSUE #387: Mapping won't work with mapping_accepts=dict and add_sources

rizar:
fix #387

### DIFF
--- a/fuel/transformers/__init__.py
+++ b/fuel/transformers/__init__.py
@@ -225,14 +225,17 @@ class Mapping(Transformer):
         if request is not None:
             raise ValueError
         data = next(self.child_epoch_iterator)
+        mapping_input = data
         if self.mapping_accepts == dict:
-            data = OrderedDict(equizip(self.data_stream.sources, data))
-        image = self.mapping(data)
+            mapping_input = OrderedDict(equizip(self.data_stream.sources,
+                                                data))
+        image = self.mapping(mapping_input)
+        image_sources = self.add_sources if self.add_sources else self.sources
         if self.mapping_accepts == dict:
-            image = tuple(image[source] for source in self.sources)
-        if not self.add_sources:
-            return image
-        return data + image
+            image = tuple(image[source] for source in image_sources)
+        if self.add_sources:
+            return data + image
+        return image
 
 
 @add_metaclass(ABCMeta)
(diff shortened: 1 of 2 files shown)


fuel/transformers/__init__.py
```diff
@@ -225,14 +225,17 @@ def get_data(self, request=None):
...
             raise ValueError
         data = next(self.child_epoch_iterator)
+        mapping_input = data
         if self.mapping_accepts == dict:
-            data = OrderedDict(equizip(self.data_stream.sources, data))
-        image = self.mapping(data)
+            mapping_input = OrderedDict(equizip(self.data_stream.sources,
+                                                data))
+        image = self.mapping(mapping_input)
+        image_sources = self.add_sources if self.add_sources else self.sources
```
dmitriy-serdyuk: It is easier to read when the default case goes first.

fuel/transformers/__init__.py
```diff
@@ -225,14 +225,17 @@ def get_data(self, request=None):
...
         if self.mapping_accepts == dict:
-            image = tuple(image[source] for source in self.sources)
-        if not self.add_sources:
-            return image
-        return data + image
+            image = tuple(image[source] for source in image_sources)
+        if self.add_sources:
+            return data + image
+        return image
```
dmitriy-serdyuk: Same here.

fuel/transformers/__init__.py
rizar: I remember reading that having conditions like `if not smth` is a bad style.

dmitriy-serdyuk: You probably mean `if not x is None`. My common sense tells me that the `else` body is exceptional and rare.

But it is not a big concern.

nouiz:
I think it was for Python 2.4 support.

Le mer. 19 avr. 2017 15:44, Dzmitry Bahdanau <notifications@github.com> a
écrit :

dmitriy-serdyuk:
@rizar , you want to fix the test? And I will merge it.

PR #392, opened 2017-05-08
Introduce `source_shapes` for H5PYDataset

rubenvereecken:
This pull request is meant to initiate discussion and is by no means finished.

I needed to get the dimensions of my data before reading any data from my HDF5 files. There is the `num_examples` attribute but of course that's only limited to one dimension. I could not find any straightforward way to get all dimensions.. except `H5PYDataset.source_shapes` seemed to represent what I wanted. But it wasn't really implemented. So it might very well be that I missed a way of getting my sources' dimensions but in the meantime I've implemented the `source_shapes` attribute to accomplish what I need, albeit in not all possible scenarios.

If you agree that this `source_shapes` attribute is useful, I could look into how to complete the feature because currently it only works if the user has provided no custom slices (which suits my use case just fine for now).

### DIFF
--- a/fuel/datasets/hdf5.py
+++ b/fuel/datasets/hdf5.py
@@ -488,13 +488,29 @@ class H5PYDataset(Dataset):
                     shapes = None
                 source_shapes.append(shapes)
             self.data_sources = tuple(data_sources)
+            # TODO why is this a tuple, while `.sources` is a list?
             self.source_shapes = tuple(source_shapes)
             # This exists only for request sanity checking purposes.
             self.in_memory_subset = Subset(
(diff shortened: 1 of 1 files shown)


dmitriy-serdyuk:
This is a useful feature. Do you plan to get one batch from the dataset to compute its shape?

rubenvereecken:
No, as far as I know h5py datasets have the `shape` attribute which should be just fine. I've never used dimension scales though, nor do I know about variable-length datasets. Either way, I think variable length only goes in the first dimension? Whereas if you'd get a batch from the dataset you don't know anything about the first dimension anyway.

PR #400, opened 2017-08-28
Local cache continued

### LINKED ISSUE #173: Local cache for datasets

maximecb:
Taking over the work from #383 (sidenote: you guys should probably close #173 and #383).

All I've done so far is fix the conflicts and get tests running locally. Opening this PR so I can get your feedback as to what remains to be done/fixed.

One thing I did notice is that if I just run `nose2 tests` after installing fuel, I get this error:

```
ConfigurationError: Configuration not set and no default provided: local_data_path
```

Wondering if we should set caching to disabled by default, and what default value `local_data_path` should be set to in order to disable caching, is it just the empty string?


maximecb:
Following an in-person discussion with @rizar, I will make it so Fuel doesn't give an error if the cache path is not set. I will also add a mention of the caching behavior in the docs.

maximecb:
@rizar @nouiz 

I've made it so caching is disabled by default, and Fuel doesn't complain if the `FUEL_LOCAL_DATA_PATH` variable is not set. I also added a section about caching to the docs. The tests are all passing locally on my machine, with and without caching.

Regarding the failing scrutinizer tests: they are related to some scrutinizer-specific extension of the pep257 linter. I can't run these checks on my machine because their linter seems to be proprietary. Arnaud suggests disabling scrutinizer on this repository, a motion that I would agree with.

Let me know if there is anything else that needs changing.

docs/caching.rst
```diff
@@ -0,0 +1,18 @@
...
+point to a shared network drive. For example, when configuring multiple
+machines in a cluster to work on the same data in parallel.
+However, this can easily cause network bandwidth to become saturated.
+
+To avoid this problem, Fuel provides a second configuration variable
+named ``local_data_path``, which can be set in ``~/.fuelrc``. This
+variable points to a filesystem directory to be used to act as a local
+cache for datasets.
+
+This variable can also be set through an environment variable as follows:
```
rizar: Can you please mention that caching is currently only implemented in `H5PyDataset`, and that in order to add caching to other dataset one should use `cache_file` function?

rizar:
Regarding the fate of Scrutinizer, it would be great to hear from @bartvm  who started using at the first place.

docs/caching.rst
maximecb: @rizar Done!

maximecb:
@rizar now passing the scrutiny

rizar:
Thanks!

maximecb:
@rizar NP. Let me know if there are other issues I could potentially help with, on this project or others.

PR #402, opened 2017-09-14
Adding camvid dataset

bordesf:
Hi,
I am starting doing some research on image segmentation. So, I was thinking that it could be useful for everyone to add those datasets on fuel. Here we have the camvid dataset [Camvid]. Currently this dataset is only available on a github page (https://github.com/alexgkendall/SegNet-Tutorial/tree/master/CamVid). I was thinking to create a zip file and upload it on the web, in order to make it downloadable by fuel (After asking the permissions of the authors). Is there any ftp server where I could upload it eventually ?

[Camvid] http://mi.eng.cam.ac.uk/research/projects/VideoRec/CamVid/

### DIFF
--- a/fuel/converters/__init__.py
+++ b/fuel/converters/__init__.py
@@ -16,6 +16,7 @@ from fuel.converters import caltech101_silhouettes
 from fuel.converters import celeba
 from fuel.converters import cifar10
 from fuel.converters import cifar100
+from fuel.converters import camvid
 from fuel.converters import dogs_vs_cats
 from fuel.converters import iris
 from fuel.converters import mnist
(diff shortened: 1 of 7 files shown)


charlesreid1:
Hello,

I am also interested in this data set, and several others that aren't in the fuel examples directory. A few questions/comments.

Questions for fuel maintainers:

The SegNet tutorial and (presumably) data set the OP linked to are released under a CC non-commercial license. Fuel is released under an MIT license, which is incompatible (allows commercial usage). But since the data itself is not being hosted in fuel, is this a problem? Do the maintainers of fuel have a policy about what kind of license a data set must be released under to include it as an example?

It seems like it is a slippery slope to start adding many data sets to the examples, unless the repository maintainers intended the fuel repo to become a catalog of data sets. I believe the examples are intended to illustrate how to create your own loaders/converters in fuel. Is this a correct interpretation, or do you welcome PRs with new data sets?

Question for maintainers/regular fuel users:

What is the recommended workflow that folks should use to share their own fuel loaders/converters for popular datasets that don't make it into the fuel examples directory - should the fuel repo be forked? Or should we create an external package like camvid_fuel that imports fuel and then defines some loaders/converters, and can then be used like `import camvid_fuel; X_train, Y_train = camvid_fuel.load_data()`?

PR #405, opened 2018-08-28
Bump Pillow to be sure it is a secure version

nouiz:


dmitriy-serdyuk:
Some tests fail with the new pillow `FAIL: tests.transformers.test_image.TestRandom2DRotation.test_random_2D_rotation_batch_stream`. I think their defaults changed.

dmitriy-serdyuk:
It looks like the old pillow had a bug there. I tested on images of bigger sizes and pillow 3.4 -- 5.2 works fine and the same here.

Not sure why old pillow had this answer, but in this test the rotation angle is 12.25 degrees. It means that the difference in the output image will be less than 1 pixel. So, the new pillow is giving the right answer.

dmitriy-serdyuk:
Basically, it fails now only because of `--allow-external` in blocks. I removed it in mila-udem/blocks#1198.

Feel free to merge both PRs.

dmitriy-serdyuk:
Tests passed, I merge

